PR #11, opened 2017-01-13
reduce to functionality we seem to agree on.

struberg:
That means no portable SPI for now (mp-config-1.0)
And with the lack of SPI also no Builder

I also fixed the copyright year and from IBM to "Author and Authors".


Emily-Jiang:
-1. Did you see my comments on the mailing list. All files are going to stay except removing the ordinal method.

struberg:
Emily that makes no sense. 
Either we have ConfigSources - then properly done. Or we do not have them. 
Without any discovery it makes absolutely no sense to have any SPI **yet**!

Consider we keep the ConfigSource. What to do with them? You see?

Emily-Jiang:
Disagree. Please discuss on the mailing list and see my comments. The ConfigSource is used to add config sources to the builder programmatically.

struberg:
Well this already got discussed on the mailing list and got a few +1 as most common denominator.

I'll try to explain again, but first let me understand what you aim for. 
How should that be used what you did propose? ConfigProvider.getConfig() but with SPI. How do SPI fit into this picture? Please elaborate!

Maybe related: You said that you also want to keep the Builder? And how does ConfigProvider.getConfig() and the ConfigBuilder relate to each other? Please elaborate as well!

keilw:
Btw, does @struberg have an Eclipse  committer account yet? Otherwise his PRs cannot be merged even if the claim was accepted.

struberg:
Yes, I have one ancient one from the Aether project. Also this doesn't matter. PR can be merged from non eclipse committers. I wonder from where you got your information from?

keilw:
Check the mailing list threads on that. Especially answers by Wayne.

Emily-Jiang:
>How should that be used what you did propose? ConfigProvider.getConfig() but with SPI. How do SPI fit into this picture? Please elaborate!
When you provide a non-default config source, e.g. myProp.xml, you can add the config file using
ConfigProvider.getBuilder().addConfigSource(configSource).getConfig(), where the configSource is an instance of ConfigSource. I plan to contribute a helper class for reading the .xml, .properties and .json. At the moment, the user can easily parse the file themselves.

>Maybe related: You said that you also want to keep the Builder? And how does ConfigProvider.getConfig() and the ConfigBuilder relate to each other? Please elaborate as well!
See my above comment

struberg:
You didn't answer my question. Let me rephrase: If you call ConfigProvider.builder().......getConfig() 20 times with different parameters. What result would you then get with ConfigProvider.getConfig(). Which one of the 20? Or something completely different? 
Point me to the javadoc etc where this is defined. Or try to write a concise JavaDoc for it.

struberg:
Since we reached an agreement for moving further with a more complete approach this pull request is now obsolete.

PR #12, opened 2017-01-13
Update README, Config and ConfigSource

keilw:
Fixed the link to DeltaSpike Config The old link was not DeltaSpike Config but something inspired by it.

The JavaDoc of `ConfigSource` states
```
  @return the 'name' of the configuration source, e.g. 'property-file
	 *         mylocation/myproperty.properties'
```
For the "name of the configuration source" isn't it better to simply call it `getName()`.

### DIFF
--- a/src/main/java/io/microprofile/config/spi/ConfigSource.java
+++ b/src/main/java/io/microprofile/config/spi/ConfigSource.java
@@ -76,11 +76,11 @@ public interface ConfigSource {
 	String getValue(String propertyName);
 
 	/**
-	 * The name of the config source might be used for logging or analysis of
+	 * The id of the config source might be used for logging or analysis of
 	 * configured values.
 	 *
-	 * @return the 'name' of the configuration source, e.g. 'property-file
+	 * @return the id of the configuration source, e.g. 'property-file
 	 *         mylocation/myproperty.properties'
 	 */
-	String getName();
+	String getId();
 }
--- a/src/main/java/io/microprofile/config/spi/ConfigSource.java
+++ b/src/main/java/io/microprofile/config/spi/ConfigSource.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 IBM Corp. and others
+ * Copyright (c) 2016, 2017 IBM Corp. and others
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


keilw:
Thanks for the reminder, also updated `Config` accordingly.

struberg:
That's pretty blunt to first break things, then changing it back to how it was in my original proposal and then also claiming authorship.
One could hope it was unintended, but since you tried to muddy my proposal for months we may assume you actually did read it?

Anyway, check here for the original version done in June last year:
https://github.com/struberg/javaConfig/blob/master/api/src/main/java/io/microprofile/config/Config.java#L47

keilw:
It says DeltaSpike Config which it is not. 
I had a look like the one by Gunnar that also exists. Wonder, if they can be combined somehow?
If this pre-proposal was worth mentioning, then so was Gunnar's I'd say, but I leave that up to @Emily-Jiang

struberg:
> It says DeltaSpike Config which it is not.
That's not what I meant. I was just wondering about you just changing linebreaks and then claiming authorship? That's a bit steep...

Bty, my proposal and DeltaSpike is basically the same, I just extracted the pieces out from DeltaSpike to have a 'config-only' project. That's it. DeltaSpike contains tons of other useful features and is primarily CDI focused, but if people _only_ want config then it's a tad too much.
But happy to point to both. 

Gunnars proposal is also a good one, but we didn't even target those aspects yet. This is in the direction of owner (http://owner.aeonbits.org/docs/usage/). It is great for end users but not always usable for frameworks.

src/main/java/io/microprofile/config/spi/ConfigSource.java
```diff
@@ -16,62 +16,71 @@
...
-	
 
+	/**
+	 * The name of the config source might be used for logging or analysis of
+	 * configured values.
+	 *
+	 * @return the 'name' of the configuration source, e.g. 'property-file
+	 *         mylocation/myproperty.properties'
+	 */
+	String getName();
```
Emily-Jiang: I planed to change it to getId(), which is better as the name is confusing, e.g. is it the config file name...

keilw: Sounds equally good. Is it possible to change that after the PR? It also addresses things like the DeltaSpike URL or last ASF license header in the Readme file.

keilw: Indeed, `getId()` goes in a slightly similar direction as the correlation ID topics @struberg also discussed about.

keilw:
I changed the method go `getId()` now as suggested. Please review.

struberg:
@Emily-Jiang @keilw Can you please explain to me for what this method is actually good for now that you want to ditch the dynamic discovery?

keilw:
It replaces `getName()` but that at least from the JavaDoc was not meant for dynamic discovery (otherwise the JavaDoc may have been incomplete or insufficient) 
"The id of the config source might be used for logging or analysis of  configured values."
If you consider supporting log or event correlation in a distributed Cloud environment, it certainly can't hurt to have such information, whether you simply put it into log information or use it for more sophisticated analysis like Kibana, etc.

struberg:
So you are for dropping the ordinal but cannot explain what the getConfigName() method does?

As the class is still 1:1 copied over from my proposal (even including JavaDocs) I can only tell you that it originally only made sense in conjunction with the ordinal mechanism.
https://github.com/struberg/javaConfig/blob/master/api/src/main/java/io/microprofile/config/spi/ConfigSource.java#L93

Not sure what it should do now after people seemingly randomly add/remove parts of methods which originally were heavily relating to each other. That's the reason why I ask. I have a hard time to grasp the overall systematic when important parts get cut out.

Btw, getId() is not better than getName(). That's why I named it getConfigName(). That's what it effectively is. We might also name it getConfigLocation().

keilw:
Ask @Emily-Jiang please, she suggested `getId()` as the better method name. Just so much on the ordinal. No other framework that's widely succesful (mostly Archaius or Spring though the latter uses more and more by Netflix, Dropwizard or others, I doubt Spring uses DeltaSpike however) in this area knows or uses this concept. It's about to be replaced in Tamaya, too. And everyone who's involved in a discussion other than @struberg seems to prefer a better alternative. You see what happened in DeltaSpike. Inconsistent, often badly documented methods. With one sentence claiming the method returns the name of the "config" while the next one suggests, it's the name of the "config source".  That's why either at Microprofile or places like the JCP people look for a "best of breed" vendor-neutral solution, not something dictated by a single person or company ;-)

keilw:
And please stick to this PR, it is not about dropping the ordinal, that might come with a different one.

keilw:
Some refactoring created conflicts, so going to create another PR for the problems that have not been addressed.

PR #32, opened 2017-01-20
Fixed wrong urls in POM and Readme plus license header in Readme

keilw:
This is the subset of the prior PR addressing IP issues (and fixing https://github.com/microprofile/microprofile-config/issues/10)


Emily-Jiang:
Thanks @keilw !

PR #42, opened 2017-02-01
CDI API for injection of configuration values

OndrejM:
This PR proposes CDI-based API to access configuration values within a CDI container.

It consists of 2 parts:
  - plain value injection using `@Inject @ConfigProperty`
  - reading values from injected instance of a custom interface annotated by `@ConfigRegistry`, ala DeltaSpike Data repositories

The only new files in API module are `@ConfigProperty` nad `@ConfigRegistry` - see their javadoc for more details and code examples.

There are examples in `CDIPlainInjectionTest` in TCK module for both approaches and various use cases (static value injection, dynamically loaded values, optional values, ...). 

I pick some examples here to demonstrate the idea:

```
    @Inject
    @ConfigProperty("my.long.property")
    Long injectedLongValue;
```

```
    @Inject
    @ConfigProperty("my.long.property")
    Optional <Long> injectedOptionalLongValue;  
```

```
    @ConfigRegistry
    public interface MyConfig {
 
        @ConfigProperty("my.long.property")
        Long getLongValue();  

         @ConfigProperty("my.long.property")
        Optional <Long> getOptionalLongValue();  
    }
 // and usage:
        @Inject
        MyConfig myConfig;
```


tck/src/main/java/org/eclipse/microprofile/config/tck/CDITestsFunctions.java
```diff
@@ -0,0 +1,43 @@
+package org.eclipse.microprofile.config.tck;
```
OndrejM: In your repo, you only add classes from TCK into your deployment. How do you include classes required by the implementation? This is the sole purpose of the `MicroProfileConfigTestConnector` so far - I needed to add classes from the implementation into deployment archive. It should also be possible to create an arquillian container that contains those classes, but I don't know how to do that. I was using Weld Arquillian container - I would need to add those classes into its classpath.

Emily-Jiang:
Thank you Ondrej for the PR! I agree ConfigProperty is needed. I don't think it is necessary to have ConfigRegistry. I think it does not add any value. It will not force any implementation to do anything. I suggest in the config doc - add CDI integration section, we should say the Config should be provided as a CDI bean.

tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPlainInjectionTest.java
```diff
@@ -0,0 +1,235 @@
...
+        // setting configuration via system properties
+        System.setProperty(key, value);
+    }
+
+    private void ensure_property_undefined(String key) {
+        // clearing configuration in system properties if previously set
+        System.getProperties().remove(key);
+    }
+    
+    private static <T> T get_bean_of_type(Class<T> beanClass) {
```
OndrejM: I know I can use inject in Arquilian test, but I wanted to get the implementation details out of the way and write a clean and legible test suite. Tha's also why I moved methods like this to a separate util class.

tck/pom.xml
```diff
@@ -27,6 +27,135 @@
...
+            <artifactId>shrinkwrap-api</artifactId>
+            <scope>compile</scope>
+        </dependency>
+
     </dependencies>
+    
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
OndrejM: I removed the block, but add a maven property to enable changing the regex for method names in the TCK module. For tests, I believe that using names with underscores is much more legible and closer to a plain hand-written text.

tck/pom.xml
```diff
@@ -27,6 +27,135 @@
             <version>6.9.9</version>
             <scope>compile</scope>
         </dependency>
+        
+        <dependency>
+            <groupId>junit</groupId>
```
OndrejM: I didn't include anything impl specific - my implementation resides in a completely different repo, which only depends on the TCK maven module.
The additional dependencies are only to support testing (JUnit, Hamcrest) or make Arquillian work out of the box (Weld connector). 

If the point is to remove the dependency to the Weld connector and force the implementation to pull an appropriate connector, I agree we can go this way.

If the point is to choose Test-NG as the single and unified testing framework, I don't mind. I chose jUnit just as a temporary solution because I'm more productive with it. I plan to rewrite the tests to TestNG, I just wanted to focus on the proposed API.

OndrejM:
@Emily-Jiang I know that ConfigRegistry is just a sugar over injecting Config itself, but it's much more dev-friendly and closer to the CDI programming model. Moreover
 - this approach is already very popular and familiar from DeltaSpke Data and Spring Data. 
 - It does not require repeating string keys over and over again, if the configuration value is used in more places. 
 - it makes it easy to define all configuration keys accepted by an application in one single place, if whole application uses just one interface to access the config, or all such interfaces reside in the same package

In the end, the ConfigRegistry illustrates what I had in mind with the CDI-first approach - if the CDI API was designed before the Java SE, we the Config interface wouldn't be even required in the initial version of the config API (assuming that we target application code first, before general use cases like CDI extensions and Java SE code).

That said, I agree it is not necessary to have it in the first version, I just wanted to point out an alternative approach to accessing values that might change dynamically. IMO, we can even specify additional support for dynamic values later, since CDI `Instance` already provides it for free.

OndrejM:
> add CDI integration section, we should say the Config should be provided as a CDI bean.

This was my implicit assumption, I just forgot to add a test case for this. I'll add it into the spec doc, but I wanted to wait until the docs are cleaned up and synchronized with the current API.

struberg:
> since CDI Instance already provides it for free.

While Instance<T> implements Provider<T> it has a totally different behaviour.
Instance<T> will just give you access to Contextual References in a dynamic way. But not to anything dynamic inside a Contexutal Reference (esp no config).
We could rather make ConfigValue implements Provider and rename getValue() to get().

Emily-Jiang:
@struberg What you suggested sounds good to me. @OndrejM I still think ConfigRegistry is not needed at this stage and it does not fit with the current design. Do you mind leaving it out?

gunnarmorling:
One advantage of the interface based approach ("ConfigRegistry") is the very nice handling of default values via Java 8 default methods. And it doesn't require non-final fields in user code.

OndrejM:
I agree to remove the ConfigRegistry from this PR. It can be added later in next releases, Ill make sure we discuss it after config 1.0. Thanks for showing support for it, @gunnarmorling.

Emily-Jiang:
Thanks Ondrej!

OndrejM:
@struberg Where can I find the `ConfigValue` you mentioned above? I don't see it in the current master, is it in a PR somewhere?

Emily-Jiang:
@OndrejM https://github.com/struberg/javaConfig/blob/master/api/src/main/java/io/microprofile/config/ConfigValue.java. Mark mentioned this interface could extends the Provider.

OndrejM:
Thanks, Emily. I found out that we discussed `ConfigValue` also [here](https://groups.google.com/d/msg/microprofile/DQ2nsyGwEeg/mGO8ukk6DQAJ). I'll follow up on that discussion rather than commenting here.

smillidge:
The property name could be defaulted to the fully qualified class name + field that way there is no need to specify the name. 

A default value could also be specified if the property is not found.

Emily-Jiang:
+1 @smillidge ! However, I think the default name should just the field name.

smillidge:
perhaps simple classname.fieldname I would think just field name was a bit unqualified.

Emily-Jiang:
Well, in your property file, do you prefix the classname for your key? I guess not. Hence the suggestion.

struberg:
It's totally up to the user which prefix he likes to use for his keys. Package name as default is imo a perfectly legit option.

OndrejM:
I also think that defaulting the key to the field name is not enough. Imaging having a field `size` in two different beans - most of the time, it should be a separate property for each bean.

I would go for at least classname.fieldname (to avoid breaking the config when refactoring). Having packagename.classname.field is also OK, as it is very commonly used and natural, though impractical when doing refactoring.

api/src/main/java/io/microprofile/config/inject/ConfigProperty.java
```diff
@@ -0,0 +1,57 @@
...
+ *    javax.inject.Provider{@code <Long>} longValueInjector;  
+ *            // injects a provider for the value of my.long.property property to resolve the property dynamically
+ * </code>
+ * </pre>
+ * @author Ondrej Mihalyi
+ */
+@Qualifier
+@Retention(RUNTIME)
+@Target({METHOD, FIELD, PARAMETER, TYPE})
+public @interface ConfigProperty {
```
OndrejM: So far, I've seen people using 2 different naming conventions for this kind of annotation and a new injection point type (discussed [here](https://groups.google.com/d/msg/microprofile/DQ2nsyGwEeg/WGVcTMGdEAAJ)):

 - `@Inject @PropertyConfig("mykey") PropertyValue<String> mykeyValue`
 - `@Inject @ConfigProperty("mykey") ConfigValue<String> mykeyValue`

Please comment here to let me know which is better.

Emily-Jiang: @Inject @ConfigProperty("mykey") ConfigValue<String> mykeyValue 
makes more sense.

Emily-Jiang: For developers, I don't think the default name would start with class name. What is the normal prefix? I think for the config source under meta-inf\microprofile-config.properties, the key will not have any prefix. For the config source files outside the app, it depends. If the source is for a particular app, it might have the prefix with the app name etc. Therefore, I think it is better not to prebake any prefix.

OndrejM:
By @Emily-Jiang 
> For developers, I don't think the default name would start with class name. What is the normal prefix? I think for the config source under meta-inf\microprofile-config.properties, the key will not have any prefix. For the config source files outside the app, it depends. If the source is for a particular app, it might have the prefix with the app name etc. Therefore, I think it is better not to prebake any prefix.

I think we are looking at the same problem from different point of view. Emily, what you suggest is making life easier for a developer if the config key is already fixed before using it in the code. In this case, I think it is wise (and natural) to expect that the config name is defined in the annotation.

All the others, who commented, expect that the config key is not yet fixed and will be derived from the source code (class + field name).  In that case, it is valid to expect that if a developer made his life easier, he made it more complicated for the person who provides the configuration. This is a similar approach to how JPA generates table names from the code - you either accept the generated names in the database and all data migration SQLs, or have to specify the mapping in the JPA annotations so that the table names are defined by you and not by the generator.

api/src/main/java/io/microprofile/config/inject/ConfigProperty.java
smillidge: I agree classname.package name is a godd default as @OndrejM says this is more amenable to refactoring while still providing a level of name scoping. Obviously the developer can still specify the name directly in the annotation. I'm just advocating this as a default if no name is specified.

Emily-Jiang:
@OndrejM I am confused about this whole discussion. Surely the config file exists before you try to access it. Can you come up with a file containing some properties? I don't think the prefix with package+class makes any sense. Remember the config property can be used by multiple classes? What do you use for the key in the property file?

OndrejM:
@Emily-Jiang - the config file has to exist during runtime, otherwise injection would fail. But you have 2 ways to define it:
 - either you define the configuration keys first, and then the application source code has to adapt
 - or you define the application logic (code) first, the key is generated and the configuration keys in the config file have to adapt

I would say it's a false premise that the config key is always defined by business requirements. It may be so in majority of corporate projects, but it is very often not the case in my experience, at least during initial development.

It's the same as with JPA - you either design your database first, and specify the table names in your app, or you design your app first, and then the database may be generated or respect the generated table names.

For example, if you have 2 properties, let's say `file.size` and `pool.size`, and you want to use them in your app, you either have to specify them by their name (`@ConfigProperty(key="file.size")`), or you have to use a class name and field that matches the key, e.g. `class File, field size`.

In the other case, nobody prescribed the key names, so you may choose them and later document your application for ops. In that case, you may code freely without specifying the key, because in most cases the generated key won't clash. If you really want to use the same config key, you have to specify it via key name in the annotation, e.g. `class File, and field size annotated with @ConfigProperty(key="pool.size").

Is my point of view clear enough now?

Emily-Jiang:
@OndrejM Thanks for the detailed explanation! I understand what you meant now. Personally, I like the default to be the field name, which is clearer. In your scenario, the field name has to be something like file_size. Anyway, what you suggested classname+field name is fine.

Emily-Jiang:
Also you need to do git commit -s to sign off the commit

keilw:
I think I remember that from projects like UOMo, but there the leading repository started at Eclipse.org, so some things are slightly different. Still a question why some corporate reps would consider themselves "Individual" while others represent their company?

OndrejM:
Hi @Emily-Jiang, I have edited and signed-off all my commits (I did a forced push to get them to github). The Eclipse IP checks have passed.

Emily-Jiang:
@OndrejM a couple of comments on your PR. ConfigProperty.java has the incorrect javadoc. All injection should be like: @Inject @ConfigProperty("myprop") ConfigValue<String> myValue; As for ConfigValue, I suggest to have just one method of Optional<T> getValue() and nothing else. As for the copyright, please use just mention the company name and others, instead of listing your own name. When I make more changes, I can add IBM there for an instance. As a minimum, please fix the copyright.

OndrejM:
@Emily-Jiang
Ad. ConfigProperty:
    My intention was to support all these cases:
      1. @Inject Object - to inject the value directly, because in many cases, this is valid even in dynamic scenarios, when the CDI bean has short scope (e.g. RequestScope)
      2. @Inject Optional - this is to make the CDI API in line with the Java SE API in Config interface, which returns Optional from getValue method. Optional also provides nice functional features which are not (yet) available in ConfigValue
     3. @Inject ConfigValue - this was discussed earlier

I don't mind removing the option 1 from this PR if you are strongly against it.
But supporting only ConfigValue creates a discrepancy between the SE and CDI API. I'd like either that both support ConfigValue (ideally with additional methods copied from Optional) or that both support at least Optional.

Ad. Copyright header
   Although it might seem I represent Payara Services as their employee, I'm in fact a separate legal entity working closely with Payara Services. I'm also an individual member of the Eclipse foundation. I'm working on MicroProfile both as a member of the community in my free time, and also within my working time for Payara Services. Therefore I thought it is correct to include both my name and Payara Services in the copyright. If it seems strange or even wrong, I can list just Payara Services, I don't care much.

Emily-Jiang:
For 1 and 2), it is a trap. The injected value is static and it is the fact. Why use it if the 3 ) is well recommeded and working. I'll create a PR to clean it up. 

For copyright, the 'others' part include you already as you are one of the authors. We have spent a lot of time on copyright. If you don't mind, please remove your name and from the copyright header, which will make things much easier.

keilw:
Despite many other companies (including Google with Mark Davis as a Unicode founder) and probably also individuals contribute to ICU4J it says `Copyright (c) 2016 IBM Corporation and others.`
So it seems OK here for most modules Emily/IBM started. If Eclipse Foundation is fine with some permanent employees of Payara, Tomitribe or other companies joining as Individuals rather than representing these companies (although it does not sound like a lack of interest by their employers in Eclipse or Microprofile ;-) then why not. Adding an author tag for those who contributed a larger portion sounds fine for that as it does in other projects.

OndrejM:
I'll change the copyright headers then.

For 1 and 2, I understood, but had a different opinion. But we can remove them from this PR and the version 1.0 and discuss later. 
However, in that case, I would consider copying some functional methods from Optional into ConfigValue so that it is easier to work with. It would help avoid calling ConfigValye.getValue() without checking if the value exists and getting null exceptions if it doesn't.

Emily-Jiang:
See the above commit to demonstrate what I meant. Since PropertyValue was originally from DeltaSpike. We should acknowledge all the original authors.

Emily-Jiang:
@OndrejM if you update all tests to reflect the new changes and we should be good to go.

keilw:
If every single one of those DeltaSpike contributors signed the ECA that should be OK, otherwise better leave those who can't or don't want to.

Emily-Jiang:
@keilw good point.

OndrejM:
@keilw Just to clarify my status with Eclipse Fnd and Payara Services - I cannot join Eclipse as an employee of Payara Services, because I'm simply not legally their employee, although I work very closey with them and represent them. I also consider myself as a community contributor to MicroProfile, as I would continue to support the project even if I didn't work for Payara Services. I hope that's clear enough now. I changed the copyright header to mention only Payara Services though, and I consider myself to be included in the "others" group :)

keilw:
@OndrejM Guess if you do this e.g. as a contractor or similar while working for Payara and not in your spare time at home, it sounds OK.

gunnarmorling:
@OndrejM @Emily-Jiang so are you saying `@Inject @ConfigProperty int myInt` is not going to be supported?

If so, that's a big turn off for me, it makes usage just much more cumbersome. Of course `ConfigValue` can provide notification for value changes (though I thought we all agreed that's a non-goal for 1.0?!), but in cases where that's not needed, injection of the actual value should definitely be supported.

Let's make user's life easy :)

keilw:
No idea if there ever will be a wider standard, but even when it was thought a first phase could enter Java EE 8 that phase also did not aim for property change events.

OndrejM:
> @gunnarmorling: @OndrejM @Emily-Jiang so are you saying @Inject @ConfigProperty int myInt is not going to be supported?

My opinion is that we should support also static injection because it is straight-forward and sufficient in many scenarios. However, I wanted to demonstrate my attitude to decisions that are not widely accepted. With my first version of this PR I wanted to show what is possible, but now I'd rather keep only what we can easily agree on, provided that there is an open door for future discussion and enhancements in next iterations. 

I'd focus on finalizing what we can agree on, so that we have a releasable version. If time allows, we can discuss bringing back the support for injecting values directly (`@Inject @ConfigProperty Integer myInt`), but although I have reasons to think it should be part of config v1.0, I can accept moving that decision after v1.0.

@Emily-Jiang It would be very helpful to keep a backlog of all the topics that we haven't agreed on but we want to come back to them later. It could be also used as a roadmap for future enhancements. Github issues is not a good place, because many different types of issues are mixed, unless we properly label the issues. Maybe a shared google doc would be enough. I'd already put there _injecting static values_ and _ordering of config sources_.

emilyjiang:
@gunnarmorling As you know the injected property is static and will not get the changes if the config source is dynamic config source. This is like a trap.

@OndrejM +1. I think the [wiki](https://wiki.eclipse.org/MicroProfile) is the best place to track the issues and we can record the resolution. I have edited the wiki by adding the config subsection but it will take overnight for it to show up.

gunnarmorling:
> As you know the injected property is static and will not get the changes if the config source is dynamic config source. This is like a trap.

Emily, IIRC, we all agreed that config value updates is a non-requirement for 1.0. Has this changed?

But even _if_ we say that this should be supported, I don't see any reason why users shouldn't be allowed to choose between "simple and non updatable" (i.e. `@Inject int myInt`) and "a bit more complex but updateable" (i.e. `@Inject ConfigValue myInt`) depending on their requirements. Not every case needs updatable values, so we should not make this case more complex than needed, esp. since both can co-exist.

Btw. by using a reasonably short scope (e.g. `RequestScope`) as injection target, you'd also get fresh values when injecting the straight config values instead of a container such as `ConfigValue`. "static" really is a very confusing term to use here.

heiko-braun:
I agree with gunnar, that updates shouldn't be part of 1.0. But if you walk down this path, I think it should not happen on the level of a single injection point. A higher level API to register for change events seems more reasonable IMO.

struberg:
As far as I remember we decided to target the 4 different use cases coming with ConfigValue _after_ we are satisfied with the core config functionality. So I'm also +1 on removing this cut-down variant of ConfigValue. Especially since Ondrej only took parts of the functionality and we really need much more to get a clean solution.

Emily-Jiang:
@struberg @OndrejM @gunnarmorling I am aiming for the short and clean API for the injection. With the current version, all property injection will just use
@Inject @PropertyValue("myName") PropertyValue<String> myValue;
Is the above complicated? It does everything and cope with all kind of scenario. In my value it is clean and easy to use. We don't even need to mention the dynamic but it works for all conditions.

OndrejM:
I really abstain. I explained my opinion above, but I'm fine with going with only ConfigValue injection as @Emily-Jiang suggests, at least for v1.0. In the end, we can experiment with additional MP-compatible CDI extensions to provide injecting values statically for the people who like it.

Using ConfigValue is only indirectly related to dynamic config sources. The main strength of it is that it provides an object wrapper around config values, which can be extended in the future with additional methods and with dynamic behavior if config sources will be dynamic in the future. Even now, it is possible to define a custom config source, and it's completely up to the implementor to support dynamic updates.

> @struberg: So I'm also +1 on removing this cut-down variant of ConfigValue. Especially since Ondrej only took parts of the functionality and we really need much more to get a clean solution.

I only took the methods we need now for injection. I foresee that we can add more methods in the future (without any breaking changes), provided that we agree that the methods make sense and fit a bigger picture. I surely look forward to discussing more improvements, one at a time, although I guess after v1.0.

gunnarmorling:
> I am aiming for the short and clean API for the injection [...] Is the above complicated?

Emily, I think our first and foremost concern should be to make an API simple and convinient for the _user_. If that requires more work for _us_ than that's a trade-off I'm willing to make every day (of course there are limits).

Having `ConfigValue()` _only_ makes the API harder to use as it requires an additional method call when working with config values: it's `if ( myProp.get() > 0 )` vs. `if ( myProp > 0)`. I don't think that's a pattern we should force onto users, hence there should be support for direct injection of property values.

Does it mean more work for us? Yes, a bit, but as said I think that's acceptable and preferred for the sake of a better usable API.

Also consider that `ConfigValue<T>` only can expose boxed values (`Integer`) instead of primitives (`int`). This may may be undesirable depending on the use case. One could think of having dedicated types such as `IntConfigValue` but that'd increase the API surface significantly.

keilw:
We saw that in Java SE 8 with `IntProvider`, etc. for primitives.

Emily-Jiang:
@gunnarmorling if the direct injection works without any shortcoming, I am happy for them to be included. I don't want to ignore dynamics when we make decisions, as it will come either in 1.0 or 1.1. We need to think the future. The last thing I want to do is to deprecate methods and put a lot of warnings to say it is not going to work for Dynamic config sources. 
Besides the injection support @Inject @ConfigProperty ("myname") T myvalue; does not work if the T is custom written class. The @Inject @ConfigProperty("myname") @ConfigValue(T) myValue works for all conditions. 

We can add the shortcut version based on request and feedback. It is nicer than deprecating methods.

gunnarmorling:
> deprecate methods and put a lot of warnings to say it is not going to work for Dynamic config sources

Injecting the value directly naturally will not make them observable. But as I said before, that should not stop us from supporting the simpler usage pattern for cases where that's not required. Also consider what Heiko said, `ConfigValue` may not be the best means to support updatability anyways; we may be better off using a listener scheme at a higher level than single properties. But that's all guesswork, we'll need to experiment with this before deciding.

> does not work if the T is custom written class. 

Why that?

OndrejM:
+1 for including direct injection, along with the option to use ConfigValue. There are no shortcomings and it makes devs life much simpler. If devs are not clever enough to consider ConfigValue, no one would save them before assigning ConfigValue.getValue() into a var and pass it to another object... Good API should discourage bad practices but should not stand in the way.

If we create all examples in the docs using ConfigValue, we can influence many people. If we only allow ConfigValue, we will only piss off many.
We may also add a warning into the docs that direct injection should be used only with @RequestScoped beans and other short scopes. And encourage to use ConfigValue in other conditions.

emilyjiang:
This is not guess work. ConfigValue was demonstrated by Deltaspike and we tried experiment and it does work. As for the listener, it is implementation details for the dynamic config source to notify the config changes. It is used in Archaius and maybe others. However, it should not be visible to the end user.

In order to support custom written injection, the implementation needs to supply a producer for the type. I don't think you can do a producer with return type T, can you?

OndrejM:
> @emilyjiang In order to support custom written injection, the implementation needs to supply a producer for the type. I don't think you can do a producer with return type T, can you?

Good point. But it is possible to define producers dynamically by a CDI extension, scanning all injection points of all CDI beans for types. This is what Tomitribe Sabot does: https://github.com/tomitribe/sabot/blob/master/sabot/src/main/java/org/tomitribe/sabot/ConfigurationExtension.java#L106

gunnarmorling:
> plus on the Config interface, we only have
> Optional getValue(String propertyName, Class propertyType);

Uff, just seeing this.

How do you envision handling of _mandatory_ values then? Do you suggest that people always need to access values like so if they want to ensure it's present:

    int myInt = config.getValue("com.example.myKey", Integer.class)
        .orElseThrow( () -> new IllegalStateException("Config property com.example.myKey not present" ) );

I really think there must be a method which raises an exception itself if the requested value isn't present. I.e. there should be `getValue()` and `getOptionalValue()`.

+1 for keeping the API lean, but let's not loose usability out of sight. Being forced to always deal with `Optional` if the requested value actually isn't, makes it very cumbersome to use.

struberg:
@OndrejM I still don't get the reason for Sabot. Look at DeltaSpike. You can even have a 

```Java
@Produces 
@ConfigProperty 
@Dependent 
public Object createConfig(InjectionPoint ip) {
  return config based on injection point.
}
```
Since nobody else is using this qualifier there is also no AmbiguousResolutionException. 
And this will work for all types where a `Converter` got registered.

Do I miss something?

struberg:
@gunnarmorling 
I like the idea to have 2 methods (one for mandatore, another for optional values).

Emily-Jiang:
This is a good point @gunnarmorling . We should have these two methods in the Config and ConfigValue. Are we settled on the CDI injection to just have @Inject @ConfigProperty ConfigValue<T> myValue; for the first release?

gunnarmorling:
> We should have these two methods in the Config and ConfigValue

Ok, cool. I'll send a PR for this change.

> Are we settled on the CDI injection to just have @Inject @ConfigProperty ConfigValue myValue; for the first release?

No, IMHO there should be support for direct injection for the reasons layed out above. You agreed to support both on `Config`, so why would you do it not consistently for CDI injection?

We can try and have a voting if you like, though to me it's pretty clear that most folks here prefer to support direct injection and `ConfigValue`.

struberg:
The ConfigValue stuff needs much more discussion. And all need to understand the 4 usecases.
Just @Injecting a ConfigValue is not gaining too much if we have no way to specify e.g. a cache time, fallbacks, etc. Probably do another hangout about this?

gunnarmorling:
> In non-CDI case, there is no direct lookup any types, rather than config.getValue(myprop, Class).

Sorry, I'm not following what you mean by that.

> I don't think this conversation get anywhere. Let's discuss more in the next hangout.

Sure, we can, but honestly that entire debate is very tiring. I'm not sure whether I can make the Hangout and we tend to have no documented outcomes off those, unlike written communication which gets properly logged. Hence I'm no big fan of it.

On the matter itself, it's still unclear to me why you are against it. There's clearly a use case (easily accessible, non-modifiable values) and you are primarily arguing against it based on a feature (updatable properties) which we all agreed should not be part of 1.0. So we really should be put this aside and stay focused on the things we actually want to support in 1.0. Of course we should keep the future in consideration but as said before there is no reason to deprecate direct injection once `ConfigValue` is there.

Can you try and clearly describe why you are against it? Please also consider that others expressed their preference to have this, so I kinda wished we could move on. Thanks!

api/src/main/java/io/microprofile/config/inject/ConfigProperty.java
OndrejM: > @Emily-Jiang: I recommend to use classname.fieldName for the default.

It is already stated in the javadoc as you recommended: "`class_name` is the simple name of the class being injected to". Is that OK or am I missing something?

OndrejM:
This PR is not only proposing injection using ConfigValue. Since the discussion about direct injection deserves its own thread, I create another PR #86 - please continue the discussion there.

In this issue, we should only discuss if this PR can be merged. Direct injection can be merged separately, once we have a common agreement.

gunnarmorling:
> In this issue, we should only discuss if this PR can be merged. Direct injection can be merged separately, once we have a common agreement.

Ok. My vote then is to not merge this PR as `ConfigValue` in the proposed form doesn't provide any value add. Instead the direct injection PR should be the basis for further discussion and be merged eventually. Let's introduce `ConfigValue` once it actually provides a benefit.

OndrejM:
> @gunnarmorling ConfigValue in the proposed form doesn't provide any value add

I think there is a value in `ConfigValue`, unless we also support dynamic direct injection using `Instance`. With plain injection like `@Inject @ConfigProperty String value`, it is not possible to use optional config properties and check if they exist - the injection would simply fail. It makes sense to give the developers ways to access both mandatory and optional config values.

gunnarmorling:
> the injection would simply fail

I see; so one cannot inject `null` via CDI?

> it is not possible to use optional config properties and check if they exist

I see where you are coming from. I had in mind to use `Optional` for that. But if you are reluctant to use that as a field type (indeed it's not meant for that originally by the JDK team), then yes, we'd have to roll our own.

To me this shows that the interface-based approach me and you had suggested is generally superior over injecting values (or `ConfigValue`) directly:

* it'd allow to use `Optional` as a return value instead of inventing our own
* you'd always get the latest value (helping a lot with the "updatable value" story down the road)
* you don't need non-final fields
* it provides intuitive handling of default values via default methods
* it allows easy integration with Bean Validation and other techs

Personally, I think it's a mistake to give up on this in the light of all these advantages and stick to the value injection approach.

struberg:
you **can** inject null, but only if the scope is @Dependant. 
Which is actually the only CDI scope you can use anyway.
Because any @NormalScoped would result in a proxy - which is not possible for native types, final classes and classes without an non-private default ct.

OndrejM:
Thanks, @struberg, for clarification. I just remember that Sabot gives deployment error if a prop is not defined, but that was possibly a business decision, preventing app deployment if the configuration is missing.

I think that we should strive to get config 1.0 version as soon as possible, therefore I wouldn't push for this PR into 1.0. But maybe we can still have config 1.1 before MP 1.1, having some of the enhancements like this catch the MP 1.1 train. I'm strongly in favor to have both direct injection and the interface-based approach in the config spec, but not necessarily in the first version. The solution with ConfigValue provides a good enough solution for all usecases for the time being.

Emily-Jiang:
+1 on ConfigValue, which solves all use cases. It is the interface based approach (look at the versions I put in my pr). We should push this in and move on the discussion on other issues.

struberg:
We did not yet discuss the use cases of ConfigValue but you want to push this in?
I don't get it I fear.
I would just go for direct injection in config-1.0 and only add the dynamic support later. 
Or we finally pick up the PR from about 5 months ago where I proposed the fully featured ConfigValue. But doing a half baken implementation is meh.

Emily-Jiang:
@struberg ConfigValue should just be read as a proxy. Please don't compare with the original version. It is just a name.

struberg:
But if it's just a proxy (and not providing any further features) then we should use javax.inject.Provider&lt;T&gt; instead of introducing a new class.

Think back to where we started: @OndrejM wanted to use a Provider initially but people feared that it might clash with ConfigValue (the original one). In a discussion I mentioned that this is very easy to solve as ConfigValue might later implement Provider<T> anyway so there is no conflict. This class wasn't meant to be introduced right now.

gunnarmorling:
> +1 on ConfigValue, which solves all use cases. It is the interface based approach 

No, that's not what we meant. "interface based approach" referred to the OWNER style approach of injecting interfaces with methods returning config values instead of injecting plain config values.

> But if it's just a proxy (and not providing any further features) then we should use javax.inject.Provider<T> instead of introducing a new class.

Agreed. If we can do everything with `Provider` which can be done with `ConfigValue` in the proposed form, then let's use that existing vehicle.

OndrejM:
> Agreed. If we can do everything with Provider which can be done with ConfigValue in the proposed form, then let's use that existing vehicle.

+1. 
In my original proposal, there was no ConfigValue, only direct injection and dynamic injection using Provider/Instance (I prefer Instance, which provides additional method `isUnsatisfied()` to check whether the config value is defined). However, I was not sure whether it can be implemented in a satisfactory way, because I'm not sure if it's possible to create a producer for `Instance` or it is necessary that the CDI engine has to provided it and resolve its value dynamically using appropriate value producers. In the latter case, we would have little control over the `Instance` proxy itself and the value would be in fact static with deferred resolution (once resolved it wouldn't change).

Therefore I agreed to go with a custom wrapper like `ConfigValue`, which would be used instead of `Instance`, but we would have full control over how it is created and what it does.

We now have 2 competing ways to inject values - using the `ConfigValue` wrapper, allowing us full control, and standard CDI API (see the new PR #86 - it can be extended to define how `Provider`/`Intance` injection works). The standard CDI API way is limited but easy to grasp and use.

If I could choose, I'd support both and let users choose themselves what is better in their case. Supporting all shouldn't be much confusing - both cases are quite straightforward to me. Even supporting the annotated interface approach later (past 1.0) makes sense to me and adds very little additional burden or confusion.

Let's discuss both solutions a bit during hangout - I would only block one of the solutions if there are serious limitations that we cannot solve in the scope of config v1.0.

struberg:
With applying PR #42 is there anything left to do for now? 
It seems we managed to get all the functionality in by using @ConfigProperty + Provider<T>, isn't?
Anything I miss?

OndrejM:
After merging #86 (which uses Provider/Instance for dynamic access to config values), injecting a ConfigValue proxy is no longer necessary.
Looking closely on this PR, it only adds injection of the ConfigValue proxy and test setup classes, which are copied and already merged in via #86. After removing ConfigValue from this PR, there is nothing to merge.

Therefore I'm closing this PR.

Issue #13, opened 2017-01-17
ConfigProvider contextuality is wrong


rmannibucau:
ConfigProvider has 1 big issues in the current design:

-> it handles the lookup which is wrong, in particular in the microprofile context which has CDI has a strong backbone

Also note that - assuming it is a valid design - it assumes there is a context which is a classloader which is wrong in several java applications so key should be generic probably (interface with only hashCode() and equals()?)

Concretely the config API should just allow to:
1. build a config (with sources, converters, accessors, autoloaded from SPI or manually built for libraries for instance)
2. provide a default config instance in CDI context (@ApplicationScoped)
3. if considered relevant provide a bootstrap config (the default one) to let configure bootstrap components like CDI extensions (this can require some CDI work to ensure it is not hacky but technically there is no blocker, in particular in microprofile context)

In short: there is no point but drawbacks to re-implement what CDI already does which is the contextuality of an instance (Config there).

To make it more obvious: this issue is about removing any contextuality/lookup feature from the config project and let it be part of the container/IoC integration only.

Emily-Jiang:
The CDI support will come later in the next release. By the way, the config instance needs to be provided by a producer as a config instance is per classloader. By the way, the Config design is trying to align with Config JSR proposal and we will put the CDI integration shortly after the 1.0 release. We had a long discussion on the thread [https://groups.google.com/forum/#!topic/microprofile/VJEEAOsVj5E[126-150]](url)

rmannibucau:
@Emily-Jiang no, the producer is not needed? At least there is no constraint at all to require it. Also the "per classloader" rule is wrong and depends the environment. The Config JSR as well doesn't need any contextuality since the IoC or container (OSGi for instance) will handle it.

So concretely there is not a single reason to implement such logic in a config solution, and there is no point having a classloader based or lookup solution in a config solution.

Please revise these statements which are not backed by anything concrete.

struberg:
In my spec document and API JavaDoc I only say 'per Application'.

It's totally up to the implementation whether it uses a ThreadLocal (as in your sample), a WeakMap<ClassLoader, ..>, an OSGi BundleActivator, or even a CDI Extension (also automatically 1 per Application). 

The API does _not_ mandate to use the ClassLoader (except to 'identify' the Application). All the rest is a detail of the underlying implementation and should not affect the API.

rmannibucau:
[...]

PR #57, opened 2017-02-15
basic refactoring to reflect issue #13 point

rmannibucau:
Here are the changes:

1. Config should have a close method which is unlinked to the contextuality - think unmanaged case. It can belong to provider too but in term of user API it looks more friendly this way IMHO
2. ConfigProviderResolver owns the "mapping" between a classloader and a config and reflect a light Map API for that with static utility for global lookups (personally i would have dropped it from the spec and wait to see if users don't have enough of common IoCs but seems a blocker for Emily and Mark)
3. Keep ConfigProvider as the API link between the API/user code and implementation (to load the builder) *only*

I'm not yet a fan of ConfigProviderResolver and wonder if we shouldn't replace the ClassLoader by a Key which would be an interface but let's keep it for next thought round, there are already enough there for now ;).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ tck/bin/
 .project
 build/
 .classpath
+.idea
+*.iml
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -45,7 +45,7 @@ import org.eclipse.microprofile.config.spi.ConfigSource;
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  *
  */
-public interface Config {
+public interface Config extends AutoCloseable {
     /**
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
@@ -87,5 +87,11 @@ public interface Config {
      * @return all currently registered {@link ConfigSource configsources}
      */
     Iterable<ConfigSource> getConfigSources();
+
+    /**
+     * Let the implementation release any resource it can need and then the config is no more usable.
+     */
+    @Override
+    void close();
 
 }
--- a/api/src/main/java/org/eclipse/microprofile/config/ConfigProvider.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/ConfigProvider.java
@@ -23,10 +23,17 @@ import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
 import org.eclipse.microprofile.config.spi.ConfigSource;
 import org.eclipse.microprofile.config.spi.Converter;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Iterator;
+import java.util.ServiceLoader;
+
 /**
  * <p>
  * This is the central class to access a {@link Config}.
  * </p>
+ *
+ * <p>It enables to access a {@link ConfigBuilder} to build a ready to use {@link Config} instance.</p>
  *
  * <p>
  * A {@link Config} contains the configuration for a certain situation. That
@@ -34,11 +41,6 @@ import org.eclipse.microprofile.config.spi.Converter;
  * created Configuration
  * </p>
  *
- * <p>
- * The default usage is to use {@link #getConfig()} to automatically pick up the
- * 'Configuration' for the Thread Context ClassLoader (See
- * {@link Thread#getContextClassLoader()}).
- * </p>
  *
  * <p>
  * A 'Configuration' consists of the information collected from the registered {@link ConfigSource ConfigSources}.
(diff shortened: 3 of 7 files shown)


Emily-Jiang:
I found this refactoring is rather confusing. Now, you ask the implementor to supply both ConfigBuilder and ConfigProviderResolver using the loader pattern. This creates more work for OSGi environment. I can only see the change is from ConfigProvider.getConfig to ConfigProviderResolver.lookup. This changes cause more confusion rather than helping.

rmannibucau:
@Emily-Jiang always has been the case except builder was not a spi (well it was originally then when imported there it got dropped probably). For any environment migration you only implement the resolver, nothing else (really) which is the only not portable part of an implementation.

Emily-Jiang:
I don't understand what you are trying to improve. Everyone has different perception. I failed to see anything wrong with the current approach. What you have got is just another variation.

Emily-Jiang:
please list the problems and explain why your solution is better.

rmannibucau:
Fundamental issue is config api has not enough knowledge to get the storage right and therefore this shouldnt be a first citizen of the solution. There is no environment where you dont have a good storage (by definition) so this is by design an extension/bonus of config project.

Therefore it has to be dropped or made obvious and isolated (resolver).

struberg:
ad 1.

> close. It can belong to provider too but in term of user API it looks more friendly this way IMHO

Yes, we need such a functionality. But having it on Config would give the user the impression that he should use it. 
But actually closing the config is up to the config integration. That means if you just say `ConfigProvider#getConfig()` then you do **not** need to call close/release!
This is really only needed if you want to do some deep-scale integration into a not out-of-the-box supported environment.

ad 2.) 
In my original proposal there has been only one class (ConfigProvider) which corporates all the functionality. We did split this out into the more 'internal' parts (My internal SPI interface became the current `ConfigProviderResolver`). Just to get it out of the eyes of the casual user. 
There are multiple arguments for having a factory accessor. The main ones are that you cannot portably access it via a DI framework before the DI framework is booted and that we do not like to require a specific DI framework to broaden adoption.

> Fundamental issue is config api has not enough knowledge to get the storage right 
That's why only the API is fixed and the implementation is totally up to adopt to the underlying environment. E.g. The implementations could be different depending whether to run on SE, a fat EE iron or OSGi. The 'consumer' still only needs to use `ConfigProvider#getConfig()` and be done.

rmannibucau:
1. you need close for not managed instances so until this feature is dropped you need it;
2. not when i did my PR on your branch the the mentionned reason there.

Concretely if you extract the locator code which is a transversal concern then you start to reimplement an IoC which is a shame for any application which already has one (all). There is also no techincal justification (even for CDI which doesn't need it for extensions as shown by multiple PoC we did Mark).

If config-api wants to solve the IoC part of the code then microprofile should drop CDI from its stack and just use this solution no?

Config-api scope should IMHO just be:

- Config value reading (raw values or with converters): Config, ConfigSource, ConfigConverter
- Config building logic (ConfigProvider -> ConfigBuilder)

And provide some integrations based on that which would create a default contextual config instance reusing the IoC sources and converters when possible:

- CDI
- Spring
- ...

Whatever you do you will not solve the chikcen egg issue with a new contextual solution: getConfig() sill prevents to use JPA or CDI in sources in extensions so you move the issue but you hit the same just after so globally you don't solve any problem with introducing a new contextuality in the api.

Emily-Jiang:
The close should be implementation details. The implementation is free to add this in. It is risky to restructure the current apis towards the end of the cycle as the solution is a different variation.

struberg:
A way to programmatically close a ConfigSource is mandatorily required to not create mem leaks and free up unused resources. 

There are 2 possible ways:
1. ConfigSource#close() + blow up with an IllegalStateException for managed ConfigSources
1. ConfigProvider#unregister(Config)

I like the unregister better because it is not directly in front of the eyes of a casual user but more an internal detail for power users.

Emily-Jiang:
I am going to close this PR as the code base has changed. Do we need to re-discuss this issue? We are locking down the current codebase for config 1.0 release. Please reopen another issue to continue the discussion in the config 1.1 release.

rmannibucau:
Well the design issue is not solved AFAIK and I still find it sad it hits any release but don't have enough time to work on that and it is not that impacting while it doesn't affect EE.

OndrejM:
I create an issue #136 for the close method to track it in future iterations.

Emily-Jiang:
@OndrejM From the discussion, both config and configSource should have a close method. @rmannibucau will adding close() be your only concern?

rmannibucau:
This and moving all the contextuality management in a dedicated class (resolver) instead of mixing it with the config itself.

Emily-Jiang:
Can you modify the issue #136 accordingly? We can continue the discussion there and hope to find a solution for the issue and get it into Config 1.1. I plan to close this PR. Please shout if you think otherwise.

PR #70, opened 2017-03-02
Alternative API to support ordering config sources using comparators

OndrejM:
This pull request created on behalf of @atsticks  to support the discussion about ordering solutions.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/ConfigProvider.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/ConfigProvider.java
@@ -23,6 +23,8 @@ import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
 import org.eclipse.microprofile.config.spi.ConfigSource;
 import org.eclipse.microprofile.config.spi.Converter;
 
+import java.util.Comparator;
+
 /**
  * <p>
  * This is the central class to access a {@link Config}.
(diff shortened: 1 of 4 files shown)


OndrejM:
@atsticks, could you please one or more test cases to show, how to use comparators to resolve conflicting keys from multiple sources? There's already `CustomConfigSourceTest`, it could be used as a template for adding multiple sources.

api/src/main/java/org/eclipse/microprofile/config/ConfigProvider.java
```diff
@@ -73,6 +75,16 @@ private ConfigProvider() {
     }
 
     /**
+     * Sets the default config source comparator to define the ordering of auto-discovered
+     * config sources.
+     * @param comparator the new comparator to be used, not null.
+     * @return the previous intance, never null.
+     */
+    public static Comparator<ConfigSource> setDefaultComparator(Comparator<ConfigSource> comparator){
```
OndrejM: The comparator can easily sort config sources, if it can identify them, e.g. by class name (all are known in advance).
What else can a comparator rely on in ordering, if it doesn't know anything about the available config sources? 

Some ideas are explained [here](https://github.com/eclipse/microprofile-config/issues/65#issuecomment-282554858) but none is explained in this PR. We need to specify the behavior if the default comparator, if no custom comparator is defined. Otherwise a portable application would be forced to define a comparator.

Apart from checking for an ordinal defined within the config source values or an optional method, we can also define ordering groups, which could cover most use cases (like "Application", "Container", "Environment", "Remote", ...), to make it easier to infer assumptions about config sources.

struberg:
@OndrejM do you like to close this PR in hindsight of our last hangout?

Emily-Jiang:
close this based on the hangout on 2nd March - keep getOrdinal() not using comparator.

PR #71, opened 2017-03-02
A question about config sources API

OndrejM:


api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -31,7 +31,7 @@
  * <p>A ConfigSource will get picked up via the
  * {@link java.util.ServiceLoader} mechanism and must get registered via
  * META-INF/services/javax.config.spi.ConfigSource</p>
- * The other custom config source can be added programmatically via {@link org.eclipse.microprofile.config.ConfigProvider}.
+ * The other custom config source can be added programmatically via {@link org.eclipse.microprofile.config.ConfigProvider}. << How?
```
OndrejM: It is not clear to me how a custom source can be added programmatically to existing sources?
AFAIK, only defining a completely new config is possible, but that wouldn't change the config returned by `ConfigPrvider.getConfig()`. Is there any way to change this config programmatically? Or only via service loader?

Emily-Jiang: The config source is added via ConfigProvider.getBuilder().withSources(...). This is not to change existing sources. config is readonly and its value should not be changed.

OndrejM: So is my understanding correct that `ConfigProvider.getConfig()` will always return the same set of sources once app is booted?
Will `ConfigProvider.getBuilder().withSources(...)` define an alternative config, which will not affect config returned by `ConfigProvider.getConfig()` in any way?

Emily-Jiang: ConfigProvider.getConfig() is managed and always return the same one for the same classloader. ConfigProvider.getBuilder().withSoruces(...) creates a new instance each time, hence unmanaged. If you call ConfigProvider.setConfig(), which will store the config. Later on, when you call ConfigProvider.getConfig(), which will return you the one just stored. Remember, setConfig() is only successful if there is no config for this classloadder was stored. Otherwise, you will get exception. ConfigProvider.releaseConfig() will release the config.

OndrejM: I understand. I think that the javadoc should be improved to avoid confusion because it mixes information about 2 different ways to access them, which cannot be mixed. It should clearly say that a ConfigSource is registered either by the default config using service loader, or registered programmatically to a new config created by a builder. I would also move the details about service loader to ConfigProvider, because it is not directly related to the interface itself, rather to the ConfigProvider that loads it.

OndrejM: @Emily-Jiang I'm closing this PR - question answered. Please don't forget to improve the javadoc ;-)

PR #77, opened 2017-03-03
Splitting up Config#getValue() into getValue() and getOptionalValue()

gunnarmorling:
As discussed in the CDI PR, I've added `getOptionalValue()`. Not sure what I need to do from an administrative PoV to get it merged? Happy about any pointers. Thanks!


gunnarmorling:
Le sigh, being required to sign off my own commit seems weird. Not sure what this adds, but I did it. On the CLA, I thought the one of my employer (Red Hat) is sufficient?

struberg:
heh, being through the same right now ;)

Emily-Jiang:
If the ip-validation is not fixed, the pr will not get merged.

gunnarmorling:
Understood; what do I need to do though in order to get it fixed?

gunnarmorling:
Ok, done.

gunnarmorling:
Can this one be merged?

struberg:
Thanks @gunnarmorling 

There was kind of overlapping work done in parallel due to us being to slow reviewing and applying changes - sorry for that! 

We now have the following methods
```Java
    <T> T getValue(String propertyName, Class<T> propertyType);
    Optional<String> getValue(String propertyName);
    String getRawString(String propertyName);
```

This situation did occur since we had parallel work by @Emily-Jiang targeting a similar goal. But now it seems to me that we kind of have the functionality a bit duplicated and doesn't really fit together, isn't it?

The methods with both key + type are consistent: getValue (no Optional) vs getOptionalValue (with Optional) 

But the methods with just the String parameter seem to not fit into this picture: 
- getValue returns an Optional 
- getRawString not. 

This feels not in sync with the other methods.

I also thought about getRawString a bit.
getRawString is only needed IF there is a special Converter for String type registered. Otherwise getValue and getRawString will always return the same result. I think this is really a pathological situation. And while the raw string information is _sometimes_ needed, one could still take the values directly from the ConfigSources in that case. So could we probably live without this and move to the methods:

- String getValue(String key) 
- Optional&lt;String&gt; getOptionalValue(String key) 

@Emily-Jiang @gunnarmorling @OndrejM wdyt?

gunnarmorling:
Do you indeed mean

    String getValue(String key)
    Optional<String> getOptionalValue(String key)

Or rather

    <T> T getValue(String key, Class<T> type)
    <T> Optional<T> getOptionalValue(String key, Class<T> type)

That would make sense IMHO (i.e. only keeping methods returning `T`, keeping none returning `String`). +1 for removing `getRawString()`. Providers can experiment with it if needed.

struberg:
Yes, currently the method-pair with parameters key + type and the method-pair without the type parameter do not really fit together.
getValue with key+type returns the native value, getValue with just the key returns an Optional. That's not too intuitive imo.

gunnarmorling:
Do we need the methods without the type parameter at all?

struberg:
hmmm, well, in the end it's just a convenience method now.

In my experience most business code is going to use the injection way. Despite that we **must** support programmatic access as well. This is often needed if you do some frameworks etc. But it might well be that it's enough if people can write 
```Java
String myCfg = config.getValue("someprj.somekey", String.class);
```

Might be worth to go for it!

@Emily-Jiang @OndrejM @viktorklang @gunnarmorling wdyt?

gunnarmorling:
That would be my preference, esp. considering that we may be required to
loosen the SPI contract to not only work with a map of string -> string
possibly in the future. So the less places the string nature is encoded,
the better imho.

Emily-Jiang:
I'm fine without the convenient method. It does introduce confusions.

struberg:
Great, @gunnarmorling  do you like to ship an additional PR?

viktorklang:
I agree with not specialcasing String.class

Issue #62, opened 2017-02-22
ConfigProvider is a bit crowded


struberg:
We should probably move a few bits from the ConfigProvider out of the eyes of the casual user.
We could e.g. move the following methods to ConfigProviderResolver:

- registerConfig
- releaseConfig
- ConfigBuilder inner interface

struberg:
Any comments about this idea?

gunnarmorling:
I looked at the PR, but I couldn't fully figure how you envision all the pieces to be used. Maybe you could sketch some code using them to clarify the picture?

struberg:
You mean the features itself or just the refactoring?

I just moved the methods from ConfigProvider to ConfigProviderResolver.
So the usage would be ConfigProviderResolver.getInstance().setConfig() etc.

gunnarmorling:
The purpose of `ConfigProviderResolver` isn't quite clear to me.

Its comment says "Service provider for ConfigProviderResolver". Should it be "Service provider for ConfigProvider"?

struberg:
@gunnarmorling ConfigProviderResolver was initially a small internal interface delegating 1:1 to the internal implementation. Please feel free to suggest a better name for that class in a separate ticket.

PR #79, opened 2017-03-03
#62 configprovider rework

struberg:
move the complex and seldom used parts away from ConfigProvider.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
@@ -45,17 +45,27 @@ public interface ConfigSource {
     Map<String, String> getProperties();
 
     /**
-    * Return the ordinal for this config source. The higher the more important. If a property is specified in multiple config sources, the value
-    * in the config source with the highest ordinal will be used.
-    * The ordinal for the default config sources:
-    * <ol>
-    *  <li>System properties (ordinal=400)</li>
-    *  <li>Environment properties (ordinal=300)</li>
-    *  <li>/META-INF/microprofile-config.properties (ordinal=100)</li>
-    * </ol>
-    * @return the ordinal value
-    */
-    int getOrdinal();
+     * Return the ordinal for this config source. The higher the more important. If a property is specified in multiple config sources, the value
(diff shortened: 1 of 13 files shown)


api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -69,6 +79,6 @@
      *
      * @return the unique 'id' of the configuration source, e.g. 'property-file mylocation/myproperty.properties'
```
Emily-Jiang: The javadoc needs to be updated.

struberg:
> I prefer not to change this class as it hides the ConfigBuilder, which is better not to be exposed too widely.

That exactly was the reason why I moved all those methods and classes from the ConfigProvider (user facing class) to the spi module and thus the integration layerl.

Emily-Jiang:
In my view, making the ConfigBuilder a standalone classing is making it more visible rather than hiding it.

struberg:
The point is to make ConfigProvider easier to grok while moving more complex use cases away from the cacual developer.

Btw, I need to fix the Copyright. It should not say "Copyright Mark Struberg and others", but "Copyright Romain Manni-Bucau and others".
Romain clearly wrote the initial version with a pull request to my original proposal on July 20th, 2016:
https://github.com/struberg/javaConfig/pull/2/commits/0945b23cbf9dadb75fb97551837e6ce3d26496f6
https://github.com/struberg/javaConfig/pull/2/commits/84112ecb8fc693ff55d0896b19dfa2a8a50b35ef

OndrejM:
I see your point, @Emily-Jiang ,but I think ConfigBuilder really belongs to SPI rather than to the API package and I agree with @struberg. We could try hiding it even within the SPI package, e.g. into ConfigProviderResolver, but that's a little too unnatural.

struberg:
> We could try hiding it even within the SPI package, e.g. into ConfigProviderResolver, 
> but that's a little too unnatural.

yes @OndrejM , that was my feeling as well. ConfigProviderResolver used to be an inner class of ConfigProvider (originally ConfigProvider#SPI). Moving the ConfigBuilder to this place seems a bit too much.

PR #83, opened 2017-03-03
Improve Javadoc etc

Emily-Jiang:


gunnarmorling:
Hey Emily, could you commit with your e-mail address you are using here on GitHub, too? That way the commits will be properly attributed to you in the contribution history.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -69,13 +69,11 @@
...
-     * @return The resolved property value as a String-Optional, which will bypass converters.
+     * @return The resolved property value as a String, which will bypass any converters.
      *
-     * @throws IllegalArgumentException
-     *             is thrown if the propertyName maps to an object that is not a
-     *             String.
+     * 
      */
-    Optional<String> getString(String propertyName);
+    String getRawString(String propertyName);
```
gunnarmorling: What's the rationale for changing it to not return `Optional` any more?

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,50 @@
...
+ * @author <a href="mailto:rsmeral@jboss.org">Ron Smeral</a>
+ * @author Ondrej Mihalyi
+ * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
+ * 
+ * @param <T> Type of the configuration value
+ */
+
+import java.util.Optional;
+
+public interface ConfigValue<T> {
```
gunnarmorling: If these are all the methods on `ConfigValue`, I don't see good reason for adding it at this point. You'll get the same functionality (and easier to use) when directly injecting values into beans. We should introduce `ConfigValue` if it provides additional value.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,50 @@
...
+ * e.g. @Inject @ConfigProperty("myProp") ConfigValue<String> myValue;
+ * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
+ * @author <a href="mailto:rsmeral@jboss.org">Ron Smeral</a>
+ * @author Ondrej Mihalyi
+ * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
+ * 
+ * @param <T> Type of the configuration value
+ */
+
+import java.util.Optional;
```
gunnarmorling: The import should go before the JavaDoc.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -20,18 +20,18 @@
...
  * <li>/META-INF/microprofile-config.properties (ordinal=100)</li>
  * </ol>
  * 
- * <p>A ConfigSource will get picked up via the
- * {@link java.util.ServiceLoader} mechanism and must get registered via
+ * <p>ConfigSource will get picked up via the
+ * {@link java.util.ServiceLoader} mechanism and and can be registered via
  * META-INF/services/javax.config.spi.ConfigSource</p>
- * The other custom config source can be added programmatically via {@link org.eclipse.microprofile.config.ConfigProvider}.
+ * Other custom config source can be added programmatically via {@link org.eclipse.microprofile.config.ConfigProvider}.
```
OndrejM: I would clarify that it is not possible to add a custom config source programmatically into the default config retrieved by ConfigProvider.getConfig(). It is only possible when building a completely new config instance.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -24,7 +24,7 @@
  * <p>A Converter can specify a {@link javax.annotation.Priority}.
```
OndrejM: For supporting @Priority, we should include common annotations JSR (JSR-250) into MP, thoughts?

api/src/main/java/org/eclipse/microprofile/config/Config.java
OndrejM: I assume that the method is intended for debugging rather than for business logic, therefore it is counter-productive to wrap the result in an Optional. It would make even sense to me to rename the method to something much more complicated to discourage using the method in normal business logic :) But I'm serious - far too many people would use it for normal code if its name stays like this.

gunnarmorling: > it is counter-productive

Why is that? The value may be present or not, which is exactly the semantics of `Optional`. `Optional` was introduced to represent an optional return value; let's use it instead of requiring people to think about whether it can be null or not. `Optional` conveys that clearly.

gunnarmorling: That said I'm unsure what's the use case for this method in the first place? Should it actually be present here if we are concerned about it being used? Raw values can be obtained from config sources so we may leave it out here in the first round and see whether there's actual demand for it?

Btw. the "raw value" isn't necessarily a String, thinking about a JSON data source for instance. I.e. I'm starting to get doubts whether it's a good idea to use String as the type for representing raw values (here, but more importantly on `ConfigSource`).

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
gunnarmorling: I see your point, but my experiences with 250 are very mixed.

There are several conflicting dependencies with (parts of) its types out there, some are even in the JDK itself. So it's very prone to cause classpath conflicts, i.e. users will end up with the same types being pulled in multiple times from several dependencies. Which is esp. an issue when thinking about Java 9 / Jigsaw. Hence I'd try to avoid that dependency as much as possible.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
OndrejM: I think that adding ConfigValue in this PR doesn't make sense at all. It should be either in a separate PR, or coupled with the CDI PR (#42), where I already added it. I think that for a start, it makes sense to have it even with 2 methods, we can add more methods later. 

@Emily-Jiang I suggest that you create a PR with ConfigValue against my [repo](https://github.com/OndrejM/microprofile-config/tree/CDI_API_ondrejm), so that I update the existing PR #42.

I'd discuss direct injection separately, I created PR #86 for this reason.

api/src/main/java/org/eclipse/microprofile/config/Config.java
OndrejM: IIRC, it is really meant for debugging purposes to see what is the final value over all config sources (even if multiple sources provide it), but before it goes into a converter. Since converters work with String, it's natural to return raw value as a String.

Maybe we should leave this method out for now, and only suggest that implementations provide vendor-specific means to access the final raw value, provided that it is only for debugging as I assume.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
OndrejM: I agree, but I don't see how we get `@Priority` annotation into MicroProfile then. Which API artifact would bring it? In the current state, there's no way to build an app with `@Priority` with just MP dependencies.

api/src/main/java/org/eclipse/microprofile/config/Config.java
gunnarmorling: > Since converters work with String, it's natural to return raw value as a String.

That's what I meant above, I doubt basing it on String is the right thing to do. I've opened #87 to discuss this separately.

gunnarmorling:
The question is, does it have to be pulled in by the API itself? I don't
think there is any dependency from the API requiring it. Instead it can be
left to providers or the application using the mP.

OndrejM:
> @gunnarmorling  The question is, does it have to be pulled in by the API itself?

I get it now. I was not sure if makes sense to instruct anyone to use `@Priority` if it is not provided by the container. But since it's an edge case and it is possible to provide the annotation within the application, I'm fine to leave the spec as is.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
Emily-Jiang: @OndrejM I'll create this against your PR #42 and merge in.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
Emily-Jiang: No, it is possible. The calling of ConfigProvider.getConfig() should trigger the builder to add default configure sources and autodiscovered configure sources. Otherwise, the ConfigProvider.getConfig is not useful. Thoughts? We should clarify this.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
Emily-Jiang: By the way, CDI has dependencies on JSR-250 for its interceptors/decorators/alternatives. Since Config has dependencies on CDI, JSR-250 dependency may not need to be discussed.

api/src/main/java/org/eclipse/microprofile/config/Config.java
Emily-Jiang: This method is useful for debugging purpose as commented by @OndrejM !

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
Emily-Jiang: ConfigValue can do the injection for the custom class type T. You can do 
@Inject @ConfigProperty("myPropName") ConfigValue<MyType> propValue;

also the injection can cope with dynamic changes.

Emily-Jiang:
Wondering why my sign off failed ip-validation check? Any suggestions?

gunnarmorling:
You have commits with two different e-mails in this PR, one of which
doesn't have the ECA on record.

I think the best would be if you do your ECA with the e-mail address you
have on GitHub (you also can change the email address in the ECA tool) and
squash all commits into a single one, using that e-mail address.

That way the ECA check should be happy and the commits will properly be
attributed to your account in the GH history and stats.

struberg:
+1, but please note that we should **never** squash history which contains commits of different persons as that would destroy code provenance. So please only squash your own commits.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
OndrejM: See, that's why I recommended added more test code to demonstrate the usage. This part is not very clear to me. How can I change what is returned by ConfigProvider.getConfig() programmatically? I mean, can I somehow hook into the builder used to build that config, without adding config sources via the service locator? Would this work: `ConfigProvider.getBuilder().withSources(myCustomSource); Config configWithMyCustomSource = ConfigProvider.getConfig();`  ?

api/src/main/java/org/eclipse/microprofile/config/Config.java
OndrejM: Maybe we could avoid the getRawString method in config-1.0 and rethink if we need to standardize this or leave it up to the implementations?

OndrejM:
What I do to pass eclipse IP checks is that I rebase my local branch interactively, and then re-commit each revision with the correct user (email) and sign-off. Then I force push my changes. Not ideal either, but the history is preserved. I don't know how that affects comments in the PR.

Emily-Jiang:
thanks @gunnarmorling @struberg @OndrejM! Finally worked out! I ended up abandoned some commits and redo it under a different PR. I'm going to merge this javadoc changes.

api/src/main/java/org/eclipse/microprofile/config/Config.java
keilw: `Optional` as mentioned in recent calls is quite ugly especially when you try to combine it with type-safety you quickly stretch it beyond its limit. 
Typesafe Config just to give one example has `String getString(String path);` and getters for most other base types including primitives. `ConfigObject getObject(String path);` returns a `ConfigObject` which is a specialized `ConfigValue` And `Object getAnyRef(String path);` is the "raw" equivalent simply returning the underlying value as `Object`. No `Optional`, although the library is now strongly tailored to Java SE 8, e.g. HOCON dealing with java.time values like `LocalDate` or `Duration`.

Issue #84, opened 2017-03-04
Add unwrap() methods to Config and ConfigProvider.Builder


gunnarmorling:
There should be a method `<T> T unwrap(Class<T>)` on `Config` and `ConfigProvider.Builder`, allowing to access additional functionality offered by specific implementations.

Emily-Jiang:
See more discussion on #90

OndrejM:
I suggest discussing this across all API features - I'd like to have a unified pattern across all MP APIs to access vendor-specific features. Having `unwrap` methods only in Config and not in other features where they would also make sense is not worth it.

Emily-Jiang:
I reactivate this as I think potentially we can use the unwrap to provide more debug messages, e.g. config key name, which source it is from etc. It is extensible as well. Thoughts?

Emily-Jiang:
This one is on hold until we have a use case appear.

radcortez:
I think this would be useful to add. The use case is clear. It is a safe way for the user to access the underlying implementation and use implementation specific API's, without an explicit cast that may not work.

PR #85, opened 2017-03-04
#84 Adding Config#unwrap() and ConfigProvider.Builder#unwrap()

gunnarmorling:
Signed-off-by: Gunnar Morling <gunnar.morling@googlemail.com>

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -51,7 +51,7 @@ public interface Config {
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
      *
      * If this method gets used very often then consider to locally store the configured value.
-     * 
+     *
      * @param <T>
      *             the property type
      * @param propertyName
@@ -88,4 +88,15 @@ public interface Config {
      */
     Iterable<ConfigSource> getConfigSources();
 
+    /**
+     * Returns an instance of the specific class, allowing to access additional
+     * APIs offered by specific providers.
+     *
+     * @param type Class representing the type to unwrap to
+     * @param <T> The type to unwrap to
+     * @return An instance of the given type
+     * @throws IllegalArgumentException If the current provider does not
+     * support unwrapping to the given type
+     */
+    <T> T unwrap(Class<T> type);
 }
--- a/api/src/main/java/org/eclipse/microprofile/config/ConfigProvider.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/ConfigProvider.java
@@ -179,6 +179,18 @@ public final class ConfigProvider {
         ConfigBuilder withConverters(Converter<?>... converters);
 
         /**
+         * Returns an instance of the specific class, allowing to access additional
+         * APIs offered by specific providers.
+         *
+         * @param type Class representing the type to unwrap to
+         * @param <T> The type to unwrap to
+         * @return An instance of the given type
+         * @throws IllegalArgumentException If the current provider does not
+         * support unwrapping to the given type
+         */
+        <T> T unwrap(Class<T> type);
+
+        /**
          * Build the {@link Config} object.
          *
          * @return the Config object


gunnarmorling:
The use case is to provide a way for accessing extended provider-specific
functionality not offered by the spec'ed API.

It's a common pattern across multiple specs to offer such unwrap() method,
e.g. see EntityManager#unwrap() or ValidatorFactory#unwrap().

OndrejM:
I agree it's common in many JSRs. However, I'm not sure if it's necessary, at least for v1.0. It should always be possible to achieve the same with a static method, like `ProviderSpecificConfig.unwrap(config)`, or not?

gunnarmorling:
A static method is harder to find; having that method right on `Config` makes the user much more aware of the possibility and allows for fluent invocations. Providers will come up with additional functionality for sure (e.g. if we don't expose `getRawValue()`), so it makes sense to offer `unwrap()`, following an established pattern, right from the get-go.

struberg:
I only know JPA who has such a unwrap feature. But there it is because in EJBs you don't get the native EntityManager but a proxy which is provided by the container.

To access container specific extensions or a mp-config vendor you could simply do an upcast. 
Thus I also see no need for such a method. 
@gunnarmorling can you please explain which use case wouldn't work by just doing an upcast?

gunnarmorling:
Bean Validation has unwrap methods, [too](https://docs.oracle.com/javaee/7/api/javax/validation/ValidatorFactory.html#unwrap-java.lang.Class-).

> which use case wouldn't work by just doing an upcast?

Casting requires that the provider specific interface must implement the unwrapped contract, which limits its flexibility. E.g. a provider may offer several "facets" to unwrap to, each offering a set of methods belonging together in some way but not extending the unwrapped contract.

Also, if config is injected via CDI, can we guarantee that the  bean - if proxied - implements all provider-specific extension interfaces? Perhaps yes, but it seems to be another potential pitfall.

Finally, having an explicit `unwrap()` makes people more explicitly understand that it is a supported - yet non-portable - operation than when using a cast which may feel more "hacky" to some. That's a rather soft fact, I know :)

OndrejM:
> A static method is harder to find

I see the disadvantages. But I'd rather treat the requirement of `unwrap()` as optional new feature, and we want to fix the scope for 1.0, not add new features to it. Since a static method would do the same, having `unwrap()` is just a syntactic sugar and can be postponed to later versions. I'd postpone this discussion after 1.0. @gunnarmorling please create an issue and refer to this PR to track it.

struberg:
+1 for postponing this feature

gunnarmorling:
> Since a static method would do the same

It wouldn't, see my comment above about extensions not implementing the unwrapped contract.

Look, I'm not suggesting to add it for the addition itself, but because there's a clear use case for this. It's clearly defined, it has a use case, the change is prepared so I don't see why we should postpone it.

> +1 for postponing this feature

But why? You don't address any of my comments at all.

Taking a step back, I feel way too much energy is spent here to "fight" against proposals for whatever reasons instead of focusing energy on actual problems. Contributing to this group feels like an uphill battle constantly to me and right now I feel like quitting because it seems not like the best usage of my time. Sad panda :(

struberg:
> But why? You don't address any of my comments at all.

I thought I did: a standard upcast is perfectly fine to achieve the exact same. And the theoretical case with a proxy is also not of interest because
a.) we don't need any proxy at all
b.) CDI make subclass proxies of the concrete classes, so that would include the proprietary methods anyway.

Anything I missed?

gunnarmorling:
> a standard upcast is perfectly fine to achieve the exact same
> 
> Anything I missed?

Yes. See my comment above: 

> Casting requires that the provider specific interface must implement the unwrapped contract, which 
> limits its flexibility. E.g. a provider may offer several "facets" to unwrap to, each offering a set of 
> methods belonging together in some way but not extending the unwrapped contract.
> ...
> having an explicit unwrap() makes people more explicitly understand that it is a supported - yet 
> non-portable - operation than when using a cast

struberg:
Oki I see. What other class do you think is intended? 
For JPA we have a set of defined use cases (ofc + vendor specific ones). e.g. java.sql.Connection.

What interface do you think would make sense? I still don't get the use cases but happy to add such a feature if we find a good one. 

I would just like to avoid adding methods which turn out to not be of much use.
Because it's easy to add this unwrap method in 1.1 but it's hard to get rid of it later ;)

OndrejM:
> I still don't get the use cases but happy to add such a feature if we find a good one.
:+1: 
@gunnarmorling I'm not saying it's not a necessary/useful feature, but I'm also not sure about the usecase to support adding a new method now. I'd like to discuss later, but until we know WHY we would add this, I'd like to postpone it after 1.0

And I still don't get why a static method is so different, except it is not so obvious and fluent. You can do everything with a static method what you could do with an unwrap method. Let me compare:

With unwrap:
```
config.unwrap(ConfigExtension.class).refreshConfig();
```

With a static method:
```
ConfigExtension.unwrapConfig(config).refreshConfig();
```

If ConfigExtension is provided by the implementation, it can see internals and easily convert a standard API to impl-specific interface (either using cast, or calling an additional get method after cast or similar).

I'd like to people using the static method first, to see some real-life usecases, and then consider adding a common unwrap method. The reason I'm not much in favor of unwrap in the API is that it is still non-portable (you have to provide an impl-specific class as an argument), therefore why do a half-portable API in the first place? To be honest, I've been working with JPA a couple of years and never thought about using unwrap, unless in some cases during debugging. And a static method would work in all those scenarios as well.

OndrejM:
@gunnarmorling I'm very sad that you feel discouraged. I hope we can talk about it during the hangout and hopefully fix that. We all want to create an encouraging atmosphere, open to all ideas and discussions. But at the same time, I believe we want to create a well-crafted API, in a lean way, adding small well-thought pieces step by step. I believe it makes sense to postpone ideas until it is crystal clear that they are good enough or simply worse than other ideas. Especially in such a widely collaborative effort, without a spec lead who would have the final word on the decision.

Emily-Jiang:
Guys, let's talk about in today's call. By the way, I think Mark and I volunteered to lead the spec.

gunnarmorling:
> I still don't get the use cases but happy to add such a feature if we find a good one.

It's the purpose of this method to enable use cases which we don't foresee yet, so that's a tricky one. Maybe something like this:

    config.unwrap( ListenerRegistry.class ).addConfigChangeListener( ... );

In BV we were and are very glad to have this loophole, e.g. it allows to innovate in the RI, adding functionality explored that way in the next spec revision. Usages typically show up based on feature requests about shortcomings in the existing API.

> except it is not so obvious

This - i.e. an API easy to explore - would be reason enough for me, YMMV. Sure, a static method may work (once you found it), personally I find they destroy the flow.

Anyways, I had assumed that method would be common sense, in no way did I expect such lengthy debate. Knowing that, I would not have suggested it. I think the feature is useful, but not worth the amount of energy already poured into it.

gunnarmorling:
> ConfigExtension.unwrapConfig(config).refreshConfig();

Note it would rather be the following assuming there are multiple unwrap targets:

    ConfigExtension.unwrapConfig( config, ListenerRegistry.class ).refreshConfig();

gunnarmorling:
Ah, one thing I forgot: Unlike casts, unwrap() allows implementors to control the allowed types to unwrap to. I.e. you can allow unwrapping to a supported public API type of the implementation (say `ExtendedConfig`, typically an interface) but raise an exception when trying to unwrap to internal implementation types (say `ConfigImpl`). I.e. it helps to reduce chances for people to shoot into their own foot by casting into non-supported internal types.

OndrejM:
> Note it would rather be the following assuming there are multiple unwrap targets:
@gunnarmorling, I assume that you agree that such a static method would provide the same feature until we agree to include unwrap into the API. In that case, the unwrap method only provide syntactic sugar and is not necessary for initial releases of config, which IMHO should focus on enabling the important usecases first.

gunnarmorling:
Ondrej, as I pointed out the static method is longer to use. We just seem to have very different ideas of what makes an API usable or not. Is unwrap() strictly needed? No. Does it improve usability? Yes. Is the change already prepared? Yes. So honestly I don't get why you are that hesistant to add this, but I'll let it go. Closing this PR.

OndrejM:
I created issue #90 to track this idea and evaluate it later. Ideally also outside of config spec, because the idea behind is general and should be considered a pattern across all specifications.

PR #86, opened 2017-03-04
CDI API - Direct injection of config values

OndrejM:
Direct injection of config values was removed from #42, to favor only dynamic retrieval of config values (in case they change during runtime).

This PR separates direct injection proposal to create a separate thread for the discussion whether to allow for direct injection and how. The discussion, which started here (https://github.com/eclipse/microprofile-config/pull/42#issuecomment-283143463) can continue in this PR now.


OndrejM:
I'd like to point out that with this way, all directly injected properties are mandatory. It is not possible to create a CDI bean if an injected config property isn't specified unless we provide a way to define a default value.

AFAIK, [Sabot](https://github.com/tomitribe/sabot/tree/master/sabot/src/main/java/org/tomitribe/sabot) would give errors during deployment, because not value matches such injection points. Similar as if there is no producer for an injection point. This may be a good thing, if the config values are mandatory for the application, to notify of a missing configuration ASAP. I think this is how all implementations should behave.

struberg:
Thanks Ondrej, applied. 

I would prefer if you could change your method names from underlines to camelCase to make them all look the same. But that's nothing we need to do immediately.

OndrejM:
Hi @struberg, I didn't expect that the PR would be merged before discussing whether we want this approach, ConfigValue approach (PR #42) or both. We should discuss in the hangout (but I will not be there this week).

The PR basically only adds ConfigProperty qualifier into API, and CDIPlainInjectionTest test into TCK to test static and dynamic injection. It also brings JUnit, by I will migrate to TestNG once other tests are migrated to Arquillian.

**To summarize this PR:**

The PR basically only adds ConfigProperty qualifier into API, and CDIPlainInjectionTest test into TCK to test static and dynamic injection. It also brings JUnit, by I will migrate to TestNG once other tests are migrated to Arquillian.

The test in the TCK requires that both static and dynamic injection works, and that the dynamic approach retrieves changed values if called multiple times. I tested this with my trivial implementation and it passed the TCK without any hacks around Provider, therefore I'm pretty confident that just using Provider/Instance would provide true dynamic behavior we are after. Basically, it is enough to inject Provider<T> or Instance<T> instead of ConfigValue<T>. Calling Provider.get() multiple times calls the producer multiple times (I was wrong when I thought that the producer would be called only the first time). Therefore it behaves in the same way as ConfigValue.getValue(). If a config value doesn't exist, my implementation just returns null, which works and is expected by the test in the TCK.

struberg:
+1 for moving to TestNG for the whole code base. 
I thought we discussed the @ConfigProperty for quite some time, isn't?
We should really push forward and commit things. We have way too many loose ends open in parallel which create conflicts when merging. This is especially a problem if PRs don't get any comment for a long time. Does the silence mean OK, abstain or veto? 

And yes, I also think the Provider/Instance will work perfectly fine for getting direct access to the configured value each time you invoke it. Note that this will now internally call Config#getValue() each time you call Instance.get() and there is currently no caching etc involved.

That's a great first step and we can build upon this. 

The next step could be adding the ConfigValue which additionally provides caching, variable evaluation, lookup chain, etc. I imagine to have 2 ways to do that a.) programmatic via Config#access() b.) via injection:

```Java
@Inject
@ConfigProperty(value="myprj.archive.port",
                cacheFor=5, 
                timeUnit=TimeUnit.MINUTES, 
                evaluateVariables=true)
private Instance<Integer> portCfg;
```

Where evaluateVariables imo should default to true and timeUnit to TimeUnit.MILLIS.

OndrejM:
OK, @struberg, I agree that it's better to merge PRs if there was at least some discussion. If someone is against or wants changes, they can issue another PR to revert/modify. 

In the case of this PR, I think there were no objections against ConfigProperty, which is also used in #42. The objections were against direct injection (`@Inject int`), but since dynamic injection using `@Inject Provider<Integer>` works as we wanted, I think it was OK to merge it.

We can also consider merging #42 to support injection via a simplistic `ConfigValue`, and open 2 separate PRs to for additional configuration, one for programmatic and one for using annotations.

struberg:
Agree @OndrejM . I think we need to support both as both are valid use cases

The following will be used if you only want to evaluate the config once at injection time:
Pro: getValue() only happens once, it's fast, doesn't trash the Config system under heavy load
Con: no pickup if underlying value changes in the Config system
```Java
@Inject
@ConfigProperty("myprj.archive.url")
private String url;
```

And injecting a 'proxy' will be used if you only want to evaluate the config each time you access the value with get():
Pro: immediately pick up config changes
Con: heavy load on Config system, slower than accessing a native value of course
```Java
@Inject
@ConfigProperty("myprj.archive.url")
private Provider<String> urlCfg;
```

In my personal opinion we even need something in between. See my cacheFor proposal elsewhere.

Emily-Jiang:
As for the injection issue, we discussed in the last hangout but we did not reach any conclusion. The outstanding question is that:
Can you inject any types by using direct injection? @Inject @ConfigProperty() Customer myCustomer;

Mark, you mentioned it is not a problem and volunteered to show what DeltaSpike did? Have you looked into it?

I do agree to use Provider instead of creating another proxy and I also experimented it did work. Therefore, #42 partially is not needed. See my PR #99, which I pushed yesterday and updated the javadoc on ConfigProperty. @OndrejM you can rework the PR #42?

I also suggest to provide a default value for the property as we are using Provider pattern. I also demonstrated on #99. I'll open any issue to discuss.

PR #91, opened 2017-03-09
Ouf of date: ConfigValue implementation for dynamic properties

struberg:
This is almost the full content of ConfigValue which is up for discussion. 
I completely leaves out CDI for now but I will ship this in a separate PR.

If you want to try this out then please feel free to check out https://github.com/struberg/javaConfig/tree/microprofile-impl
This contains a fully working implementation. This is basically the stuff from June 2016 but adopted to the new interface changes. 

have fun!


tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigValueTest.java
```diff
@@ -0,0 +1,169 @@
...
+        Assert.assertEquals(config.access("tck.config.test.javaconfig.configvalue.boolean.ja_mixedcase").as(Boolean.class).get(), Boolean.TRUE);
+        Assert.assertEquals(config.access("tck.config.test.javaconfig.configvalue.boolean.no_mixedcase").as(Boolean.class).get(), Boolean.FALSE);
+
+        Assert.assertEquals(config.access("tck.config.test.javaconfig.configvalue.boolean.j").as(Boolean.class).get(), Boolean.TRUE);
+        Assert.assertEquals(config.access("tck.config.test.javaconfig.configvalue.boolean.j_uppercase").as(Boolean.class).get(), Boolean.TRUE);
+        Assert.assertEquals(config.access("tck.config.test.javaconfig.configvalue.boolean.n_uppercase").as(Boolean.class).get(), Boolean.FALSE);
+
+        Assert.assertEquals(config.access("tck.config.test.javaconfig.configvalue.boolean.oui").as(Boolean.class).get(), Boolean.TRUE);
+        Assert.assertEquals(config.access("tck.config.test.javaconfig.configvalue.boolean.oui_uppercase").as(Boolean.class).get(), Boolean.TRUE);
+        Assert.assertEquals(config.access("tck.config.test.javaconfig.configvalue.boolean.oui_mixedcase").as(Boolean.class).get(), Boolean.TRUE);
```
Emily-Jiang: I don't really like to force to convert y, ja, j n, oui to true. It is not well known. I will suggestion to convert yes, y, on to true, which should be sufficient.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,135 @@
...
+
+
+    /**
+     * TODO feedback from gunnar: could be interesting to have this functionality also as Config#onChange(ConfigChanged)
+     * Callback which can be used with {@link #onChange(ConfigChanged)}
+     */
+    interface ConfigChanged {
+        <T> void onValueChange(String key, T oldValue, T newValue);
+    }
+}
```
Emily-Jiang: This is not what we discussed ConfigValue. What we discussed is just a proxy. This forces to configure the pulling intervals per property.

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigValueTest.java
struberg: You don't speak french? :)
Seriously, we have seen this as requests in DeltaSpike. And it doesn't hurt.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
struberg: No, this is exactly what we discussed originally. Without any caching you will trash the config system (as explained on the hangout yesterday).
This is not a feedback from 'working sample' but from years in production:
https://github.com/apache/deltaspike/blob/master/deltaspike/core/api/src/main/java/org/apache/deltaspike/core/api/config/ConfigResolver.java#L681

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigValueTest.java
Emily-Jiang: Please don't quote DeltaSpike. We are not defining specs around DeltaSpile.

Emily-Jiang: It is fine for DeltaSpike to expand this. -1 on this.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,135 @@
...
+     * Specify that a resolved value will get cached for a certain amount of time.
+     * After the time expires the next {@link #get()} will again resolve the value
+     * from the underlying {@link Config}.
+     *
+     * @param value the amount of the TimeUnit to wait
+     * @param timeUnit the TimeUnit for the value
+     *
+     * @return This builder
+     */
+    ConfigValue<T> cacheFor(long value, TimeUnit timeUnit);
```
OndrejM: I understand that this method allows defining a business requirement to cache a value for at most value+timeUnit period. This is to prevent longer caching for sensitive data, like blacklists, which should be applied ASAP. However, I'm not sure this can be ensured if config sources implement internal caching. Is there a way to notify all config sources that this value should not be cached for more than configured interval?

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,135 @@
...
+     * mycompany.some.url=${myserver.host}/some/path
+     * myserver.host=http://localhost:8081
+     * </pre>
+     * If 'evaluateVariables' is enabled, the result for the above key
+     * {@code "mycompany.some.url"} would be:
+     * {@code "http://localhost:8081/some/path"}
+     * @param evaluateVariables whether to evaluate variables in values or not
+     * @return This builder
+     */
+    ConfigValue<T> evaluateVariables(boolean evaluateVariables);
```
OndrejM: Please also clarify the default value if evaluateVariables is not set

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,135 @@
...
+
+
+    /**
+     * The specified lambda will get called on any config value changes.
+     *
+     * @return This builder
+     */
+    ConfigValue<T> onChange(ConfigChanged valueChange);
```
OndrejM: We have discussed this - it is not clear in which context (thread) the ConfigChanged lambda would be executed. I would drop this method until we clarify this.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
OndrejM: I think that ConfigValue in this PR covers too many things at once. I'd like to split some methods into a separate PR. As a base, I think it is a no-brainer to leave `as`, `get`, `getOptional`, `getKey` and move other methods into PRs on their own. We can then discuss separately.

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigValueTest.java
OndrejM: I agree with Emily, the specification should be formal and pick only widely used options. Any implementation can support more options, and if it doesn't a custom converter should solve the problem. Otherwise, we would have to support all language variants - possibly Japanese, since Fujitsu have also joined MP.

tck/src/main/java/org/eclipse/microprofile/config/tck/base/AbstractTest.java
```diff
@@ -0,0 +1,64 @@
...
+                .addPackage(DuckConverter.class.getPackage())
+                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+                .addAsServiceProvider(ConfigSource.class, CustomDbConfigSource.class)
+                .addAsServiceProvider(ConfigSourceProvider.class, CustomConfigSourceProvider.class)
+                .as(JavaArchive.class);
+
+        addFile(testJar, "META-INF/microprofile-config.properties");
+        addFile(testJar, "sampleconfig.yaml");
+
+        WebArchive war = ShrinkWrap
```
OndrejM: Maybe we would need to change WebArchive to JavaArchive later to allow that TCK can be passed also by implementations that don't support WAR deployments. But I agree to leave WebArchive for now, especially since I tried to hack around arquillian extensions and it is very hard, if not impossible, to turn a JavaArchive into a WebArchive before it is deployed to a container.

Emily-Jiang:
Please don't push in ConfigValue. I don't wish to bring down this just because it is in DeltaSpike. What is the use case here?

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
Emily-Jiang: The get has been provided by Provider. I don't see any more value added by this class.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
Emily-Jiang: I would not want to specify this on the property basis. It adds too much burden on the developers. Surely, the cache for has no effect if the underline config sources have no changes.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
Emily-Jiang: What are we going to achieve here? A use case will help.

Emily-Jiang:
Each method needs to have a use case associated with. I prefer not to add this at all in the first release. I have not got any use cases for this one.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
struberg: This is often a business requirement. Different config properties often have different requirements for being up to date. This is something we added to DeltaSpike as a result to business needs. So there IS a real world need to handle this on a per-attribute basis.

struberg: @OndrejM 

> However, I'm not sure this can be ensured if config sources implement internal caching. 

The caching of the ConfigValue is totally independent from any _possible_ caching of any ConfigSource. This is really a local cache for a single attribute evaluation inside each ConfigValue instance.

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigValueTest.java
struberg: Ok for me, we can move this to be an implementation detail. 
Just let me know which valid list of 'true' values you like to support out of the box.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
struberg: @Emily-Jiang @OndrejM 

> I don't see any more value added by this class.

- caching
- list support
- lookup chains
- callback handling
- variable replacement

That's nothing?

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
struberg: you mean if the referenced variable is not found?
Yes, that's a good point which should get clarified!
In that case we better leave the original string untouched.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
struberg: @OndrejM it will be a synchronous callback within the same thread. Of course we should better specify how it works and how it behaves.

@Emily-Jiang A use case would be 

```Java
ConfigValue cfgPort = config.access("myprj.archive.port")
    .as(Integer.class)
    .cacheFor(2, TimeUnit.MINUTES)
    .onChange(::logConfigChange);
```

and logConfigChange will then log out that a value got changed.
Of course there might be other tasks which get triggered when the configured value flips.

tck/src/main/java/org/eclipse/microprofile/config/tck/base/AbstractTest.java
struberg: It originally was a JavaArchive, but I moved it to a WAR for the following reason:
WebArchive is actually the only type which is widely supported on every container. Regardless of owb and weld standalone, servlet containers or fully blown JavaEE servers.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
struberg: @OndrejM in addition to my last reply: One of the points of having the ability to cache on a per-property basis is that this takes a lot load away from the Config system itself. So the ConfigSources usually are able to implement updates much faster because they have generally less load.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
OndrejM: @struberg The title of this PR is "ConfigValue implementation for dynamic properties" and regarding this, it doesn't offer any more value than a Provider we already accepted to be used.

The other features are not related much to the title of this PR and I would separate them to other PRs, a PR per feature, because they are mostly orthogonal. Would you agree to split it to more PRs, so that we may merge the migration of TCK tests to arquillian and further discuss the new features added by ConfigValue?

struberg: @OndrejM it adds a huge additional value indeed.
Currently you can inject a Long or a Provider&lt;Long&gt;.
A native (or wrapper) type provides no refresh at all. Using Provider will **always** go to the underlying Config system.

the `cacheFor`of ConfigValue otoh provides means to _locally_ cache values and prevents going to the underlying Config system for a specified amount of time. So it effectively removes tons of load from the Config system which gives the Config itself more room to do other tasks. It really heavily reduces the overall load.

OndrejM: I'm not saying it doesn't add any value. But I'd like to get the arquillian migration in TCK merged before we come to any conclusion about ConfigValue...

Emily-Jiang:
I would say the ConfigValue is not needed for Config 1.0. Please convince me otherwise.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
struberg: @OndrejM and @Emily-Jiang, I separated the test setup into an own commit for #107

OndrejM: Thanks @struberg! Now we can continue in this lengthy discussion :)

struberg:
@Emily-Jiang 

> I would say the ConfigValue is not needed for Config 1.0. Please convince me otherwise.

As already written dozen times above. The following use cases get added in a clean way. 

- caching
- list support
- lookup chains
- callback handling
- variable replacement

And all the work is finished. So just saying "I don't like it" is simply **not** a valid argument!
Tell me which of the features you don't like and we can discuss it. But there was no single argument against any of those features yet!

struberg:
Yes @OndrejM we also should have a `cacheFor` (or whatever we call it) **also** for the CDI way. It should effectively be in sync so you can do the same in a programmatic and in a DI way. E.g.

```Java
@Inject 
@ConfigProperty(name="myprj.archive.url", cacheFor=5000)
private Provider<String> archiveUrl;
```

OndrejM:
Maybe having a separate @CacheProperty would be better:

```
@Inject 
@ConfigProperty(name="myprj.archive.url")
@CacheProperty(for=5000, unit=TimeUnit.SECOND)
private Provider<String> archiveUrl;
```

I can create a dual PR to have all functionality proposed both in SE and CDI API.

Emily-Jiang:
Please don't merge before we reach final agreement.

Emily-Jiang:
@struberg I don't think the features you listed are absolutely necessary for the config 1.0. We have a lot of cleanup to do and doc/tests to update and add. The list goes on. We cannot keep on going adding more and more stuff. When I first write the proposal, I listed the dynamic support. That is one of the features to be provided and demonstrated.

struberg:
> Please don't merge before we reach final agreement.

Like you did yesterday despite 2 -1 votes and open clarification requests? ;)

No worries...

struberg:
> When I first write the proposal, I listed the dynamic support. 

The first config proposal from June 2016 did actually contain all those features. So this is not a valid argument. And it is all finished. It really works already. 

Tell me which use case of the 5 you think is not useful and we can discuss it.
And again: "I don't like it" is **not** an argument!

Emily-Jiang:
@struberg I'm not going to argue with you over the merging. You merged multiple commits without waiting for the final solution without any pre-warning. I messed up my environment and the commit contains multiple things which I could not lift it out. I prewarned with a comment saying it will be discussed in the hangout and it is only line and I said I would remove today if no agreement were reached.

Emily-Jiang:
For the features you listed:

    caching
    list support
    lookup chains
    callback handling
    variable replacement
I am not saying they are not valid or useful. Are they must for the Config 1.0 release? I am cautious about the deadline. We have a lot of open issues to be closed in a short timeframe.

struberg:
Oki no worries, we will get it done :)
I'm also fine to postpone it post 1.0. 
But the features are still valid.

Emily-Jiang:
@struberg agreed!

OndrejM:
+1 to leave the PR open and leave it to post 1.0

All those are nice features and I'd like to have them definitely in 1.1, but I'd like to have a CDI equivalent at the same time. For 1.0 it would be a lot of work to create the CDI equivalent, not even talking about updating the spec document and polishing TCK and javadoc (if necessary). 

I'd like to finalize what we have, cleanup the PRs and issues related to 1.0, and align the spec documents with all the rest. Still a lot of work to do. If all goes well, the MP umbrella could even skip config v1.0 and include config v1.1 if it is finished before MP 1.1

Emily-Jiang:
This PR was superseded by some later merged PRs. Close this one.

Issue #41, opened 2017-01-27
automatically load config source and mark it dynamic or not


Emily-Jiang:
Automatically load config sources and annotate it whether dynamic or not

struberg:
Can you explain what you mean with that?

Emily-Jiang:
I would like to add one more method to ConfigSource /**
     * 
     * @return the refresh interval in milliseconds
     */
    long getRefreshInterval();
Any number greater than 0 will mark the config source as dynamic.

struberg:
I made the experience that refreshing the whole config is not that optimal in practice.
You often need to specifcy different reload times depending on each key separately. 
Imo the Config itself should not do any auto-refresh but I'd defer this to the ConfigValue

OndrejM:
I agree with @struberg - config sources should be rather dumb and provide enough metadata for higher levels to use them effectively. Moreover, a hint that a config source is dynamic is not useful if there is not method to reload values. I don't see how a method like getRefreshInterval() could even be used with the current API.

I'd rather explore ways of caching/prefetching config values effectively, adding enough hints from the config source to make it effective. Among useful hints would be whether config values might change during runtime (static vs. dynamic), or how expensive it is to read/reload values, e.g. in-memory, local, remote (local network), slow-remote (internet). 

Maybe @Emily-Jiang meant the method getRefreshInterval() to be used for this purpose, in that case I would change its name, e.g. getRefreshLatency(). A higher-level mechanism could then assume cache-retention periods from this number. But it wouldn't be easy to get this relative value always correct (hard to prevent people from giving latency 1000 to fast sources, while others giving latency 100 to really slow sources).

Emily-Jiang:
We need to differentiate between dynamic sources and static sources. For static sources, we don't need to go back to reload the config properties. For dynamic sources, we need to go back to pull the config changes after a period of interval. 

The problem with baking this time delay on the ConfigValue is that you will have no clue where the config property is from. It might be from a slow database or might from a fast connection. How much time delay would you want to specify per config basis? Secondly, how do you know a particular property being resolved is from a static or dynamic config source? If the config value is from a static source, it is pointless to void cache unnecessarily.

My approach is from the source itself as we know for sure whether it is dynamic or static. The config object is always up to date.

struberg:
I don't yet understand why there is a difference from the user pov. Whether a ConfigSource does reload it's values each time, caches it for a Minute or even only loads it at startup is a question of each ConfigSource itself.

[...]

PR #99, opened 2017-03-15
#41 Add default value in the ConfigProperty

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -39,8 +39,7 @@ import org.eclipse.microprofile.config.spi.ConfigSource;
  * {@link ConfigSource configsources}, the value in the {@link ConfigSource} with the highest
  * ordinal will be used. If multiple {@link ConfigSource configsources} are specified with
  * the same ordinal, no defined ordering will be applied if they contain the same key.
- * <p>
- *
+ * 
  * <h3>Usage</h3>
  * For accessing the config you can use the {@link ConfigProvider}:
  *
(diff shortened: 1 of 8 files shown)


Emily-Jiang:
I'm going to merge this in and discuss one extra API I added getRefreshDuration this afternoon. The PR contains default value capability on the injection model.

struberg:
+1 for the default value in ConfigProperty
-1 for the getRefreshDuration. As explained in the PR (open review question) it is not clear what use case it does solve.

struberg:
Sorry Emily, Both @OndrejM  and I expressed a strong -1 against getRefreshDuration during our yesterdays meeting. And there was also an open Review comment asking for the use case from 15h ago. Let's discuss this in our hangout today.

Emily-Jiang:
That is what I explained in the above comment to discuss this afternoon. I'll detail the use cases in the hangout.

OndrejM:
I don't like doing more than 2 different things in a PR. While I agree with kavadoc changes, the refresh duration discussion has not been closed yet.

OndrejM:
While having a new default method shouldn't hurt, this one is very confusing. The javadoc explains the case of static source, but what should be returned in case of a dynamic one? It's not possible to guess how long a refresh would take. Wouldn't be enough to return just true/false for static/dynamic source? I don't see any point in guessing how fast would the source refresh...

Emily-Jiang:
@OndrejM I think the compromise would be isDynamic to at least state it is dynamic or not. Anyway, I am trying to draft out a note and ask for a wide discussion.

struberg:
Yes, the current method makes not much sense to me for now.
- There is no described use case
- It's nowhere referenced in any test
- It's nowhere implemented

Hard to tell what the intent is actually.

Issue #102, opened 2017-03-15
Remove methods from Config returning a String


gunnarmorling:
As discussed in #77, one get the same via `getValue("name", String.class)`. Providers can experiment with additional methods if needed.

PR to follow in a bit.

Emily-Jiang:
It happened too fast. We only discussed to remove the getValue without type parameter. However this PR removed the getRawString as well. I need this method to remain. The getRawString is to bypass the string converter.

struberg:
This was discussed in https://github.com/eclipse/microprofile-config/pull/77

> getRawString is only needed IF there is a special Converter for String 
> type registered. Otherwise getValue and getRawString will always return 
> the same result. I think this is really a pathological situation. And while the 
> raw string information is sometimes needed, one could still take the 
> values directly from the ConfigSources in that case.

Emily-Jiang:
It was discussed and I commented and did not agree. Getting through ConfigSources is not convenient as you need to loop through it. We need to provide a raw form where a string is not converted. That is my asking.

gunnarmorling:
> It is also useful in testing and overwrite the String converter provided by some containers for an example.

Can you elaborate on this? The use case isn't really clear to me. I.e. in which case would the suggested method return something else than `getValue("key", String.class)`?

Note you can get the raw value without it like so (and it would be even easier if `getConfigSources()` didn't return an `Iterable`):

    Optional<String> value = StreamSupport.stream( config.getConfigSources().spliterator(), false )
        	.filter( s -> s.getProperties().containsKey( "foo" ) )
        	.findFirst()
        	.map( s -> s.getValue( "foo" ) );

I.e. that functionality is already covered in a way which I personally think isn't too bad for such corner use case. I'd rather have people use that, possibly let providers add support for it (would be a great application of `unwrap()` btw.) and eventually add it in a subsequent revision if there actually is demand.

PR #103, opened 2017-03-15
#102 Removing methods from Config that return a String

gunnarmorling:


gunnarmorling:
@struberg, there you go.

Issue #95, opened 2017-03-14
Add missing NOTICE file


johnament:
Much of the work in here is derived from Apache DeltaSpike.  While its free to reuse in source form, under the guidelines of the Apache License v2, the NOTICE file must be propagated to indicate that the source is not originally developed here at Eclipse.  Likewise, files derived from Apache DeltaSpike should not include a Copyright IBM header.

keilw:
This feels like the 3rd or 4th clone of https://github.com/eclipse/microprofile-config/issues/73, https://github.com/eclipse/microprofile-config/issues/88 (which is already worked on) or https://github.com/eclipse/microprofile-config/issues/45.

While there is nothing wrong with a NOTICE file, many files e.g. @Emily-Jiang created were from scratch. The methods not only by returning `Optional `e.g. in the `Config `class were always different from a simple copy of either project mentioned in https://github.com/eclipse/microprofile-config/blob/master/README.adoc 

Everyone who at least provided "inspiration" is mentioned in the contribution history. A mandatory "NOTICE" file may not be wrong or forbidden but we should not duplicate effort to maintain by keeeping a) a NOTICE file, b) a Contribution history in all relevant sources and c) maybe a contributors.txt similar to what Hibernate Validator does. 

What'll be next, I guess @jhalterman requests a NOTICE file in https://github.com/Emily-Jiang/microprofile-faultTolerance-incubation saying "Please this was all copied from Failsafe"??? ;-/

struberg:
Well, if it really was mainly taken from failsafe, then it should also get a proper NOTICE and Copyright attribution. We are not free to do whatever we like - we are bound to law still.

johnament:
@keilw failsafe doesn't include a NOTICE file so he wouldn't be able to

https://github.com/jhalterman/failsafe

struberg:
@johnament but if that stuff really got copied over then it imo should at least get the correct Copyright attribution. One is not free to claim own copyright for sources copied from somewhere else - we are bound to law still.

johnament:
@struberg you're correct.  My qualm here is the NOTICE file required by the Apache License, which equally conveys copyright ownership of portions of the code.

keilw:
Is the NOTICE file required by the Apache License or Foundation?
Those are 2 different things and Thousands of projects use Apache License but rarely any have a NOTICE file.

keilw:
The "spin-off" by @struberg has a NOTICE file saying 
>This product includes software developed at
[...]

PR #104, opened 2017-03-15
refs: #95 add NOTICE file as required by ALv2.

struberg:
txs to @johnament for reporting!

Signed-off-by: Mark Struberg <struberg@apache.org>

@johnament @Emily-Jiang @OndrejM @gunnarmorling Please check the NOTICE file and if you have something to add. Please also read the Section 4.d of the Apache License before you change something. https://www.apache.org/licenses/LICENSE-2.0

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,13 @@
+=========================================================================
+==  NOTICE file corresponding to section 4(d) of the Apache License,   ==
+==  Version 2.0, in this case for Microprofile Config                  ==
+=========================================================================
+
+This product includes software developed at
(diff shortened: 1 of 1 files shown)


Emily-Jiang:
This file needs to be under META-INF, right? I have raised #105 to add LICENSE.txt, which should be in the same folder as NOTICE.

struberg:
umm true, it should ideally be in the root of the project and also distributed in the jars.
Will check how this is set up at the ASF. Maybe via resource plugin, need to investigate.

PR #116, opened 2017-03-30
Update configprovider.asciidoc

doychin:
Fix typo

Signed-Off-By: Doychin Denkov Bondzhev


doychin:
I'm trying to do it from github web ui but looks like this will now work.

PR #120, opened 2017-04-06
Renamed artifact, removed common annotations dep

OndrejM:
Removed the unnecessary dependency on common annotations, which is only used in Javadoc


Emily-Jiang:
I think this PR can be closed as I have fixed the artifact id/group id under #133 . Please reopen if you think otherwise.

Issue #89, opened 2017-03-09
Clarify if Config instances need to be Serializable


struberg:
When storing a Config instance inside a serializable class (e.g. @SessionScoped) then we would need to have the Config Serializable as well.

This is easy to implement for container managed Config instances at least. 
While the internal data (ConfigSources etc) are often not Serializable we can simply _re-attach_ to the current application Config again.

Not sure about Config instances one creates manually and does _not_ manage within the application. Those might end up not being Serializable.

Emily-Jiang:
The unmanaged config should not be serializable. We should put this on the javadoc.

Emily-Jiang:
MP Config hangout discussion: for @Inject config, it is guaranteed to be serialzable.

Emily-Jiang:
After more thoughts on this, the injected config is difficult to be serializable as it has a list of config sources (default sources + auto discovered sources), which are not guaranteed to be serializable.  We need to modify the javadoc changes to say they are not serailzable even for the injected ones.

OndrejM:
This is an implementation detail. But I think that even with dependent instances, the injected config instances can provide a custom serialization mechanism by implementing read/writeObject methods.

PR #127, opened 2017-04-13
#89 Added test for serializability of injected config

OndrejM:
Part of issue #89.
The test just checks whether it's possible to serialize, deserialize and whether the deserialized object is again a `Config` instance. 

It doesn't check whether it gives any config values - the config values should ideally not be serialized, but retrieved from the runtime environment (especially if it moved to another JVM)

Mark's sample implementation already passes the test.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
@@ -19,6 +19,7 @@
  */
 package org.eclipse.microprofile.config.tck;
 
+import java.io.*;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Properties;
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
```diff
@@ -19,6 +19,7 @@
  */
 package org.eclipse.microprofile.config.tck;
 
+import java.io.*;
```
Emily-Jiang: Just to be picky, "I would like the class level import".

OndrejM: I hate long lists of imports for obvious standard Java classes, but I'll fix it directly with a commit and update my IDE config appropriately just to stay consistent.

OndrejM: Hm, 5 imports instead of 1 - I can feel the pain before comitting :/ :

```
import java.io.ByteArrayInputStream;
import java.io.ByteArrayOutputStream;
import java.io.IOException;
import java.io.ObjectInputStream;
import java.io.ObjectOutputStream;
```

PR #130, opened 2017-04-28
Custom converter

jmesnil:
* [api] add ConfigBuilder#addDiscoveredConverters to be able to
  discover converters using the ServiceLoader pattern
* [spec] mention new API above and state that converters are not
  discovered by default
* [tck] add tests for customer converters

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
@@ -57,12 +57,16 @@ public abstract class ConfigProviderResolver {
     public abstract Config getConfig(ClassLoader loader);
 
     /**
-     * Create a fresh {@link ConfigBuilder} instance. This ConfigBuilder will
-     * initially contain no {@link ConfigSource} but with default {@link Converter Converters}
-     * The other {@link ConfigSource} and {@link Converter Converters} will have to be added manually.
+     * Create a fresh {@link ConfigBuilder} instance.
+     *
+     * This ConfigBuilder will initially contain no {@link ConfigSource}. The other {@link ConfigSource} will have
+     * to be added manually or discovered by calling {@link ConfigBuilder#addDiscoveredSources()}.
+     *
+     * This ConfigBuilder will initially contain default {@link Converter Converters} as well as any converters
+     * discovered using the ServiceLoader pattern.
      *
      * The ConfigProvider will not manage the Config instance internally
-     * @return the configbuilder with the default converters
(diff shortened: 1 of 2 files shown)


jmesnil:
There is a contradiction between the TCK and the spec.

The TCK assumes that config sources and converters that are loaded by the Service Loader will be added to the Config by default while the spec states the opposite in https://github.com/eclipse/microprofile-config/blob/master/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java#L60

According to the API, a Config must not contain either default or discovered config sources by default which seems useless.

struberg:
Hi Jeff!

I think I slowly get what you mean. 
You are right that this is kind of 'asymmetric' between ConfigSources and Converters.
If I remember correctly we have been discussing this with Emily, Gunnar and Romain as well already.

I think the outcome was that this is not really needed. You can always provide a Converter with a higher ordinal anyway. And there must be only exactly one Converter per type at the end. We decided against having multiple Converters per type and a 'fallback logic', because that would have added quite some complexity! The current solution is that the 'effective' Converter (the one with the highest @Priority) simply has to provide all the converting. And if there are multiple formats (e.g. various date formats for the same type) then the Converter has to prove them all itself.

This also means that it imo doesn't hurt if a Converter gets automatically detected and picked up. It makes no difference, isn't? Do you have a use case where a Converter for a custom type gets discovered and you don't want that?
If you don't want that then you most likely also don't use that Type in your config I assume. And if you would (without the converter for this custom type), then we would blow up with a deployment exception anyway. 

While Converters and ConfigSources share quite some behaviour they are in some regard also fundamentally different.

Do you think there is any use case we miss?
If so, which one?

txs and LieGrue,
strub

jmesnil:
@struberg thanks for the clarification. I see no downside to discover converters by default.

I was confused by the javadoc at https://github.com/eclipse/microprofile-config/blob/master/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java#L62. This sentence implies to me that the discovered converters are not added to the created Builder. That's why I added the `addDiscoveredConverters` method to the SPI.

Now that you have clarified this use case, I think we only need to fix this sentence and states that a empty Builder will contain default converters *as well as discovered ones*.

I'll amend my PR to remove the SPI methods I added and fix the javadoc.

struberg:
thanks for the review and feedback, Jeff!

jmesnil:
@struberg I've amended my PR. You pushed a test for custom converter in b4e1949a7a4b5f59528311f637e91c0be7cc208d but the test was not run.
I've also updated the Javadoc according to your feedback.

Issue #147, opened 2017-05-08
Tck: CustomConfigSourceTest missing deployment of CustomDbConfigSource.class 


hutchig:
The Tck test [CustomConfigSourceTest](https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/CustomConfigSourceTest.java)
has
```
.addAsServiceProvider(ConfigSource.class, CustomDbConfigSource.class)
.addAsServiceProvider(ConfigSourceProvider.class, CustomConfigSourceProvider.class)
```
but neither class is added to the archives explicitly. 
One can see the javadoc for addAsServiceProvider has ` Warning: this method does not add the specified classes to the archive.` :

```
/**
     * Adds a META-INF/services/ServiceInterfaceName {@link Asset} representing this service.
     *
     * Warning: this method does not add the specified classes to the archive.
     *
     * @param serviceInterface
     *            The Service Interface class
     * @param serviceImpls
     *            The Service Interface Implementations
     * @return This virtual archive
     * @throws IllegalArgumentException
     *             if serviceInterface is null
     * @throws IllegalArgumentException
     *             if serviceImpls is null or contain null values
     */
    /*
     * TODO: The interface should have been like this: <X> T addServiceProvider(Class<X> serviceInterface, Class<?
     * extends X>... serviceImpls) throws IllegalArgumentException; But due to how java generic works, this will cause a
     * unsafe warning for the user.
     */
    T addAsServiceProvider(Class<?> serviceInterface, Class<?>... serviceImpls) throws IllegalArgumentException;
```
...and indeed the test fails with
com.ibm.ws.microprofile.config.interfaces.ConfigException: CWMCG0011E: Unable to discover ConfigSourceProviders. The exception is: java.util.ServiceConfigurationError: org.eclipse.microprofile.config.spi.ConfigSourceProvider: Provider org.eclipse.microprofile.config.tck.configsources.<CustomDbConfigSource or>CustomConfigSourceProvider not found.

 without the additional line:
```
  .addClasses(CustomDbConfigSource.class, CustomConfigSourceProvider.class)
```
around line 54.

struberg:
Good catch! 

Just to explain why I did not see those and why all seem to work in my impl.
[...]

Issue #146, opened 2017-05-08
Tck: CDIPlainInjectionTest war missing classes TestSetup, AdditionalMatchers


hutchig:
In [CDIPlainInjectionTest](https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPlainInjectionTest.java)
We see
```
@Test
    public void can_inject_dynamic_values_via_CDI_provider() {
...        
   ensure_all_property_values_are_defined();
...
   private void ensure_all_property_values_are_defined() {
        TestSetup.ensurePropertyDefined("my.string.property", "text");
```
but TestSetup is not in the jar, so one gets: 
```
can_inject_dynamic_values_via_CDI_provider(org.eclipse.microprofile.config.tck.CDIPlainInjectionTest)  Time elapsed: 0.153 sec  <<< FAILURE!
java.lang.NoClassDefFoundError: org/eclipse/microprofile/config/tck/testsupport/TestSetup
```
The test case needs: 
.addClasses(TestSetup.class, AdditionalMatchers.class)
added to the war creation ~line:60

Emily-Jiang:
Thanks @hutchig ! Please check whether the latest code fix is ok. I tried to fix it but Mark raced and committed first. If the fix is ok, please close the issue.

hutchig:
We did get AdditionMatchers.class in Mark's change but we have still not got the required
"TestSetup.class" that is in Emily's commit. So one more change needs to get to the branch.

Emily-Jiang:
@hutchig Mark deleted the TestSetup.java in his commit.

struberg:
yes sorry, did a git-pull but got no changes, no sure what happened.

The TestSetup.java was just used in one test, so instead of adding the class I just pulled the one method which got used over to this one test.

@Emily-Jiang do you want to fix it or should I?

hutchig:
Working now.

PR #148, opened 2017-05-09
fix tck issues #146 #147

Emily-Jiang:


Emily-Jiang:
It seems the issues were fixed.

Issue #139, opened 2017-05-05
Differentiate the non existing property and empty property


Emily-Jiang:
Currently in the config spec, we stated that we cannot differentiate the property with value of "" and non existing property. I would like the javadoc relax about this. We should allow the implementation to differentiate if they can. We should not spec the limitation. Or we should fix this. I think we can differentiate it by calling getConfigNames().

OndrejM:
I agree that any spec shouldn't limit what impl can do, only if it really makes sense to limit. I'd mention in the javadoc that the behavior for an empty value is undefined, in case of an empty value it's recommended that it's treated as non-existent value but empty values should be avoided to maintain portable behavior.

Emily-Jiang:
I'll remove the tck that asserts the limitation and modify the javadoc accordingly.

struberg:
Can you please first write up the use cases where such a "" vs null might happen?
Then it's more visible what you like to change.

Emily-Jiang:
ConfigsourceA:
allergy = ""

On reading this property, I would like to know the value is "", which is different from the property does not exist.

PR #149, opened 2017-05-14
#139 not spec the limitation as it can be achievable

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -102,9 +102,7 @@ public interface Config {
     /**
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
-     *
-     * An empty string representation is interpreted as not-existing configuration.
-     *
+     *     
      * If this method is used very often then consider to locally store the configured value.
      *
      * @param <T>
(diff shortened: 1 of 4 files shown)


struberg:
Just to explain why the original decision was made: In DeltaSpike we allow to 'erase' configured values in a higher ordinal by overwriting it with an empty value.

For this mechanism to work it makes sense to interpret an empty String as 'not existing value'.

We can now discuss whether it makes sense for MP-config to do the same or not. But removing this feature would be a design change we need to discuss.
So yes, it's not a bug, it's a feature right now!

Emily-Jiang:
For your scenario, if the empty string is treated as non-existing property, how could you erase the property? Basically, if the empty string is only in sourceA, the property does not exist. If the empty string is in sourceA with a higher priority, it erases the config property defined in other config sources with a lower priority. This sounds hacking.

On the other hand, I would like to come up with a better design for the property erasing. Depending on the value is not good. I would recommend to introduce a property naming convention such as $aa$ is for erasing the property aa. You can raise an issue to discuss the requirement.

We have already discussed on the issue 139. Both Ondrej and I agreed we should not spec the limitation. I would like to get this sorted asap.

struberg:
in properties file (ordinal=100): myapp.somekey=bla
in environment (higher ordinal): myapp.somekey=
(empty)

From an appliation pov that hat would be the same like not having myapp.somekey defined in any ConfigSource at all.

Emily-Jiang:
@struberg with what you suggested, we no longer support the config property having a null value as a valid property. e.g. myapp.somekey=
(empty)

Emily-Jiang:
I recommend you raise a separate issue to cover the property deletion. As my explanation earlier, relying on the empty value is not the ideal solution for this.

Emily-Jiang:
@struberg updated the javadoc based on your comments! Thanks!

Emily-Jiang:
Done the javadoc improvement based on your suggestion, Mark. If you have any further comments, please open a separate issue for further tidyup.

PR #171, opened 2017-06-15
listing site documentation on microprofile.io

tveronezi:
This file is used by the microprofile.io website in order for the listed projects to have more control on what is published on http://microprofile.io/projects

The entry in this file will list the project documentation in a list like the one in the screenshot.

screen shot 2017-06-15 at 2 57 49 pm

PR #172, opened 2017-06-16
Individual site config file

tveronezi:
This is the same as https://github.com/eclipse/microprofile-config/pull/171 but with the Signed-off-by in the commits.

### DIFF
--- a/site.yaml
+++ b/site.yaml
@@ -1,6 +1,5 @@
 %YAML 1.2
 ---
-friendly_name: "My custom name"
 documentation:
   - title: Configuration for Microprofile
     file: spec/src/main/asciidoc/microprofile-config-spec.asciidoc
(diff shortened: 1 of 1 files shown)


Emily-Jiang:
From your description, the file is optional if I don't care about the order. Right? If so, which order it will use to construct the content. Please confirm.

tveronezi:
@Emily-Jiang this file is not required. It can be used by a project to add custom configuration like a friendly name or a list of documents hosted in the project itself. This PR creates the list of documents like the screenshot bellow...

screen shot 2017-06-16 at 12 03 24 pm

The order of the documents in that list is the order of the documents in this site.yaml file.

Currently, the live version of this project documentation looks like this...



https://microprofile.io/project/eclipse/microprofile-config

dblevins:
Without this file there will be no documentation listed in the right.

Emily-Jiang:
Thanks @tveronezi @dblevins for the explanation! Just try to understand the usage.

PR #177, opened 2017-07-04
Fixes #176 minor clarification to javadoc of Converter

### LINKED ISSUE #176: Clarify that Converter should return null when null value given
Minor change required to the javadoc for Converter

smillidge:


api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -67,7 +68,7 @@
     /**
      * Configure the string value to a specified type
      * @param value the string representation of a property value
-     * @return the converted value
+     * @return the converted value or null if value is null
```
Emily-Jiang: Not necessarily. If a converter can throw exception if the value is null though. It depends on the converter. I would rather not spec this.

smillidge:
Personally I would spec it. Otherwise the builtin converters could work differently on different platforms. I looked at the Geronimo implementation to check what to do and aligned with that as I had originally planned to throw IllegalArgumentException for a null value.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
Emily-Jiang: ah. I see what you meant. I would rather put the null if value is null on the spec not this javadoc as this javadoc is applicable for all converters not just the builtin converters.

smillidge: OK I would update the spec then to say all Builtin converters return null if a null value is provided. I still think an implementer of a ConfigSource would want some guidance on null handling.

OndrejM: @Emily-Jiang How would the configuration then treat exceptions from a converter? I think that it should be explicitly forbidden to throw an exception and rather return null or another value if makes sense.

OndrejM: @Emily-Jiang, I don't understand why a converter would want to throw an exception for null values? Empty value is a valid value and can always be converted to null with any converter.

struberg: +1 @OndrejM I defined it because most of the times null gets ignored and then people complain about things behaving undefined. Whether we finally need it or not - the behaviour is now defined!

Emily-Jiang: For a custom converter, it is decided by the application developer whether a null value is accepted. Hence my suggestion.

smillidge: I understand the Converter handling null and perhaps converting it into a valud value. 
If that is the case then it needs documenting that the Converter needs to handle null input without throwing a NPE or any exception.

I can change my PR if you want?

Emily-Jiang:
@smillidge sure, please update the pr. thanks

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
Emily-Jiang: sure. thanks

smillidge:
PR updated

Emily-Jiang:
Thanks @smillidge ! looks good.

PR #186, opened 2017-07-05
Adding explanation for where this `config_ordinal` applies.

### LINKED ISSUE #159: Scope of where config_ordinal applies
`configsources` mentions the following:

> The ordinal for property file based <<ConfigSource,ConfigSources>> can be configured using the key `config_ordinal` inside the property file.

However, the only property file mentioned is `META-INF/microprofile-config.properties` so does this only apply to this property file?

johnament:
This closes #159

### DIFF
--- a/spec/src/main/asciidoc/configsources.asciidoc
+++ b/spec/src/main/asciidoc/configsources.asciidoc
@@ -35,22 +35,22 @@ It can also be used to implement a drop-in configuration approach.
 Simply create a jar containing a `ConfigSource` with a higher ordinal and override configuration values in it.
 If the jar is present on the classpath then it will override configuration values from <<ConfigSource,ConfigSources>> with lower `ordinal` values.
 
-The ordinal for property file based <<ConfigSource,ConfigSources>> can be configured using the key `config_ordinal` inside the property file.
-
 [source, text]
 ----
 config_ordinal = 120
 com.acme.myproject.someserver.url = http://more_important.server/some/endpoint
 ----
 
+Note that `config_ordinal` can be set within any `ConfigSource` implementation.  The default implementat of `getOrdinal()` will attempt to read this value, if found and a valid integer, the value will be used, otherwise the default of `100` will be returned.
+
 [[default_configsources]]
 === Default ConfigSources
 
 A Microprofile-Config implementation must provide <<ConfigSource,ConfigSources>> for the following data out of the box:
 
-* System properties (ordinal=400)
-* Environment variables (ordinal=300).  
-* A `ConfigSource` for each property file `META-INF/microprofile-config.properties` found on the classpath. (default ordinal = 100)
+* System properties (default ordinal=400).
+* Environment variables (default ordinal=300).
+* A `ConfigSource` for each property file `META-INF/microprofile-config.properties` found on the classpath. (default ordinal = 100).
 
 [[custom_configsources]]
 === Custom ConfigSources
@@ -82,7 +82,7 @@ public class CustomDbConfigSource implements ConfigSource {
     public String getName() {
         return "customDbConfig";
     }
-    
+
 }
 
 ----
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
@@ -58,9 +58,11 @@ import java.util.Map;
  * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
  * @author <a href="mailto:gpetracek@apache.org">Gerhard Petracek</a>
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
+ * @author <a href="mailto:john.d.ament@gmail.com">John D. Ament</a>
  *
  */
 public interface ConfigSource {
+    String CONFIG_ORDINAL = "config_ordinal";
     /**
      * Return the properties in this config source
      * @return the map containing the properties in this config source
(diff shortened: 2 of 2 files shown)


Emily-Jiang:
please see my comments for the change.

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -48,9 +46,10 @@ com.acme.myproject.someserver.url = http://more_important.server/some/endpoint
 
 A Microprofile-Config implementation must provide <<ConfigSource,ConfigSources>> for the following data out of the box:
 
-* System properties (ordinal=400)
-* Environment variables (ordinal=300).  
-* A `ConfigSource` for each property file `META-INF/microprofile-config.properties` found on the classpath. (default ordinal = 100)
+* System properties (ordinal=400).
+* Environment variables (ordinal=300).
+* A `ConfigSource` for each property file `META-INF/microprofile-config.properties` found on the classpath. (default ordinal = 100).
+** The ordinal for this config source can be configured using the key `config_ordinal` inside the property file.
```
OndrejM: This line should be removed.

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -35,8 +35,6 @@ It can also be used to implement a drop-in configuration approach.
 Simply create a jar containing a `ConfigSource` with a higher ordinal and override configuration values in it.
 If the jar is present on the classpath then it will override configuration values from <<ConfigSource,ConfigSources>> with lower `ordinal` values.
 
-The ordinal for property file based <<ConfigSource,ConfigSources>> can be configured using the key `config_ordinal` inside the property file.
```
OndrejM: This shouldn't be removed but reworded instead. Ordinal property can be applied to any config source, not only "property file based" config sources.

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -48,9 +46,10 @@ com.acme.myproject.someserver.url = http://more_important.server/some/endpoint
 
 A Microprofile-Config implementation must provide <<ConfigSource,ConfigSources>> for the following data out of the box:
 
-* System properties (ordinal=400)
-* Environment variables (ordinal=300).  
-* A `ConfigSource` for each property file `META-INF/microprofile-config.properties` found on the classpath. (default ordinal = 100)
+* System properties (ordinal=400).
```
OndrejM: In System properties and env vars, this should be changed to "(default ordinal = 400)" as with `config_value` the ordinal can be changed also for these config sources.

spec/src/main/asciidoc/configsources.asciidoc
johnament: I disagree.  The main problem with this line is that it implies you can add other property files, outside of `META-INF/microprofile-config.properties`.  We don't present support this outside of writing a custom property source.  Therefore we should not specify within 1.0 that you can do this, which is why I put it as part of the property file.

A change like what is being described means that we would need to ask the config source for the value of `config_ordinal` each time, which is potentially expensive.

struberg: I'd say we should make clear that the config_ordinal mechanism is guaranteed to work for built-in ConfigSources. And of course we can recommend that other ConfigSources use the same trick. But at the end only the getOrdinal() method counts, regardless how this value got evaluated or defined in the custom ConfigSource. 

We also cannot test every custom ConfigSource in our TCK, so any specification in this area is moot anyway, isn't?

Emily-Jiang: I agree with Ondrej. As explained in my comments on the issue, config_ordinal should be treated same for all config sources, default one or custom one. It is perfect valid and good to specify config_ordinal in the property files or custom config sources.

johnament: I have mixed feelings on this.  To Mark's point, what I think we can do is provide the default implementation of `getOrdinal` returning the property value if present, 100 if not present.  It still seems very awkward.

Emily-Jiang: The default values are different for different config sources. 
* System properties (ordinal=400)
* Environment variables (ordinal=300).  
* A `ConfigSource` for each property file `META-INF/microprofile-config.properties` found on the classpath. (default ordinal = 100)

For custom config sources, developers would have to implement the method of getOridinal().

johnament: Take a look at let me know if this aligns.

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -35,22 +35,22 @@ It can also be used to implement a drop-in configuration approach.
...
 
-The ordinal for property file based <<ConfigSource,ConfigSources>> can be configured using the key `config_ordinal` inside the property file.
-
 [source, text]
 ----
 config_ordinal = 120
 com.acme.myproject.someserver.url = http://more_important.server/some/endpoint
 ----
 
+Note that `config_ordinal` can be set within any `ConfigSource` implementation.  The default implementat of `getOrdinal()` will attempt to read this value, if found and a valid integer, the value will be used, otherwise the default of `100` will be returned.
```
Emily-Jiang: typo implementation. I would rephrase this to make it readable. In general, it is ok.

Emily-Jiang:
@OndrejM can you double check to see whether you are happy except the typo I mentioned?

spec/src/main/asciidoc/configsources.asciidoc
OndrejM: > A change like what is being described means that we would need to ask the config source for the value of config_ordinal each time, which is potentially expensive

I really don't think that it implies that the value of config_ordinal is checked each time. It should be checked only once when a config source is created, am I wrong? And checking a single value is negligible to checking all the other values that an application usually requires.

Making support for `config_ordinal` in custom sources optional completely defeats its purpose. If a config source ordinal isn't configurable, it's not easy to configure its order if it clashes with other config sources unless all ordinals in other config sources are modified. 

Therefore I'm against relying solely on the output of the getOrdinal() method, which would optionally return value of the `config_ordinal` property.

OndrejM: With the solution I propose, there's still a way to opt out of using the `config_ordinal`, because the source has complete control over what it returns from the `getValue()` method. If it's really necessary to avoid reading the `config_ordinal` property, the `getValue()` method can treat this property in a special way and e.g. return the value of `getOrdinal()` intad of reading it from configuration.

OndrejM:
We agreed with @Emily-Jiang during the hangout today, that we'll merge this PR and release RC2. Afterwards I'll create a PR with my suggestions and we may discuss whether we want it for 1.0 or discuss it later.

PR #198, opened 2017-07-20
Minor clarification on ordinal precedence.

mikecroft:
This might be already clarified elsewhere, but this is a useful place to have it.

Signed-off-by: Mike Croft <mike.croft@payara.fish>

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -73,7 +73,7 @@ By default there are 3 default ConfigSources:
 * all `META-INF/microprofile-config.properties` files on the ClassPath.
 (default ordinal=100, separately configurable via a config_ordinal property inside each file)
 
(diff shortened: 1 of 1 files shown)


johnament:
@mikecroft do you want to fix the ip check?

mikecroft:
Ah, yeh, I'll make a new pr. I did this change on GitHub so it's got the noreply email in.

struberg:
It's already specified in ConfigSource#getOrdinal() and also in the introduction section afair. So this section is probably redundant then.

mikecroft:
I still think it's useful here @struberg - mostly for people who are just interested in the spec as a user, rather than implementer. The README is often the starting point and it's the text used for the main project page on the microprofile.io website.

mikecroft:
Fixed the IP check @johnament

Issue #200, opened 2017-07-25
Fix Sonatype validations errors to release in Maven Central


OndrejM:
The following validations are failing when deployed only API and parent artifacts:
 - `Invalid POM: /org/eclipse/microprofile/config/microprofile-config-parent/1.0/microprofile-config-parent-1.0.pom: Project description missing, Developer information missing`
 - `Missing: no sources jar found in folder '/org/eclipse/microprofile/config/microprofile-config-api/1.0'`
 - `Missing Signature: '/org/eclipse/microprofile/config/microprofile-config-parent/1.0/microprofile-config-parent-1.0.pom.asc' does not exist for 'microprofile-config-parent-1.0.pom', 'microprofile-config-api-1.0-javadoc.jar', 'microprofile-config-api-1.0.pom', 'microprofile-config-api-1.0.jar'`

Here are Sonatype requirements: http://central.sonatype.org/pages/requirements.html

OndrejM:
A guide to sign with a GPG key: http://blog.sonatype.com/2010/01/how-to-generate-pgp-signatures-with-maven/

johnament:
@OndrejM  is this all done?

johnament:
Closing this out, seems done.

PR #201, opened 2017-07-26
Fixing release for Maven Central

OndrejM:
Applying fixes to deploy to Maven Central to 1.1-SNAPSHOT version. For 1.0 they aren't stored in the repo because the tag is already created and I didn't want to change the tag.


OndrejM:
Fixes #200

PR #207, opened 2017-07-29
Adding a TCK test for default property keys on injection points.

### LINKED ISSUE #197: Add TCK test for default named ConfigProperty
Add a test to validate the javadoc comment for ```ConfigProperty```

If it is not specified, it will be derived automatically as {@code <class_name>.<injetion_point_name>},
     * where {@code injection_point_name} is the field name or parameter name,
     * {@code class_name} is the fully qualified name of the class being injected to with the first letter decaptialised.

I would add it but I am not clear on what the name should be from the description above.

johnament:
This closes #197

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPlainInjectionTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPlainInjectionTest.java
@@ -51,11 +51,15 @@ import org.testng.annotations.Test;
  */
 public class CDIPlainInjectionTest extends Arquillian {
 
+    private static final String DEFAULT_PROPERTY_BEAN_KEY = "org.eclipse.microprofile.config.tck.CDIPlainInjectionTest.defaultPropertyBean.configProperty";
+
     @Deployment
     public static Archive deployment() {
+        // this is needed since there's a requirement to validate on start up
+        System.setProperty(DEFAULT_PROPERTY_BEAN_KEY, "pathConfigValue");
         return ShrinkWrap.create(WebArchive.class)
-                .addClasses(CDIPlainInjectionTest.class, SimpleValuesBean.class, DynamicValuesBean.class, AdditionalMatchers.class
-                           , TestConfigSource.class)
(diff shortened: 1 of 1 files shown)


johnament:
I'm not a huge fan of setting the system property, otherwise I can add it as an entry in microprofile-config.properties.

tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPlainInjectionTest.java
```diff
@@ -51,11 +51,15 @@
  */
 public class CDIPlainInjectionTest extends Arquillian {
 
+    private static final String DEFAULT_PROPERTY_BEAN_KEY = "org.eclipse.microprofile.config.tck.CDIPlainInjectionTest.defaultPropertyBean.configProperty";
```
OndrejM: Shouldn't the D in `defaultPropertyBean` be a capital D? It should be the name of the class, right?

johnament: No, because the javadoc mentions the first letter is [decapitalized](https://github.com/eclipse/microprofile-config/blob/3abbaa9cf8351ea4da49b64a3188e32c91be6790/api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java#L105) (as Steve mentioned as well in his report)

Also, since this is only mentioned in javadocs I think I'm going to update the actual spec to reflect this.

Emily-Jiang: We had some discussions on this in the past and settled to be the decaptiallized version. I then added that to the javadoc.

OndrejM: OK, I see. It's very confusing in this case, but I guess we can't do anything about it since it's already in the final version. I'm fine with that since it's a rare edge case to have a nested CDI bean class.

PR #218, opened 2017-08-13
Adding a TCK test that clarifies values being empty as being present.

### LINKED ISSUE #216: Clarify config value presence
The API javadocs for `Config.getValue` mentions:

```
@throws java.util.NoSuchElementException if the property isn't present in the configuration.
```

Geronimo Config takes an over zealous approach to this, throwing NoSuchElementException if the property exists but has an empty value.  Payara takes a more conservative approach, only throwing this if the value is null.

The fact that the TCK doesn't test for empty string makes me thing the Payara thinking is more correct, but wanted to get confirmation (and possibly a TCK assertion for this).

johnament:
This closes #216

Signed-off-by: John D. Ament <johndament@apache.org>

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValuesTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValuesTest.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package org.eclipse.microprofile.config.tck;
+
+import org.eclipse.microprofile.config.Config;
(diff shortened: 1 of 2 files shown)


tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValuesTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Test
+    public void testEmptyStringValues() {
+        assertEquals(emptyValuesBean.getStringValue(), "");
+    }
+
+    @Test
+    public void testEmptyStringProgrammaticLookup() {
+        System.setProperty(EMPTY_PROPERTY, "");
+        String stringValue = config.getValue(EMPTY_PROPERTY, String.class);
```
Emily-Jiang: I think putting the EMPTY_PROPERTY to the .properties is a better approach, as this test assume the system property will make to the runtime instantly.

johnament: @Emily-Jiang I'm not sure what you mean, this is the same approach used in `CDIPlainInjectionTest.ensure_all_property_values_are_defined`

Emily-Jiang: I prefer you add my.empty.property to microprofile-config.properties as it would represent the real use case. I think a lot of times to have them set blank and then override via a system property to set sometime more meaningful.

As the tck has the set the presence of using system property and expect it to appear as config straightaway. I am not going to block your commit though I think it can be approved. If @struberg or @OndrejM are fine with it, I am ok with it.

Emily-Jiang:
@johnament Feel free to merge this one. We can improve it later if you feel differently.

tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValuesTest.java
johnament: Ok, well now I understand what you're asking for.  I'll change it, makes sense.

johnament: Added additional test from property file.

Issue #233, opened 2017-09-04
ConfigProperty default name


Emily-Jiang:
In ConfigProperty.java, the javadoc says
 {@code class_name} is the fully qualified name of the class being injected to with the first letter decaptialised.

Originally, I proposed to be {@code class_name} is the **simple** name of the class being injected to with the first letter decaptialised. 
The decaptialised bit does make sense. 

Later on, Mark changed it to be fully qualified classname. I missed the commit. The decaptialised bit looks very odd. e.g
com.acme.myBean.myProperty
as com.acme.Mybean.myProperty is much better read.

Can we remove the decaptialisation bit? I know this is behaviour change. Considering the light usage, it should be minimal. Thoughts?

Emily-Jiang:
@struberg @OndrejM @johnament should be fix this in MP Config 1.1?

Emily-Jiang:
I'll vote yes, though.

OndrejM:
This would be a breaking change, although this is how I wanted it initially. And to be honest, I was really confused why our Payara impl decapitalizes the class name until I found out that this is specified by the spec.

I'm in favor of this fix, although we may also preserve the current behavior as deprecated and the new behavior taking precedence if both keys are found. In other words, if a Capitalized config key is found, it will be used, and if not, decapitalized key will be searched for and used if found. The new behavior should be stressed while the old behavior still supported and mentioned somewhere in the spec (e.g. as a compatibility or deprecation note).

Emily-Jiang:
+1 on Ondrej!

johnament:
Since there was no TCK test for this previously, I'm OK if we change the behavior in the release.  GConfig actually didn't support it at all.

johnament:
BTW, if you want this for MP Config 1.1, it has to have a pull request submitted today.  If it doesn't make this release, we'll need to keep both behaviors around for following releases.

Emily-Jiang:
ok. can we make agree on which one to go for? I will submit a PR once a decision is made.
1. Fully qualified classname?
2. support both as Ondro suggested.
At this stage, I lean towards to support the option 1 as it is not well used. Most runtime will go directly to Config 1.1. Can we vote and then I will do a PR?

smillidge:
2 or 2.toLowerCase()

Emily-Jiang:
Thanks @smillidge ! I'll provide a PR with 2.

OndrejM:
[...]

PR #239, opened 2017-09-06
Adding release notes for the 1.1 release, based on input from #233

johnament:

### DIFF
--- a/spec/src/main/asciidoc/release_notes.asciidoc
+++ b/spec/src/main/asciidoc/release_notes.asciidoc
@@ -1,0 +1,40 @@
+//
+// Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// Contributors:
+// John D. Ament
+
+[[release_notes_11]]
+== Release Notes for MicroProfile Config 1.1
+
+The following changes occurred in the 1.1 release, compared to 1.0
+
+A full list of changes may be found on the link:https://github.com/eclipse/microprofile-config/milestone/2?closed=1[MicroProfile Config 1.1 Milestone]
+
+=== API/SPI Changes
+
+- The `ConfigSource` SPI has been extended with a default method that returns the property names for a given `ConfigSource`  (link:https://github.com/eclipse/microprofile-config/issues/178[#178])
+
+=== Functional Changes
+
+- Implementations must now include a `URL` Converter, of `@Priority(1)` (link:https://github.com/eclipse/microprofile-config/issues/181[#181])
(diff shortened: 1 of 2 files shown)


spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -0,0 +1,42 @@
...
+[[release_notes_11]]
+== Release Notes for MicroProfile Config 1.1
+
+The following changes occurred in the 1.1 release, compared to 1.0
+
+A full list of changes may be found on the link:https://github.com/eclipse/microprofile-config/milestone/2?closed=1[MicroProfile Config 1.1 Milestone]
+
+=== API Changes
+
+The API for MicroProfile Config has not changed with 1.1.
```
Emily-Jiang: Do we need to list the individual api at all? The following paragraph is sufficient.

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -0,0 +1,42 @@
...
+
+The `ConfigSource` SPI has been extended with a default method that returns the property names for a given `ConfigSource`.  (link:https://github.com/eclipse/microprofile-config/issues/178[#178])
+
+=== Functional Changes
+
+- Implementations must now include a `URL` Converter, of `@Priority(1)` (link:https://github.com/eclipse/microprofile-config/issues/181[#181])
+- The format of the default property name for an injection point using `@ConfigProperty` has been changed to no longer lower case the first letter of the class.  Implementations may still support this behavior.  Instead, MicroProfile Config 1.1 expects the actual class name to be used. (link:https://github.com/eclipse/microprofile-config/issues/233[#233])
```
Emily-Jiang: How about changing "MicroProfile Config 1.1 expects" to "MicroProfile Config 1.1 **requires**"

OndrejM: IMHO, "expects" is fine, leaving space for implementations to support the old behavior. If it was "requires", then it would imply that even if the old configuration key is present, the new key is still required, which doesn't make sense to me.

spec/src/main/asciidoc/release_notes.asciidoc
OndrejM: I guess we don't, if there's no breaking change, but why not if it's already written?

johnament: The individual APIs are not listed, so I'm not sure what you mean.  I do suspect that these three sections should be listed within each release.

Emily-Jiang: This sentence above said no change in MP Config. In the following sentence said there was a change in ConfigSource. I did not know whether the sentence _"The API for MicroProfile Config has not changed with 1.1."_ meant there was no change in Config.java or you meant Config specification. It cannot be later as there was a change in ConfigSource.java. My suggestion is to remove the sentence completely.

OndrejM: I guess John meant that the public API wasn't change, only SPI has some changes. I would just remove the above sentence, because the sentence below explains that the only change is in the `ConfigSource` and that's enough.

johnament: So you want to remove the entire section?

johnament: @Emily-Jiang I don't see how you're reading that.  "MicroProfile Config" is the proper name for this component.

Emily-Jiang: @johnament I was trying to find some reason for link this sentence in question with the following sentence about what change goes in. Shall we just remove the following sentence ("The API for MicroProfile Config has not changed with 1.1.") and everything will be clear.

johnament:
@OndrejM @Emily-Jiang let me know if this works.

PR #241, opened 2017-09-07
Create change log 

smillidge:
Creates a simplified log of important changes in Config 1.1. This captures the important changes and is much simpler than browsing the milestone issues or the commit log.


smillidge:
Doh just seen John's PR

Issue #249, opened 2017-09-25
OSGi support for CDI 2.0 in API


arjantijms:
Currently the OSGi headers that are generated for the API artefact have a hard exclusion for CDI 2.0. This means in an environment where CDI 2.0 is available, the requirement for a CDI version that's less than 2.0 generally can't be satisfied and the config API won't work.

The problem can be seen as follows:

```bash
unzip -p microprofile-config-api.jar META-INF/MANIFEST.MF
```

```
Manifest-Version: 1.0
Bundle-License: Apache License, Version 2.0
Bundle-SymbolicName: org.eclipse.microprofile.config
Built-By: johnament
Bnd-LastModified: 1504820572283
Bundle-ManifestVersion: 2
Import-Package: javax.enterprise.util;version="[1.1,2)",javax.inject,o
 rg.eclipse.microprofile.config;version="[1.0,2)",org.eclipse.micropro
 file.config.spi;version="[1.1,2)"
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
Tool: Bnd-3.4.0.201707252008
Export-Package: org.eclipse.microprofile.config;version="1.0";uses:="o
 rg.eclipse.microprofile.config.spi",org.eclipse.microprofile.config.i
 nject;version="1.0";uses:="javax.enterprise.util,javax.inject",org.ec
 lipse.microprofile.config.spi;version="1.1";uses:="org.eclipse.microp
 rofile.config"
Bundle-Name: MicroProfile Config Bundle
Bundle-Version: 1.1.0
Created-By: 1.8.0_101 (Oracle Corporation)
Build-Jdk: 1.8.0_101
```

The problem here is in `Import-Package: javax.enterprise.util;version="[1.1,2)"`

The `2` should become a `3`.

OndrejM:
I don't know it OSGi allows it but there should be no upper boundary on version name - CDI is a JSR therefore every new version should be backward compatible with 1.2 and therefore should work with MP.

arjantijms:
I agree, that could be another option, just don't specify an upper bound at all. I think that's supported indeed.

Emily-Jiang:
[...]

PR #250, opened 2017-09-25
#249 Update bnd file to support CDI 1.1 to 2.x

arjantijms:


Emily-Jiang:
I don't think we should do the changes right now. CDI 2.0  is not supported in MicroProfile as yet.

Emily-Jiang:
Besides, the import package was worked by bnd. When MP pulls in CDI 2.0, the pom.xml will be updated with the version range on cdi dependencies, which will then generate the version range correctly in the manfest.mf. We should not change bnd.

arjantijms:
Indeed, when you change the pom to CDI 2.0, the MANIFEST.MF will reflect that. 

The question is though, for now, do you actually want to bail out when the CDI implementation happens to be capable of supporting CDI 2, even though it in fact does support CDI 1.2 as well?

I.e. Weld 3.0 is 'reported' as CDI 2.0, but supports CDI 1.2 just perfectly fine. Only MP config will not run on it, since it explicitly says not to run on CDI 2.0.

Without this change, GlassFish 5 and Payara 5 will not be able use MP config, even though they could. The limitation seems artificial, unless I'm missing something ;)

OndrejM:
I'm sure we have discussed support of CDI 2.0 (I think that @johnament  raised this a while ago on the mailing list). And I'm pretty sure that we decided that MP should run with future versions of any JSR as they are backwards compatible. Therefore MP should run with CDI 2.0 although it doesn't use any of its new features.

Emily-Jiang:
Not really. In the last discussion, we decided not to auto enable cdi 2.0. I think the big enablement will come when we do MP 2.0, which should include all JavaEE8 corresponding features, JAX-RS 2.1, JSON-P 1.1, CDI 2.0. As a part of MP 2.0 release, all the current MP specs need to be updated to support the new specs. All other specs need to update their dependencies on CDI, not just this one.

OndrejM:
@Emily-Jiang this isn't about updating dependencies for MP config - the dependency stays at CDI 1.2. But the OSGi manifest restricts any future version of CDI so it works only with a CDI implementation that provides CDI 1.2 but not with a CDI 2.0 implementation, which also provides CDI 1.2 which is a subset.

I'm strongly against such restrictions because they slow down innovation. I don't see a reason why an MP implementation can't decide to upgrade the CDI component to support CDI 2.0, while still being compatible with MP, which requires CDI 1.2 but should allow later versions.

In this [thread](https://groups.google.com/d/msg/microprofile/8Ll4PQ4Qvp8/ehty_7B0AAAJ) everybody agrees that MP should require CDI 1.2 but allow CDI 2.0. In your own words, @Emily-Jiang: " If a container has CDI 2.0 implementation, it should work with microprofile 1.0 release. "

The problem is that if a container is OSGi based and has CDI 2.0 impl, it wouldn't work with the current MP config API artifact, because the OSGi manifest doesn't allow CDI 2.0

struberg:
Cdi will remain backward compatible.
We could easily go for [1.1,].
Means anything equals or bigger cdi 1.1

LieGrue,
Strub

smillidge:
+1 this should be merged into the next version

Emily-Jiang:
hang on. Please see my previous comments. The change is better to be handled in the pom.xml not here. Besides, the PR failed IP-validation.

api/bnd.bnd
```diff
@@ -1,5 +1,10 @@
 -exportcontents: \
     org.eclipse.microprofile.*
+
+Import-Package: \
```
Emily-Jiang: The version range should be handled in pom.xml not here.

arjantijms: How would you propose handling it in pom.xml? Via the `org.apache.felix:maven-bundle-plugin` or something else?

Emily-Jiang: @arjantijms  yes, in maven-bundle-plugin. I did experiment a bit with pom.xml. It seems it is not as visible as specifying in the bnd.bnd. sorry, I had a false impression. What you have done is good. Actually, if cdi api was properly semantic versioned, we won't have this headache. This is why I campaign for MP to do semantic versioning properly. If you fix your ip-validation, your change can be merged in.

Emily-Jiang:
@arjantijms can you fix the ip-validation error so that your PR can be merged?

arjantijms:
Sorry, forgot about this one. I think I'll just have to redo the PR using my payara.fish email address. Will check for that later today. Thanks for the reminder!

arjantijms:
PR with signed, verified and signed-off commit using @payara.fish email done at: https://github.com/eclipse/microprofile-config/pull/250

Issue #259, opened 2017-11-20
Built in converter - Array


Emily-Jiang:
@Emily-Jiang commented on [Tue Oct 31 2017](https://github.com/eclipse/ConfigJSR/issues/14)

We should add Array in our builtin converter. Archius has this support and it is universally useful.

---

@TFaga commented on [Wed Nov 01 2017](https://github.com/eclipse/ConfigJSR/issues/14#issuecomment-340951039)

+1 for supporting Arrays/Lists. 

It's quite common to have lists in the configuration, and we should have an out of the box support for them. However, we should consider if a converter is the correct choice for this. When retrieving a list we would most likely like to specify the concrete type of the List (e.g. List<String>, ...) we want to retrieve. With the way the current `Config` API methods are defined (`getValue`), you cannot pass in a class with a concrete generic type as the `propertyType` argument due to type erasure. You can only pass the base `List` class, which will either cause an "Unchecked assignment" warning by the compiler (if we assign the result to `List<String>`) or the returned list won't be parameterised or parsed as we would like (it would be `List<Object>`).

For CDI injecting of the config Lists however, it would work great as generic type information is preserved in class fields.

---

@atsticks commented on [Wed Nov 01 2017](https://github.com/eclipse/ConfigJSR/issues/14#issuecomment-341086048)

There might be other aspects, which I think is worth thinking about. To illustrate image the following example:
**ConfigSource 1**
`list=a,b`

**ConfigSource 2**
`list=c,d`

What is the final array value, if `ConfigSource1.ordinal < ConfigSource2.ordinal` ?

**Options**
```
list=c,d       // override
list=a,b,c,d // combining
```
To have more flexibility than always using the *overriding* mechanics, in Tamaya we have the [`PropertyValueCombinationPolicy `](https://github.com/apache/incubator-tamaya/blob/master/code/api/src/main/java/org/apache/tamaya/spi/PropertyValueCombinationPolicy.java)  added.

How about supporting also:
* **Sets** (would work OOTB since it could be done by the corresponding converters by implicitly removing any duplicates)
* **Maps** 

---

@struberg commented on [Wed Nov 01 2017](https://github.com/eclipse/ConfigJSR/issues/14#issuecomment-341098778)

We had this discussion in MP config as well. We ended up intentionally not supporting it as it adds tons of complexity. 

In DeltaSpike Config there is support for simple lists via asList() 
http://deltaspike.apache.org/documentation/configuration.html#_list_values_handling
[...]

PR #260, opened 2017-11-23
#259 support Array as builtin converter, pls review

Emily-Jiang:


api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -62,7 +62,20 @@
  * All Built In Converters have a {@code javax.annotation.Priority} of 1
  * A Converter should handle null values returning either null or a valid Object of the specified type.
  *
- *  *
+ * <h3>Array Converters</h3>
```
johnament: I like the idea, but I think for this to be concrete we have to update the spec as well.  The TCK should be updated to test this for all of the current types.

If we are going to include custom converters, we need to enumerate what the separators are.  We would also need to a way to escape the values.  For instance if you took your change and did instead:

```
tck.config.test.javaconfig.converter.array.pizza=large:ham,pineapple,medium:chicken,small:pepperoni
```

I'm trying to convey that the large pizza should have both ham and pineapple toppings.  There should be a way to make those two go together.

Emily-Jiang: @johnament I like your comments here:). Yes, I will change the spec about the delimiter (,) and escape character(\).

Emily-Jiang: More commits to come to this one as I have done this to demonstrate to ConfigJsr EG.

OndrejM: I agree with @johnament - the behavior of the converter should be specified in the spec because config sources can be provided also by the application and it should be clear how the array converter would process string values. More concretely, things like separators, escape sequence for a separator and converting string chunks into Java objects should be covered.

Emily-Jiang: yep. agreed. More commits to come.

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -60,6 +60,33 @@ The `Config` will use the `Converter` with the highest `Priority` for each targe
 
 A custom `Converter` for a target type of any of the built-in Converters will overwrite the default Converter.
 
+=== Array Converters
+For the built-in converters and custom converters, the corresponding Array converters are provided
```
johnament: (sorry for the extremely long comment)
Just as a thought, something to mull over.  Yes, we can provide built in support for collections for custom types.  Or we can expect that the converter implementation registers additional converters for those types.  For the built in types, what you're describing makes a lot of sense.  However, when we look the format, do we expect that the entire string is passed in to the converter, or only the segment we identified as being one value?  I wonder if it makes sense for any of the following cases as well:

- The user passes in (via API changes) what the delimiter is
- The converter takes in the raw string value (like every other converter) and returns a `Collection<E>` or `E[]` type, choosing what it knows the delimiter to be (or even trying multiple)
- Now that we're looking at collection types, Do we still only look at the highest config source for this value?  Since right now the config impl is doing the splitting, it may make sense to clarify this
- What happens if in Config 1.0/1.1 a user already registered a converter of type `List` and we introduce this change? Will their converter break?
- `Set<String>` and `List<String>` aren't classes, they're `ParameterizedType`s, as a result they can't be passed into the current API method (`getValue`)

Emily-Jiang: What we are trying to do is to expand the current design to cope with more capabilities. With specifying delimitor will introduce API changes. My view is that if a different parameter is needed, a custom written converter can be developed.
>Now that we're looking at collection types, Do we still only look at the highest config source for >this value? Since right now the config impl is doing the splitting, it may make sense to clarify this
The introduction of this new converter should not change the priority on config source. The converter is no different from other builtin converters.

>What happens if in Config 1.0/1.1 a user already registered a converter of type List and we introduce this change? Will their converter break?

No. All builtin converters have the priority of -1. All user defined ones should have a higher priority.

>Set<String> and List<String> aren't classes, they're ParameterizedTypes, as a result they can't >be passed into the current API method (getValue)

This is not supported in programmatic look but only supported in injection model. In the spec, I only mention List and Set will work in the injection model. I will add one more sentence explicitly to say programmatic look up is not support.

Emily-Jiang:
I will merge this one and then @jmesnil has done more fixes on the top of this one under #279

Issue #265, opened 2017-11-30
Overriding the ConfigProviderResolver doesn't work right


johnament:
This issue isn't 100% consistent, and I'm seeing different behavior in a running app vs a unit test so I'm a bit miffed.  I first found this in Open API, see https://github.com/eclipse/microprofile-open-api/issues/80 and related PR https://github.com/eclipse/microprofile-open-api/pull/85

I'm not sure at this point if its an issue in other specs, but this unit test shows the problem.  If you've already called `getConfig()` (at least in some cases) you can cause the wrong config to be loaded after overriding the resolver.  I can get this to happen differently in a running app, where the value seems to have already been loaded without calling `getConfig()`, I'm guess its an oddity between how TestNG runs via Surefire and standalone apps.

johnament:
No, I see it locally in a running app without calling `getConfig`.  I added that, as it was the only way to ensure that testng loaded that class first.

The problem happens when the static field is initialized before the `setInstance` is called.

OndrejM:
I see, so the first step isn't caling `getConfig()` but when the `ConfigProvider` class is loaded by a classloader. A simple fix would be to refactor the initialization into the `getConfig()` method to so that the probem happens only in the scenario with calling `getConfig()`

johnament:
I mirrored what you did in rest client (and what others did in their specs) in my config pr https://github.com/eclipse/microprofile-config/pull/266

OndrejM:
The difference to the rest client is that `Config` is supposed to be a singleton while rest client is supposed to be always a new instance. That's why the methods are called `getConfig()` and `newBuilder()`

OndrejM:
If we want to get config from the resolver every time `ConfigProvider.getConfig()` is called we have to ensure that the old config is properly released. The best way to do it is to keep the reference to it in the resolver so that all instances can be released when classloader is released. Currently, the spec expects a single Config instance per classloader.

OndrejM:
Look at RestClientBuilderResolver.instance in REST client, this is how Config singleton should be retrieved from Config resolver in the `ConfigProvider.getConfig` method. This is a simple solution that fixes your problem when you don't call `getConfig` before `setInstance`.

PR #266, opened 2017-11-30
Provider override behavior

johnament:


api/src/main/java/org/eclipse/microprofile/config/ConfigProvider.java
```diff
@@ -86,7 +84,7 @@ private ConfigProvider() {
      * @return the config object for the thread context classloader
      */
     public static Config getConfig() {
-        return INSTANCE.getConfig();
+        return ConfigProviderResolver.instance().getConfig();
```
OndrejM: As explained in https://github.com/eclipse/microprofile-config/issues/265#issuecomment-348287580, this should be changed to call `ConfigProviderResolver.instance().getConfig();` only once because `Config` is expected to be a singleton and has to be released later. 

Here's an example of the code:
```
        if (INSTANCE == null) {
            synchronized (RestClientBuilderResolver.class) {
                if (INSTANCE != null) {
                    return INSTANCE;
                }
                INSTANCE = ConfigProviderResolver.instance();
            }
        }

        return INSTANCE;
```

OndrejM: Or the same can be done with simpler code by moving the INSTANCE field and initialization of `private static final ConfigProviderResolver INSTANCE = ConfigProviderResolver.instance();` into a separate class which would be used only from `getConfig()` methods, like this:

ConfigProvider.java:
```
public static Config getConfig() {
         return ConfigSingletonHolder.INSTANCE.getConfig();
```
new ConfigSingletonHolder.java:
```
class ConfigSingletonHolder {
   static final ConfigProviderResolver INSTANCE = ConfigProviderResolver.instance();
}
```

This subtle refactoring would mean that the initialization of INSTANCE happens when ConfigSingletonHolder class is loaded and that happens when `getConfig` method is called and no sooner. But few people understand how it works so it's probably better to use the explicit code with locks and null checks which is much more straightforward. Unfortunately much harder to get it right and thread-safe.

johnament: the classloader behavior you mention is probably why in the test I had to add forced loading the class.  I suspect the CDI runtime in the running application causes the class to be loaded, resulting in the field being set before it was meant to (I believe the user in my case was setting it via a CDI extension, but need to confirm).

api/src/test/java/org/eclipse/microprofile/config/ConfigProviderTest.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
```
tevans78: Wrong date?

Issue #269, opened 2017-12-07
Revisit Common sense converter 


Emily-Jiang:
https://github.com/eclipse/microprofile-config/issues/248
It added the following support:
=== Automatic Converters

If no built-in nor custom `Converter` for a requested Type `T`, an implicit Converter is automatically provided if the following conditions are met:

* The target type `T` has a Constructor with a String parameter, or
* the target type `T` has a Constructor with a CharSequence parameter, or
* the target type `T` has a `static T valueOf(String)` method, or
* the target type `T` has a `static T valueOf(CharSequence)` method, or
* the target type `T` has a `static T parse(String)` method, or
* the target type `T` has a `static T parse(CharSequence)` method, or
* the target type `T` is an `Enum`, in which case the `Enum#valueOf(Class, String)` is being used.

In the original discussion of #248, I only suggest to add the support 
The target type `T` has a Constructor with a String parameter
the target type `T` has a `static T valueOf(String)` method,

The above will cover most scenarios and it is straightforward.

The current codebase introduces too many variety and we need to specify the search order. I feel this complicates the spec without much gain. What do people think on removing the CharSequence support, parse and Enum. Besides, we are providing a convenient method. If anyone wants to use CharSquence, they can provide their own converter. Thoughts?

johnament:
Please note:

- Fault Tolerance requires the `Enum` type since it doesn't specify that a converter supporting the enums is required.
- In addition, the `Enum` case should be even simpler since `T.valueOf` gets created automatically from any `enum` type.
- The intent of #248 was to change [Pizza.java](https://github.com/eclipse/microprofile-config/blob/1.1/tck/src/main/java/org/eclipse/microprofile/config/tck/converters/Pizza.java#L30) from having a single arg constructor to a 2 arg constructor only (plus change the converter).
- I believe we were capturing ideas for a future release to add common sense conversion to the spec

I would recommend with limiting support to just `String` and not `CharSequence` since `String` is the raw type we use in config without any other requirements.

Issue #248, opened 2017-09-21
Don't discriminate against common sense converters


johnament:
I was trying to figure out how far off Apache Tamaya was from implementing the feature set.  I noticed one test kept failing, even though I knew it wasn't registering the converter (`AutoDiscoveredConfigSourceTest.testAutoDiscoveredConverterNotAddedAutomatically` is the test).  It fails in Tamaya because Tamaya has the notion of "common sense" conversion (at least that's my term).  If the target class has a string based constructor, or a static `valueOf` method that takes a string, Tamaya will automatically convert that type.  It doesn't look at the converter on the classpath instead just loads a default converter strategy.

So I'm wondering, should we perhaps use a more explicit test that should fail for unknown reasons rather than a converter that has too simple of a use case (e.g. the class takes 1 string but really should take 2 strings since there are two values passed in).

Emily-Jiang:
+1. I think we should add this support of string based constructor and valueOf to the spec. Archaius has this notion of "common sense" as well.

Emily-Jiang:
In order to get Tamaya to pass tck, I'll raise a discussion to get a process to exclude error test cases. Thoughts?

smillidge:
+1 to add common sense conversion to the spec

struberg:
fixed the TCK and added it as an explicit requirement.

PR #275, opened 2017-12-11
#269 rework #248

Emily-Jiang:


Emily-Jiang:
@jmesnil please review. Please let me know whether the tcks run ok against your impl. thanks

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -65,9 +65,5 @@ A custom `Converter` for a target type of any of the built-in Converters will ov
...
 
 * The target type `T` has a Constructor with a String parameter, or
-* the target type `T` has a Constructor with a CharSequence parameter, or
-* the target type `T` has a `static T valueOf(String)` method, or
-* the target type `T` has a `static T valueOf(CharSequence)` method, or
-* the target type `T` has a `static T parse(String)` method, or
-* the target type `T` has a `static T parse(CharSequence)` method, or
-* the target type `T` is an `Enum`, in which case the `Enum#valueOf(Class, String)` is being used.
+* the target type `T` has a `static T valueOf(String)` method
+If the target type `T` is an `Enum`, in which case the `Enum#valueOf(Class, String)` is being used.
```
jmesnil: This is not tested by the TCK

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -69,12 +69,7 @@
  * the {@code Config} provides an <em>Implicit Converter</em>, if</p>
  * <ul>
  *     <li>The target type {@code T} has a Constructor with a String parameter, or</li>
- *     <li>The target type {@code T} has a Constructor with a CharSequence parameter, or</li>
  *     <li>the target type {@code T} has a {@code static T valueOf(String)} method, or</li>
- *     <li>the target type {@code T} has a {@code static T valueOf(CharSequence)} method, or</li>
- *     <li>the target type {@code T} has a {@code static T parse(String)} method, or</li>
- *     <li>the target type {@code T} has a {@code static T parse(CharSequence)} method, or</li>
- *     <li>the target type {@code T} is an {@code Enum}, in which case the {@link Enum#valueOf(Class, String)} is being used.</li>
  * </ul>
```
jmesnil: The line about Enum should be kept

spec/src/main/asciidoc/converters.asciidoc
jmesnil: I opend #278 to test implicit Enum conversion

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
Emily-Jiang: We discussed this and John suggested to remove it as it is covered by the valueOf. @johnament commented: Technically, this last case isn't needed.  When an Enum type is created, it automatically gets a T.valueOf method.
It is implementation detail. The enum is covered under valueOf.

jmesnil: ok, that's fine.

PR #278, opened 2017-12-12
[TCK] add test for implicit Enum converter

jmesnil:
Signed-off-by: Jeff Mesnil <jmesnil@gmail.com>

PR #280, opened 2017-12-13
[TCK] add test for implicit Enum converter

jmesnil:


jmesnil:
this supersedes #278

PR #286, opened 2017-12-15
add parse(CharSequence) support for the implicit converter

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
@@ -97,7 +97,7 @@ package org.eclipse.microprofile.config.spi;
  * <ul>
  *     <li>The target type {@code T} has a Constructor with a String parameter, or</li>
  *     <li>the target type {@code T} has a {@code static T valueOf(String)} method, or</li>
- *     <li>the target type {@code T} has a {@code static T parse(CharSequence)} method, or</li>
+ *     <li>the target type {@code T} has a {@code static T parse(CharSequence)} method</li>
  * </ul>
 
  * @author <a href="mailto:rsmeral@apache.org">Ron Smeral</a>
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
@@ -97,6 +97,7 @@ package org.eclipse.microprofile.config.spi;
  * <ul>
  *     <li>The target type {@code T} has a Constructor with a String parameter, or</li>
  *     <li>the target type {@code T} has a {@code static T valueOf(String)} method, or</li>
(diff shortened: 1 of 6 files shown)


tck/src/main/java/org/eclipse/microprofile/config/tck/ImplicitConverterTest.java
```diff
@@ -78,6 +79,14 @@ public void testImplicitConverterStringValueOf() {
         Assert.assertNotNull(value);
         Assert.assertEquals(value.getVal(), "stringValueOf");
     }
+    
+    @Test
+    public void testImplicitConverterCharSequenceParse() {
+        ConvTestTypeWCharSequenceParse value = config.getValue("tck.config.test.javaconfig.converter.implicit.charSequenceParse",
```
jmesnil: let's use an actual class (e.g. java.time.Year). We don't need to create a weirdly named class to check that implicit converter for Java runtime classes work

Emily-Jiang: added more tests to cover your usage.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -97,6 +97,7 @@
  * <ul>
  *     <li>The target type {@code T} has a Constructor with a String parameter, or</li>
  *     <li>the target type {@code T} has a {@code static T valueOf(String)} method, or</li>
+ *     <li>the target type {@code T} has a {@code static T parse(CharSequence)} method, or</li>
```
johnament: Unneeded or

tck/src/main/java/org/eclipse/microprofile/config/tck/ImplicitConverterTest.java
```diff
@@ -21,9 +21,13 @@
 
 import static org.eclipse.microprofile.config.tck.base.AbstractTest.addFile;
 
+import java.time.YearMonth;
```
johnament: This test is also broken, ProcessAnnotatedType is not fired for fields in an Arquillian test.

Emily-Jiang: Which bit is broken?

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
Emily-Jiang: fixed

tck/src/main/java/org/eclipse/microprofile/config/tck/ImplicitConverterTest.java
Emily-Jiang: used the classes to hold the field injection instead. This must have something to do with the Arquillian integrator you are using.

PR #288, opened 2017-12-16
Adding spec and TCK to support built in class converters

### LINKED ISSUE #267: Built in provider for Class types
This is needed in both the fault tolerance and the rest client specs.

There should be a built in way to convert `Class` objects (and likewise the `Class[]` class)

johnament:
This closes #267

### DIFF
--- a/spec/src/main/asciidoc/converters.asciidoc
+++ b/spec/src/main/asciidoc/converters.asciidoc
@@ -44,6 +44,7 @@ The following `Converter` s are provided by Microprofile-Config by default:
 * `OffsetTime`
 * `Instant`
 * `URL`
+* `Class` based on the result of `Class.forName`
 
 All built-in `Converter` have the `@Priority` of `1`.
 
@@ -72,7 +73,7 @@ e.g. With this config `myPets=dog,cat,dog\\,cat`, the values as an array will be
  
 
 ----
- String[] myPets = ConfigProvider.getValue("myPet", String[]);
+ String[] myPets = ConfigProvider.getValue("myPet", String[].class);
 ----
  myPets will be "dog", "cat", "dog,cat" as an array
  
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ClassConverterTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ClassConverterTest.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ *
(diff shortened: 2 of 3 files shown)


spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -72,7 +73,7 @@ e.g. With this config `myPets=dog,cat,dog\\,cat`, the values as an array will be
  
 
 ----
- String[] myPets = ConfigProvider.getValue("myPet", String[]);
+ String[] myPets = ConfigProvider.getValue("myPet", String[].class);
```
johnament: This was a simple typo I noticed going through the spec.

jmesnil:
Is it a good idea to add built-in class converters with the module changes in Java 9?
I'd expect that the adoption of Java 9 would promote the use of `forName(Module, String)`.

tck/src/main/java/org/eclipse/microprofile/config/tck/ClassConverterTest.java
```diff
@@ -0,0 +1,110 @@
...
+        assertEquals(classConverterBean.testClassSet, new LinkedHashSet<>(Arrays.asList(ClassConverterTest.class, String.class)));
+        assertEquals(classConverterBean.testClassList.size(), 2);
+        assertEquals(classConverterBean.testClassList, Arrays.asList(ClassConverterTest.class, String.class));
+    }
+
+    @Dependent
+    private static class ClassConverterBean {
+        @Inject
+        @ConfigProperty(name = "tck.config.test.javaconfig.converter.class")
+        private Class<?> testClass;
```
jmesnil: @johnament Do you have an implementation that passes the test?

In [http://github.com/wildfly-extras/wildfly-microprofile-config](wildfly-microprofile-config) that is using Weld to run the TCK, the test fails with:

```
WELD-001408: Unsatisfied dependencies for type Class<?>[] with qualifiers @ConfigProperty
  at injection point [BackedAnnotatedField] @Inject @ConfigProperty private org.eclipse.microprofile.config.tck.ClassConverterTest$ClassConverterBean.testClasse
s
  at org.eclipse.microprofile.config.tck.ClassConverterTest$ClassConverterBean.testClasses(ClassConverterBean.java:0)
```

If I update the test to remote the wildcard from all injected types (using instead `private Class testClass`), the test passes.

Emily-Jiang: I cannot get this test pass either.

Issue #287, opened 2017-12-16
ArrayConverterTest is broken


johnament:
The TCK test ArrayConverterTest is broken.  It assumes that a ProcessAnnotatedType is fired for fields in an Arquillian test, which is not the case.

johnament:
I'm not sure if we have this with other new tests, but previously we set these fields on beans directly, instead of in the test class.

The formatting of this class is also a bit hard to read.

Emily-Jiang:
@jmesnil ran the tests against his impl and the tests passed ok. This might have something to do with the arquillian integrator, maybe. Will you be able to do a PR to improve the tests?

PR #290, opened 2017-12-18
Declare an explicit bean for array types (#287)

johnament:


jmesnil:
This PR fixes the ImplicitConverterTest fixture and I confirm that the updated ArrayConverterTest passes with wildfly-microprofile-config implementation

johnament:
Yeah, heh.  The implicit fix was just something I noticed running against the gconfig impl.

If all good I'm going to merge this.

PR #296, opened 2017-12-19
improve array tests

Emily-Jiang:
Signed-off-by: Emily Jiang <emijiang@uk.ibm.com>

PR #297, opened 2017-12-20
[TCK] Fix ArrayConverterTest

jmesnil:
This supersedes #296


jmesnil:
@Emily-Jiang this PR includes your commit for #296 and fixes some assertions that were incorrect.
This new ArrayConverterTest passes with wildfly-microprofile-config implementation

PR #308, opened 2018-01-23
Move static nested CDI beans out into their own classes

### LINKED ISSUE #307: Static nested classes in TCK lead to ambiguous CDI resolution errors
```
WELD-001409: Ambiguous dependencies for type ClassConverterBean with qualifiers @Default
--
at injection point [BackedAnnotatedField] @Inject private org.eclipse.microprofile.config.tck.ClassConverterTest.classConverterBean
at org.eclipse.microprofile.config.tck.ClassConverterTest.classConverterBean(ClassConverterTest.java:0)
Possible dependencies:
- Managed Bean [class org.eclipse.microprofile.config.tck.ClassConverterTest$ClassConverterBean] with qualifiers [@Any @Default],
- Managed Bean [class org.eclipse.microprofile.config.tck.ClassConverterTest$ClassConverterBean] with qualifiers [@Any @Default]
 
```

On both [1] and [2] the usage of static inner bean classes, that are added to `WEB-INF/lib` (see `@Deployment`) as well implicitly by arquillian, the TCK tests fail with an exception. 

The most simple fix would be to turn these bean classes into regular classes, like in the other tests.

[1] https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/ClassConverterTest.java

[2] https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/ImplicitConverterTest.java

tevans78:
fixes #307

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ClassConverterBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ClassConverterBean.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2016-2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


Emily-Jiang:
LGTM! Thanks @tevans78 !

PR #309, opened 2018-01-23
Make the string constructor public

### LINKED ISSUE #305: TCK: inaccessible constructor used for converter test
ImplicitConverterTest.testImplicitConverterStringCt() uses class ConvTestTypeWStringCt that has a package local constructor only.
This will be a problem with Java 9 and in general doesn't make sense - there is usually a reason that a constructor is hidden and we may cause trouble if we "magically" called it from config.
I think this constructor should be public (and the spec should probably mention that only public implicit converters should be used).

tevans78:
fixes #305


Emily-Jiang:
thanks @tevans78 !

PR #311, opened 2018-01-25
Don't test system props in environment test

### LINKED ISSUE #310: TCK: ConfigProviderTest.testEnvironmentConfigSource() is not a reliable test
`testEnvironmentConfigSource()` uses an injected Config which has default ConfigSources... i.e both Environment variables and System properties. The ordinal values of those default sources (300 & 400 respectively) mean that if a property exists as a system property it will mask the corresponding environment variable. For the most part those values tend to be the same but there is no guarantee and the method does not safely test the things it is supposed to.

A possible fix might be to only test those properties which are unique to the environment (i.e. not system properties) but it still doesn't feel like a good test. One would also have to do the opposite for `testPropertyConfigSource()`

Alternatively, it might be better to just test a single unique property ... but I'm not sure where the TCK could set a unique environment variable?

tevans78:
fixes #310 

Still not a great test

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
@@ -42,6 +42,7 @@ import org.jboss.shrinkwrap.api.asset.EmptyAsset;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
 /**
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
```diff
@@ -71,8 +71,14 @@ public static WebArchive deploy() {
...
     public void testEnvironmentConfigSource() {
         Map<String, String> env = System.getenv();
+        Properties properties = System.getProperties();
         for (Map.Entry<String, String> envEntry : env.entrySet()) {
-            Assert.assertEquals(envEntry.getValue(), config.getValue(envEntry.getKey(), String.class));
+            String key = envEntry.getKey();
+            if(!properties.containsKey(key)) {
+                String value = envEntry.getValue();
+                String cfgValue = config.getValue(key, String.class);
+                Assert.assertEquals(value, cfgValue);
```
Emily-Jiang: Need to verify whether this line is actually hit, so that we tested something.

hutchig: What about testing that a property DOES override the env value if present - which would match what the impl has to do more closely?

Emily-Jiang: That test will be to test system property override environment variables. It is nice to have but I don't know whether it is reliable or not, as I think in a lot of environments, the variables are set to the same value.

PR #313, opened 2018-01-26
#305 clarify only public constructor used as an implicit converter

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
@@ -95,7 +95,7 @@ package org.eclipse.microprofile.config.spi;
  * <p>If no explicit Converter and no built-in Converter could be found for a certain type,
  * the {@code Config} provides an <em>Implicit Converter</em>, if</p>
  * <ul>
- *     <li>The target type {@code T} has a Constructor with a String parameter, or</li>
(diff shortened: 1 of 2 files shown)


Emily-Jiang:
@jmesnil please take a look. I deleted the enum mentioning as it is redundant. There is a automatic method vauleOf being generated for Enum object.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -95,7 +95,7 @@
  * <p>If no explicit Converter and no built-in Converter could be found for a certain type,
  * the {@code Config} provides an <em>Implicit Converter</em>, if</p>
  * <ul>
- *     <li>The target type {@code T} has a Constructor with a String parameter, or</li>
+ *     <li>The target type {@code T} has a public Constructor with a String parameter, or</li>
```
jmesnil: In that case, we should also state that public method valueOf and parse are supported (implementations can also check private methods but it is not portable)

struberg: +1 that's what we talked through in yesterdays ConfigJSR meeting and likely applies for mp-config as well.

Emily-Jiang: prefer to clearly mention all of them are public. The implementors can support private, which is beyond the spec's requirement. The spec does not disallow it.

PR #324, opened 2018-03-02
Uri converter

johnament:
This introduces a new built in required URI converter.  This is needed by rest client.

I did a separate commit to clean up the formatting on `ArrayConverterBean`, it was a bit hard to follow.


Emily-Jiang:
The URI converter should be already covered by the default converter as it has a constructor taking a single string. The next step I am going to do is to remove all built in converters instead of adding any more.

johnament:
Looking at github, I don't see your requested changes.  But you want to revert the javadoc and spec changes in favor of claiming that they just work because of implicit converters?

The difference I think is that `URI.create`throws an `IllegalArgumentException` whereas `new URI()` throws `URISyntaxException`.

Emily-Jiang:
@johnament for implementation, the call new URI(string) etc should catch the exception and wrap up with IllegalArgumentException, which should be already there in the part of implicit converter support.

struberg:
+1 for relying on implicit converter in this case.

Issue #264, opened 2017-11-29
define a mapping rule from config property name to environment variable


Emily-Jiang:
@Emily-Jiang commented on [Wed Nov 15 2017](https://github.com/eclipse/ConfigJSR/issues/24)

As the name of environment variable can only use alphabets and _ , it is difficult to override some config properties as environment variables.

e.g.
com.acme.me="blah"

This property cannot be specified as environment variable. How about to directly translate . to _ for the corresponding environment variables. 

com.acme.me -> com_acme_me
In this way, this property can be defined or override in environment variable.

---

@struberg commented on [Wed Nov 15 2017](https://github.com/eclipse/ConfigJSR/issues/24#issuecomment-344747644)

+1 to add it explicitly!

been there, done that 2 years ago ;) 

https://issues.apache.org/jira/browse/DELTASPIKE-1090
https://github.com/apache/deltaspike/blame/master/deltaspike/core/impl/src/main/java/org/apache/deltaspike/core/impl/config/EnvironmentPropertyConfigSource.java#L55

---

@kazumura commented on [Wed Nov 15 2017](https://github.com/eclipse/ConfigJSR/issues/24#issuecomment-344754084)

That means you cannot use '_' as a property key.

```
e.g.
 foo._bar
 foo_.bar
 foo..bar
 foo__bar

 Are all these same ?
```

I prefer to add a flag to specify auto converting or not.

---

@struberg commented on [Wed Nov 15 2017](https://github.com/eclipse/ConfigJSR/issues/24#issuecomment-344756004)

yes, you could not distinguish those in the EnvironmentConfigSource, but everywhere else. 
In practice it's not really an issue. 

We first catched this issue when integrating Docker where ENV=.. can also not contain any dots via scripts.

---

@kazumura commented on [Wed Nov 15 2017](https://github.com/eclipse/ConfigJSR/issues/24#issuecomment-344762543)

I agree the auto converting as the default behavior.

But there are some people who already use '_' as the keys.
My point is, should we provide them an option to use this feature or not?
I don't know how many people already use '_'. May not major.

---

@TFaga commented on [Wed Nov 15 2017](https://github.com/eclipse/ConfigJSR/issues/24#issuecomment-344763165)

+1 for `.` -> `_` as well. We're also doing this in KumuluzEE.

[...]

PR #326, opened 2018-03-14
#264 environment variable mapping rules

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
@@ -93,20 +93,20 @@ public interface ConfigSource {
      <ol>
      * <li>System properties (ordinal=400)</li>
      * <li>Environment properties (ordinal=300)
-     *    <p>Depending on the operating system type, environment variable names may only contain alphabetic or _. 
-     *    Other characters such as ., / etc are not always allowed. 
-     *    In order to set a value for a property with a name containing illegal characters as an environment variable, 
-     *    the following rules are used. 
-     *    This ConfigSource searches 3 environment variables for a given property name (e.g. {@code "com.ACME.size"}):</p>
+     *    <p>Some operating systems allow only alphabetic characters or an underscore(_), in environment variables.
+     *    Other characters such as ., /, etc may be disallowed.
+     *    In order to set a value for a config property that has a name containing such disallowed characters from an environment variable,
+     *    the following rules are used.
+     *    This ConfigSource searches for potentially 3 environment variables with a given property name (e.g. {@code "com.ACME.size"}):</p>
      *        <ol>
      *            <li>Exact match (i.e. {@code "com.ACME.size"})</li>
-     *            <li>Replace all '.' by '_' (i.e. {@code "com_ACME_size"})</li>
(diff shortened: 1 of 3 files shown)


Emily-Jiang:
@jmesnil can you take a look please?

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -90,11 +90,23 @@
...
+     * <li>System properties (ordinal=400)</li>
+     * <li>Environment properties (ordinal=300)
+     *    <p>Some operating systems allow only alphabetic characters or an underscore(_), in environment variables.
+     *    Other characters such as ., /, etc may be disallowed.
+     *    In order to set a value for a config property that has a name containing such disallowed characters from an environment variable,
+     *    the following rules are used.
+     *    This ConfigSource searches for potentially 3 environment variables with a given property name (e.g. {@code "com.ACME.size"}):</p>
+     *        <ol>
+     *            <li>Exact match (i.e. {@code "com.ACME.size"})</li>
+     *            <li>Replace all illegal characters by '_' (i.e. {@code "com_ACME_size"})</li>
```
jmesnil: I don't think mentioning "illegal characters" from the specification is correct.

How would you be able to determine which characters are illegal (from the OS env perspective) in Java to replace them.

I find the wording too open-ended and would prefer to constrain and make explicit which characters will be replaced.

Either we provide a fixed set (`.` is a must, we can also argue about `-` and `/`).

Another possibility is to specify that there can be a special env variable whose value is the list of characters that must be replaced. For example

```
MP_CONFIG_REPLACED_CHARS=".-/😀@"
```

The env config source will then replace any chars from that list by `_`.
By default, the value of MP_CONFIG_REPLACED_CHARS is `.-/`

jmesnil: ah, my review was submitted after you merge the PR so I couldn't vote.
I am `-1` for this change

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -90,11 +90,23 @@
...
+     *            <li>Replace all illegal characters by '_' and convert to upper case (i.e. {@code "COM_ACME_SIZE"})</li>
```
pilhuhn: 👍

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
Emily-Jiang: as I mentioned on the call, I will clarify the meaning of illegal characters - basically any characters not alphabetic nor _ are viewed as illegal characters from environment variable name point of view.

jmesnil: @Emily-Jiang + 1 (although I suppose you mean alphanumeric rather that alphabetic)

Emily-Jiang: @jmesnil yes, you were exactly right. I clarified via https://github.com/eclipse/microprofile-config/pull/327/files

Issue #321, opened 2018-02-21
bnd-baseline-maven-plugin breaks the build


struberg:
It seems that the hardcoded semver major-minor-patch scheme used by the bnd-baseline-maven-plugin doesn't fit well to microprofile.

What mp-config uses is more a product-master-minor scheme.
It seems that the plugin also doesn't properly detect backward compatible version changes. 

We might rather use the animalsniffer-maven-plugin instead.

Emily-Jiang:
@struberg what error message did you get? The plugin is not just for detecting backward compatible changes. It is used to properly version the packages.

PR #328, opened 2018-03-19
#321 fix the version changes and enable the baseline

Emily-Jiang:


Emily-Jiang:
@struberg it is OSGi annotation and it only applies to classes. With the package import, it should be self explanatory. I will also see whether there is a blanket update on asserting the provider type from bnd tools.

PR #337, opened 2018-03-28
Fixes #336

### LINKED ISSUE #336: Seem to be missing a test of injecting an Optional<String> that has not configuration value
I ran across a situation where an injection point like:

```java
 @Inject
 @ConfigProperty(name = "mpjwt.signerPublicKey")
 private Optional<String> publicKeyPemEnc;
```
where there was no mpjwt.signerPublicKey provided, was seen to have a value of org.eclipse.microprofile.config.configproperty.unconfigureddvalue, the ConfigProperty.UNCONFIGURED_VALUE.

This value should never be returned for an Optional injection point, so it is a bug in the config implementation I'm using, but looking at the TCK, I don't see any test of this scenario. The existing CdiOptionalInjectionTest needs to have a test of this scenario.

jmesnil:
Add an injected Optional<String> property with no property value for it
and should be set to Optional.empty

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/CdiOptionalInjectionTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/CdiOptionalInjectionTest.java
@@ -66,5 +66,7 @@ public class CdiOptionalInjectionTest extends Arquillian {
 
         Assert.assertTrue(optionalValuesBean.getStringValue().isPresent());
         Assert.assertEquals(optionalValuesBean.getStringValue().get(), "hello");
+
+        Assert.assertFalse(optionalValuesBean.getNotExistingStringProperty().isPresent());
     }
 }
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/OptionalValuesBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/OptionalValuesBean.java
@@ -40,6 +40,10 @@ public class OptionalValuesBean {
     @ConfigProperty(name="my.notexisting.property")
     private Optional<Integer> notexistingProperty;
 
+    @Inject
+    @ConfigProperty(name="my.notexisting.string.property")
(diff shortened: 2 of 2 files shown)


jmesnil:
The `CdiOptionalInjectionTest` test was checking the injection for non existing property with a int.

In our implementation, that was passing because the conversion from the `@ConfigProperty.defaultValue` String to int fails and we ended up with the expected `Optional.empty`.

However if the property is an `Optional<String>` our implementation was using the  `@ConfigProperty.defaultValue` as the property value instead of returning an `Optional.empty`.

I fixed our implementation in https://github.com/wildfly-extras/wildfly-microprofile-config/pull/44 but adding this assertion to the TCK will ensure that no failing conversion would result in an injected false-positive `Optional.empty` for String.

tck/src/main/java/org/eclipse/microprofile/config/tck/CdiOptionalInjectionTest.java
```diff
@@ -66,5 +66,7 @@ public void testOptionalInjection() {
 
         Assert.assertTrue(optionalValuesBean.getStringValue().isPresent());
         Assert.assertEquals(optionalValuesBean.getStringValue().get(), "hello");
+
```
Emily-Jiang: You need to assert notExistingProperty as well.
Assert.assertFalse(optionalValuesBean.getNotExistingProperty().isPresent());

jmesnil: The test already asserts that on line 65

jmesnil: the test already asserts that on line 65

Emily-Jiang: ah. I see. I missed the existing line. Your test is to test the scenario where no converter is involved.

Issue #325, opened 2018-03-07
Remove some built in converter and swap the order of implict converter


Emily-Jiang:
Since we have the notion of the implicit converter, it takes all classes with the following criteria:

<p>If no explicit Converter and no built-in Converter could be found for a certain type,

-  The target type {@code T} has a public Constructor with a String parameter, or
-  the target type {@code T} has a {@code public static T valueOf(String)} method, or
-  the target type {@code T} has a {@code public static T parse(CharSequence)} method

With this, all built in converters are not required to be listed as they either have valueOf or a constructor taking a String parameter. 

However, to improve the performance  (as valueOf has a cache facility while the constructor does not have) and take into the new JVM class method naming scheme, I suggest we update the search sequence for the implicit converter.
-  the target type {@code T} has a {@code public static T of(String)} method
-  the target type {@code T} has a {@code public static T valueOf(String)} method, or
- The target type {@code T} has a public Constructor with a String parameter, or
-  the target type {@code T} has a {@code public static T parse(CharSequence)} method

We can then remove the section of built-in converters.

tevans78:
We can't remove all of the built-in converters;

`Boolean(String)` only accepts `true`; not `yes`, `y`, `on` or `1`

The `java.time.*` classes throw `DateTimeParseException` from their `parse(CharSequence)` methods; according to the current spec, we need to catch and rethrow as `IllegalArgumentException`

`URL(String)` throws `MalformedURLException`

The Class converter would also have to remain as it is a very special case, requiring a ClassLoader as well.

PR #340, opened 2018-04-10
#325 update spec to be insync with javadoc

Emily-Jiang:

### DIFF
--- a/spec/src/main/asciidoc/converters.asciidoc
+++ b/spec/src/main/asciidoc/converters.asciidoc
@@ -36,16 +36,7 @@ The following `Converter` s are provided by Microprofile-Config by default:
 * `long` and `Long`
 * `float` and `Float` , a dot '.' is used to separate the fractional digits
 * `double` and `Double` , a dot '.' is used to separate the fractional digits
-* `Duration`
-* `LocalTime`
-* `LocalDate`
-* `LocalDateTime`
-* `OffsetDateTime`
-* `OffsetTime`
-* `Instant`
-* `URL`
 * `Class` based on the result of `Class.forName`
-* `URI`
 
 All built-in `Converter` have the `@Priority` of `1`.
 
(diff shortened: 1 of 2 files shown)


Emily-Jiang:
@jmesnil please review. thanks

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -96,7 +87,8 @@ e.g. With this config `myPets=dog,cat,dog\\,cat`, the values as an array will be
 
 If no built-in nor custom `Converter` for a requested Type `T`, an implicit Converter is automatically provided if the following conditions are met:
 
-* The target type `T` has a public Constructor with a String parameter, or
-* the target type `T` has a `public static T valueOf(String)` method, or
-* the target type `T` has a `public static T parse(CharSequence)` method
+* The target type {@code T} has a {@code public static T of(String)} method, or
```
johnament: Is it worth pointing out that implementations can provide additional support (e.g. `URI.create`) that are functionally equivalent, or perhaps add new features on top of the required?

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -36,16 +36,7 @@ The following `Converter` s are provided by Microprofile-Config by default:
 * `long` and `Long`
 * `float` and `Float` , a dot '.' is used to separate the fractional digits
 * `double` and `Double` , a dot '.' is used to separate the fractional digits
-* `Duration`
```
johnament: I don't see why these are removed, but `Long`, and other `Number` impls remain.

spec/src/main/asciidoc/converters.asciidoc
Emily-Jiang: For what purpose? The change is making sure the provided converters are exactly same among implementations so that the app is portable among app servers.

spec/src/main/asciidoc/converters.asciidoc
Emily-Jiang: I left them as we need to keep the primitive. I can remove the boxed type. However, I think it is better to keep them because they are kind of special classes.

PR #341, opened 2018-04-12
Use a doPriv to get the parent classloader

tevans78:

Java 2 permissions are required to get the parent classloader.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
@@ -137,7 +137,13 @@ public abstract class ConfigProviderResolver {
         }
 
         // start from the root CL and go back down to the TCCL
-        ConfigProviderResolver instance = loadSpi(cl.getParent());
+        ClassLoader parentcl = AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
+            @Override
+            public ClassLoader run() {
+                return cl.getParent();
+            }
+        });
+        ConfigProviderResolver instance = loadSpi(parentcl);
 
         if (instance == null) {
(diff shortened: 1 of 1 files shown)


Emily-Jiang:
@tevans78 can you please open an issue to get this discussed before the PR?

tevans78:
fixes #343

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
```diff
@@ -137,7 +137,13 @@ private static ConfigProviderResolver loadSpi(ClassLoader cl) {
...
 
         // start from the root CL and go back down to the TCCL
-        ConfigProviderResolver instance = loadSpi(cl.getParent());
+        ClassLoader parentcl = AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
+            @Override
+            public ClassLoader run() {
+                return cl.getParent();
+            }
+        });
+        ConfigProviderResolver instance = loadSpi(parentcl);
```
Emily-Jiang: Can you put this line inside doPrivileged call? Otherwise, parentcl might be null.

tevans78: parentcl is allowed to be null (it will be when you reach the root) and there is a null check at the top of loadSpi ... functionally that hasn't changed by adding the doPriv

Emily-Jiang: This is not expected behavior though. What we want is that instance should be set to null if something is wrong in the run method though.

Emily-Jiang: double checked the api. The condition I worried about should not occur silently.

Issue #320, opened 2018-02-21
dynamic handling of configured values


struberg:
As discussed in the last MP steering hangout we gonna apply interesting stuff from ConfigJSR also to MP-config. The intention is to have people being able to play with it and give valuable feedback.

This ticket handles all the dynamic handling

https://github.com/eclipse/ConfigJSR/issues/9

PRs 
for 'push'
https://github.com/eclipse/ConfigJSR/pull/40
https://github.com/eclipse/ConfigJSR/pull/46 (basically the same while independently created by Tomas Langr)

for 'pull'
https://github.com/eclipse/ConfigJSR/pull/41

Gonna move them over and apply to master as discussed in our meeting

Emily-Jiang:
@struberg can you stop commit your PR directly without being reviewed?

struberg:
Emily, this was _explicitly_ acked in the mp meetings!

Emily-Jiang:
@struberg I bet no one asked you to directly push your changes without being reviewed. Besides, you went ahead to disable the bnd plugin without asking why. This must be stopped in the future. I will discuss in the mailing list. We need to add a guard to the repos in the future.

Emily-Jiang:
@kwsutter for your info. We need to find out from Eclipse how to stop a PR to be merged without being reviewed.

OndrejM:
I agree with @Emily-Jiang - committing non-trivial things directly should be avoided. Instead, all should go through a PR. It's always possible to create a PR and merge it right away, but only if the change is trivial or nobody commented on the PR for a while.

OndrejM:
I suggest reverting the commit and any other direct commits that bypassed a PR and raising a proper PR and allow a short discussion before it's merged.

struberg:
calm down folks. That has been discussed in the MicroProfile steering meeting a month ago. I was asked twice to push the changes from the ConfigJSR also over to mp-config.

Emily-Jiang:
@struberg I think you were asked to move changes. You can move the changes but it needs to be reviewed. However this changes are prematured though. After yesterday's meeting, I think we do need to revert it after all.

Emily-Jiang:
We need to revisit this and rework the current commit.

smillidge:
Where is the spec for this? Is this still scheduled for Config 1.3 or will this be pushed back as there is no spec or TCK to look at.

struberg:
[...]

PR #342, opened 2018-04-12
#320 undo the commits done previously as more discussions are needed …

Emily-Jiang:
…after the current release


Emily-Jiang:
@jmesnil @OndrejM undid the commits done in this #320

OndrejM:
I'm OK with reverting as not all aspects are clear and probably notifications should be sent via some messaging mechanism.

The only messaging/eventing mechanism we now have in Microprofile is CDI events - so what about using them for now just to specify push config notifications for CDI containers and postpone support in Java SE until we have an SE-based messaging mechanism?

rmannibucau:
Can this be reapplied?

There is no way to use CDI events since it is broken during the startup so using CDI events would be broken by default so the proposal was more acceptable. Sugar can be added on top of it later (with cdi events if you really want). It is at least aligned with the fact the config has its own contextual storage and doesn't reuse CDI one.

Emily-Jiang:
@rmannibucau more discussion is needed. The previous approach was not mature enough anyway. We left this issue out of the forthcoming release and discuss more for the next release.

rmannibucau:
Well Mark implemented it in deltaspike in a very elegant/efficient way so guess it can just be imported in mp config but currently there is no real way to be dynamic with mp config which is quite an issue and kind of breaks the configsource contract :(.

Should probably be solved quickly and for next release.

PR #353, opened 2018-04-20
Clarification and test for using converters and injected config

### LINKED ISSUE #348: Clarify in Javadoc and spec that the default config includes discovered converters
As pointed out in https://groups.google.com/forum/#!topic/microprofile/3kXLoA1KVz8, I didn't find any hint in Javadoc or Spec text that injected `Config` and `ConfigProvider.getConfig()` should include discovered converters.

This is probably covered by the TCK but it makes it easier for implementers to clarify in the text.

OndrejM:
Registered converters are applied in getValue() methods. Default config and injected config are equal (plus test). Default config registers discovered converters.

Also renamed test methods from using underscores to camelCase to align to other test code.

Fixes #348.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -47,8 +47,8 @@ import org.eclipse.microprofile.config.spi.ConfigSource;
  * The config objects produced via the injection model <pre>@Inject Config</pre> are guaranteed to be serializable, while
  * the programmatically created ones are not required to be serializable.
  * <p>
- * If one or more converters are registered for a class of a requested value then one of the registered converters
- * is used to convert the string value retrieved from config sources. 
+ * If one or more converters are registered for a class of a requested value then one of the registered converters 
+ * which has the highest priority is used to convert the string value retrieved from config sources. The highest priority means the highest priority number.
  * For more information about converters, see {@link org.eclipse.microprofile.config.spi.Converter}
  *
  * <h3>Usage</h3>
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -90,8 +90,6 @@ public interface Config {
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
      * 
-     * Registered converters are applied if they match the propertyType class
-     *
      * If this method gets used very often then consider to locally store the configured value.
      *
      * @param <T>
@@ -110,8 +108,6 @@ public interface Config {
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
      *     
-     * Registered converters are applied if they match the propertyType class
-     * 
      * If this method is used very often then consider to locally store the configured value.
      *
      * @param <T>
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -48,7 +48,8 @@ import org.eclipse.microprofile.config.spi.ConfigSource;
  * the programmatically created ones are not required to be serializable.
  * <p>
  * If one or more converters are registered for a class of a requested value then one of the registered converters 
- * which has the highest priority is used to convert the string value retrieved from config sources. The highest priority means the highest priority number.
+ * which has the highest priority is used to convert the string value retrieved from config sources. 
+ * The highest priority means the highest priority number.
  * For more information about converters, see {@link org.eclipse.microprofile.config.spi.Converter}
  *
  * <h3>Usage</h3>
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -46,6 +46,10 @@ import org.eclipse.microprofile.config.spi.ConfigSource;
  * <p>
  * The config objects produced via the injection model <pre>@Inject Config</pre> are guaranteed to be serializable, while
  * the programmatically created ones are not required to be serializable.
+ * <p>
+ * If one or more converters are registered for a class of a requested value then one of the registered converters
+ * is used to convert the string value retrieved from config sources. 
+ * For more information about converters, see {@link org.eclipse.microprofile.config.spi.Converter}
  *
  * <h3>Usage</h3>
  *
@@ -85,6 +89,8 @@ public interface Config {
     /**
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
+     * 
+     * Registered converters are applied if they match the propertyType class
      *
      * If this method gets used very often then consider to locally store the configured value.
      *
@@ -104,6 +110,8 @@ public interface Config {
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
      *     
+     * Registered converters are applied if they match the propertyType class
+     * 
      * If this method is used very often then consider to locally store the configured value.
      *
      * @param <T>
--- a/spec/src/main/asciidoc/configprovider.asciidoc
+++ b/spec/src/main/asciidoc/configprovider.asciidoc
@@ -26,23 +26,27 @@ For using Microprofile-Config in a programmatic way the `ConfigProvider` class i
 It allows access to different configurations (represented by a `Config` instance) based on the application in which it is used.
 The `ConfigProvider` internally delegates through to the `ConfigProviderResolver` which contains more low-level functionality.
 
-There are 4 different ways to create an `Config` instance:
+There are 4 different ways to create a `Config` instance:
 
-* In CDI managed components a user can use `@Inject` to access the current application configuration.
+* In CDI managed components, a user can use `@Inject` to access the current application configuration.
   The default and the auto discovered <<configsource,ConfigSources>> will be gathered to form a configuration.
+  The default and the auto discovered <<converters,Converters>> will be gathered to form a configuration.
+  Injected instance of `Config` should behave the same as the one retrieved by `ConfigProvider.getConfig()`. Injected values should be the same as if retrieved from the injected `Config` instance.
 
 * A factory method `ConfigProvider#getConfig()` to create a `Config` object based on automatically picked up `ConfigSources`
   of the Application identified by the current Thread Context ClassLoader classpath.
+  The default and the auto discovered <<converters,Converters>> will be gathered to form a configuration.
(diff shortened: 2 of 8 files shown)


api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -85,6 +89,8 @@
     /**
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
+     * 
+     * Registered converters are applied if they match the propertyType class
```
ljnelson: If I have a registered converter for type `List<String>` and the injection point is of type `Collection<? extends String>`, is this considered a "match" in the language of this specification?

spec/src/main/asciidoc/configprovider.asciidoc
```diff
@@ -26,23 +26,27 @@ For using Microprofile-Config in a programmatic way the `ConfigProvider` class i
 It allows access to different configurations (represented by a `Config` instance) based on the application in which it is used.
 The `ConfigProvider` internally delegates through to the `ConfigProviderResolver` which contains more low-level functionality.
 
-There are 4 different ways to create an `Config` instance:
+There are 4 different ways to create a `Config` instance:
 
-* In CDI managed components a user can use `@Inject` to access the current application configuration.
+* In CDI managed components, a user can use `@Inject` to access the current application configuration.
   The default and the auto discovered <<configsource,ConfigSources>> will be gathered to form a configuration.
+  The default and the auto discovered <<converters,Converters>> will be gathered to form a configuration.
```
ljnelson: Are you sure you meant to use `Converters` here and not `ConfigSources`?  I fail to see how a collection of `Converter` instances constitutes a configuration.

OndrejM: OK, the wording for converters should be different. The point is that default and auto discovered converters are used.

api/src/main/java/org/eclipse/microprofile/config/Config.java
OndrejM: The only thing that the spec says about this is "The Type parameter of the interface is the target
type the String is converted to". I admit this isn't clear enough because if the target type is List at can also be assigned to a Collection. That would work for programmatic API, but for injection it's way too complicated to support because a producer would have to be created for all supertypes and interfaces.

We'll clarify that for injection, only the exact target type can be injected. Injection of supertypes like `Collection` isn't portable.

ljnelson: OK.

In an alternate universe 😄 you could say this in the general specification:

* A `Converter<T>` must be registered for all type conversion.
* `T` must be the _maximally concrete type_ for the conversion.  _Maximally concrete_ means something like `HashSet<String>`, not `Collection<String>`; i.e. the "leaf" of the type hierarchy.  If a `T` were, say, `Collection<T>`, then only `Collection<T>` and `Collection<? extends T>` injection points would work.  If `T` were, by contract, `HashSet<String>`, then all of its transitive type closure injection points would work.

Then to implement CDI injection, you don't scan for injection points, you scan maximally concrete conversion types and blindly register a producer for each of them.  Problem solved for all possible injection or programmatic lookup.

But since that's not how things are written, and we have to scan for injection points, I think your clarifications are fine.  As I noted in the Google Group, there are enough limitations with the specification and CDI integration that I'd err on the side of talking explicitly about what is supported, and say that if there's anything else someone's thinking about, point out that it is undefined.

spec/src/main/asciidoc/configprovider.asciidoc
ljnelson: OK.  Strictly speaking, this isn't true either, is it?  I'm not trying to be difficult: I'm just pointing out that strictly speaking what is required is that conversion happens via `Config#getValue(nameOfInjectionPoint, SomeType.class)`, right?  If so, note that `Converter`s are not in that method's javadoc, or in the specification covering that method's behavior.  (I understand that the specification _implies and hints_ that they _should_ be used _somewhere_ in any given implementation of `Config#getValue(String, Class)`, but it doesn't say that.)  Again, not trying to be annoying! it's just that if this is a spec from which I'm supposed to implement things, I have to connect an awful lot of dots with an awful lot of assumptions right now.

OndrejM: I added this to the javadoc of `Config.getValue`:

> Registered converters are applied if they match the propertyType class

I think that it clarifies that converters should be applied. But only those which are registered with the config. For the default config used in injection, default and autodiscovered converters are registered, which is explained here in the spec and also in the javadoc of `ConfigProvider.getConfig()` and `ConfigProperty`.

Is something missing or unclear even after I added these clarifications?

api/src/main/java/org/eclipse/microprofile/config/Config.java
OndrejM: This would be more of a recommendation than a specification. I think that implementations don't have to scan injection points. They can also choose to scan converters and classpath for all classes with implicit conversion. It's up to the implementation.

However, it makes sense to add hints into the spec for implementers even if they are not necessary to specify behavior. I think that JSRs even have a separate document for implementers for that purpose.

ljnelson: How would an implementation scan the classpath for "all classes with implicit conversion"?  Since it cannot, I don't think this is a case where the implementation could use that strategy.  That cuts off a whole tree of possible behavior, so it may make sense to enumerate what is explicitly defined behavior and what is not.

Anyway, some clarifications are better than none so I wouldn't hold this PR up for all possible fixes.

spec/src/main/asciidoc/configprovider.asciidoc
OndrejM: I use those words intergangably... Strictly speaking, MicroProfile doesn't certify any implementation, it's rather a recommendation and a set a compatibility tests. There's nothing like impls must do something, or otherwise... But if the word "must" makes things clearer, I can change it

api/src/main/java/org/eclipse/microprofile/config/Config.java
OndrejM: Application servers can easily scan all the classpath because it's limited to the deployed WAR or EAR. And, in fact, most of them already do. It's definitely up to the implementation which approach they choose.

spec/src/main/asciidoc/configprovider.asciidoc
ljnelson: I mean, I don't really care.  :-)  It's a question of whether I'm in violation of something if I implement `Config#getValue(String, Class)` without using `Converter`s behind the scenes.  If this is legal, then it's a "should"; if it's not legal (i.e. if there's a TCK test for it or something), then the specification MUST :-) specify it with "must".

api/src/main/java/org/eclipse/microprofile/config/Config.java
ljnelson: OK.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -46,6 +46,10 @@
  * <p>
  * The config objects produced via the injection model <pre>@Inject Config</pre> are guaranteed to be serializable, while
  * the programmatically created ones are not required to be serializable.
+ * <p>
+ * If one or more converters are registered for a class of a requested value then one of the registered converters
```
Emily-Jiang: It should say the Converter with the high priority will be used to convert the string.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -104,6 +110,8 @@
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
      *     
+     * Registered converters are applied if they match the propertyType class
```
Emily-Jiang: What do you mean by saying 'Registered converters'? It depends. If the registered converter is having the lower priority than the implicit converter, it will not be used.

spec/src/main/asciidoc/configprovider.asciidoc
Emily-Jiang: @ljnelson it is a recommendation on how to create a config. If you write your implementation by completely ignoring the converters by using some reflection per se but you can satisfy the getValue(String, Class) contract, it is absolutely fine. Our TCKs only verify the api behaviours not on how you want to implement it. As @OndrejM pointed out, you can ignore some recommendations implied by the spec, if you can pass the tcks. You will be fine. Therefore, there is no 'must' to force implementors to do the best of practices but the implementors must fulfil the api contracts and pass tcks.

ljnelson: Thanks, @Emily-Jiang; now I'm afraid I'm a little more confused.  Is the TCK the source of truth?  For example, the specification does not say anything about whether converters must be used or not.  It describes their behavior assuming they will be used, but there's no requirement that they be used.  But maybe the TCK _does_ enforce their usage (I don't know yet; haven't run it).  Are you saying that in a scenario like this the TCK would be the source of truth: regardless of what the spec says, if my code passes the TCK I'm OK?

api/src/main/java/org/eclipse/microprofile/config/Config.java
ljnelson: I had a question here too.  In English, one's "highest priority" is also one's "first" priority (i.e. a "high priority" has a low number, and a "low priority" has a high number).  I assume "high priority" here means simply "greatest `value()` element of the `@Priority` annotation?

Emily-Jiang: The highest priority is the one with the high number.

tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPlainInjectionTest.java
```diff
@@ -85,31 +86,58 @@ public void can_inject_simple_values_when_defined() {
         assertThat(bean.doublePropertyWithDefaultValue, is(closeTo(3.1415, 0.1)));
     }
 
+    /*
+     * Refers to chapter "Accessing or Creating a certain Configuration"
+     */
     @Test
-    public void can_inject_dynamic_values_via_CDI_provider() {
-        clear_all_property_values();
+    public void injectedValuesAreEqualToProgrammaticValues() {
```
OndrejM: should call clear properties first

OndrejM: clear should be called before all tests

OndrejM: ensure props defined should be also executed for each test

Emily-Jiang:
I'll merge this PR as per Ondro's comment on gitter.

Emily-Jiang:
closes #348

PR #359, opened 2018-05-04
[osgi] add portable java contract for osgi support

rotty3000:


api/bnd.bnd
```diff
@@ -2,9 +2,20 @@
     org.eclipse.microprofile.*
 
 Import-Package: \
-    javax.enterprise.util;version="[1.1,3)", \
+    javax.enterprise.*;-remove-attribute:=version, \
```
Emily-Jiang: I am not comfortable on removing the version all together.

Emily-Jiang:
@rotty3000 can you please open an issue to have this support discussed? If we do want this kind of change, I would like to make every spec to adopt this not just this spec.

Issue #358, opened 2018-05-02
Case sensitivity of environment variables in TCK CDIPropertyNameMatchingTest


neilgsyoung:
CDIPropertyNameMatchingTest assumes that the environment in which the TCK runs is case sensitive. In particular, the assumption is that my_string_property and MY_STRING_PROPERTY may be configured as distinct variables. This is not the case on, for example, Windows.

The test should be reworked to tolerate case insensitive environments.

jmesnil:
we can split this test in 2.

* one simple test that checks the mapping rule without any case conversion
* another test that checks with case conversion. In this 2nd case, we can them use JUnit assume to check it is not run on Windows.

neilgsyoung:
@jmesnil and @Emily-Jiang, I believe the test in its current form will run cross-platform. I'm happy to tackle @jmesnil's suggestion but how about I open a new issue and we do it the other side of RC4?

Emily-Jiang:
I double checked and found out #361 fixed this issue properly. Close this issue and move out of the milestone as it was fixed before MP Config 1.3.

PR #361, opened 2018-05-17
change env variable 

neilgsyoung:
Signed-off-by: Neil Young <neil_young@uk.ibm.com>


tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPropertyNameMatchingTest.java
```diff
@@ -43,7 +43,7 @@
  * The following environment variables must be set prior to running this test:
  * "my_int_property" with the value of "45"
  * "MY_BOOLEAN_PROPERTY" with the value of "true"
- * "my_string_property" with the value of "haha"
+ * "MY.STRING_PROPERTY" with the value of "haha"
```
jmesnil: This change is not valid. The environment variable can not contain dots.

What is the intent of your PR?

Emily-Jiang:
This closes #358

tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPropertyNameMatchingTest.java
Emily-Jiang: @jmesnil In Windows, it cannot differentiate the case. my.string.property is treated as the same as MY.STRING.PROPERTY. The test will not pass when running on Windows. Maybe we should ignore this test. Thoughts?

neilgsyoung:
I do apologise @jmesnil and @Emily-Jiang , the change worked on Windows but won't work on other platforms. I've tweaked the test, removing the offending env var (and therefore removing the testing of case sensitivity).

Emily-Jiang:
@jmesnil I am happy with the changes now. Can you double check to see whether you have other concerns? I would like to merge it asap to start a final RC.

jmesnil:
@Emily-Jiang lgtm

PR #373, opened 2018-06-23
Link to microprofile-config-api in Maven Central

arend-von-reinersdorff:
Fix link destination from microprofile-config-spec to microprofile-api.


Emily-Jiang:
thanks @arend-von-reinersdorff !

Issue #384, opened 2018-09-04
Backport ConfigAccessor and ConfigSnapshot from ConfigJSR


struberg:
ConfigJSR introduce a few advanced features for handling dynamic values. 
See the discussions over here 
https://github.com/eclipse/ConfigJSR/issues/9
and
https://github.com/eclipse/ConfigJSR/issues/8

jmesnil:
@struberg is it still relevant? ConfigAccessor is already in the code base?

PR #385, opened 2018-09-04
DON'T MERGE - Config accessor backport. refs: #384

struberg:
This is a backport from quite a few changes we did in ConfigJSR.

What's left missing right now are the TCK changes. Will push them tomorrow.


Emily-Jiang:
Thanks Mark! I'll take a look and then discuss this on next week's call! Please don't merge.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -108,11 +117,12 @@
     /**
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
-     *     
+     *
      * If this method is used very often then consider to locally store the configured value.
      *
-     * @param <T>
-     *             The property type
+     * <p>Note that no variable replacement like in {@link ConfigAccessor#evaluateVariables(boolean)} will be performed!
```
Emily-Jiang: I don't think this sentence brings much value and it is rather confusing.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -90,11 +98,12 @@
     /**
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
-     * 
+     *
      * If this method gets used very often then consider to locally store the configured value.
      *
-     * @param <T>
-     *             The property type
+     * <p>Note that no variable replacement like in {@link ConfigAccessor#evaluateVariables(boolean)} will be performed!
```
Emily-Jiang: I don't think this sentence brings much value and it is rather confusing.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -124,7 +134,60 @@
     <T> Optional<T> getOptionalValue(String propertyName, Class<T> propertyType);
 
     /**
-     * Return a collection of property names.
+     * Create a {@link ConfigAccessor} to access the underlying configuration.
+     *
+     * @param propertyName the property key
+     * @return a {@code ConfigAccessor} to access the given propertyName
+     */
+    ConfigAccessor<String> access(String propertyName);
```
Emily-Jiang: The String should be T

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -124,7 +134,60 @@
...
+     *     // and right inbetween the underlying values get changed to 'newserver' and port 8082
+     *
+     *     // get the current port for the host
+     *     ConfigAccessor&lt;Integer&gt; portCfg config.resolve("myapp.port")
+     *              .as(Integer.class)
+     *              .cacheFor(60, TimeUnit.MINUTES);
+     * </pre>
+     *
+     * In ths above code we would get the combination of {@code 'oldserver'} but with the new port {@code 8081}.
+     * And this will obviously blow up because that host+port combination doesn't exist.
```
Emily-Jiang: I would not use the word blow up. Please change to
and this will obviously not work because the host+port combination doesn't exist.

api/src/main/java/org/eclipse/microprofile/config/Config.java
struberg: gnah, resolving ditches the commtent :(

So here again: My version is correct. Default is `ConfigAccessor<String>`
If one wants another type then use `as(SomeType.class)` which will then return a `ConfigAccessor<SomeType>`

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -42,7 +43,16 @@
  * The default config sources always available by default are:
  * <ol>
  * <li>System properties (ordinal=400)</li>
- * <li>Environment properties (ordinal=300)</li>
+ * <li>Environment properties (ordinal=300)
+ *    <p>Depending on the operating system type, environment variables with '.' are not always allowed.
```
Emily-Jiang: please remove this as the change in mp config needs to be brought back to config jsr, global replacement not only .

api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java
```diff
@@ -125,4 +127,25 @@
...
+     */
+    @Nonbinding
+    boolean evaluateVariables() default true;
+
+    /**
+     * Only valid for injection of dynamically readable values, e.g. {@code Provider<String>}!
+     * @return {@code TimeUnit} for {@link #cacheFor()}
+     */
+    @Nonbinding
+    TimeUnit cacheTimeUnit() default TimeUnit.SECONDS;
```
Emily-Jiang: I thought we agreed to use ChronoUnit. MP FT uses ChronoUnit. It is better to sync up.

api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java
```diff
@@ -125,4 +127,25 @@
...
+
+    /**
+     * Only valid for injection of dynamically readable values, e.g. {@code Provider<String>}!
```
Emily-Jiang: We should add a validation rule to see if this is specified on a naked value, the DefintionException should be thrown.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
Emily-Jiang: you might have not synced up with master

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java
```diff
@@ -0,0 +1,253 @@
+/*
+ ******************************************************************************
+ * Copyright (c) 2009-2018 Contributors to the Eclipse Foundation
```
Emily-Jiang: Why does this start 2009, as the first contributor did this in 2011?

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java
```diff
@@ -0,0 +1,253 @@
...
+ *******************************************************************************/
+package org.eclipse.microprofile.config;
+
+
+import org.eclipse.microprofile.config.spi.Converter;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
```
Emily-Jiang: Can we unify this to use ChronoUnit as it is used by quite a few annotations?

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -124,7 +134,60 @@
...
+     *
+     * To consistently access n different config values we can start a {@link ConfigSnapshot} for those values.
+     *
+     * <pre>
+     *     ConfigSnapshot cfgSnap = config.createSnapshot(hostCfg, portCfg);
```
Emily-Jiang: should be snapshotFor(). Maybe we should rename the method to createSnapshot()

spec/src/main/asciidoc/configaccessor.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+The call to `config.access(..)` returns a `ConfigAccessor`.
+This is basically a builder which has methods to refine the resolution, including the following:
+
+* `as(Class<N> clazz)` -- defines the return type of the property
+* `asList()` -- Declare the ConfigAccessor to return a List of the given Type.
+* `evaluateVariables(boolean evaluateVariables)`
+* `useConverter(Converter<T> converter)` -- Defines a specific {@link Converter} to be used instead of applying the default Converter resolving logic.
+* `withLookupSuffix(String postfixName)` -- sets a parameter for the resolution of a 'postfix' for the resolution
+* `withLookupSuffix(ConfigAccessor)` -- you can also use a `ConfigAccessor` to determine the 'postfix' for the resolution
+* `withDefault(T value)` -- sets the default value, used in case the resolution returns `null`
```
Emily-Jiang: missing withStringDefault

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -37,6 +37,7 @@ The following `Converter` s are provided by Microprofile-Config by default:
 * `float` and `Float` , a dot '.' is used to separate the fractional digits
 * `double` and `Double` , a dot '.' is used to separate the fractional digits
 * `Class` based on the result of `Class.forName`
+* `URL`
```
Emily-Jiang: should not be added as it is covered under automatic converters

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -54,41 +55,68 @@ The `Config` will use the `Converter` with the highest `Priority` for each targe
 A custom `Converter` for a target type of any of the built-in Converters will overwrite the default Converter.
 
 Converters can be added to the `ConfigBuilder` programmatically via `ConfigBuilder#withConverters(Converter<?>... converters)`
-where the type of the converters can be obtained via reflection. However, this is not possible for a lambda converter. In this case, use the method `ConfigBuilder#withConverter(Class<T> type, int priority, Converter<T> converter)`.
+where the type of the converters can be obtained via reflection. However, this is not possible for a lambda converter.
+In this case, use the method `ConfigBuilder#withConverter(Class<T> type, int priority, Converter<T> converter)`.
```
Emily-Jiang: please revert this as it needs to be in a different PR.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -124,7 +134,60 @@
...
+     * {@link ConfigAccessor} which must be consistent.
+     * The returned {@link ConfigSnapshot} is an immutable object which contains all the
+     * resolved values at the time of calling this method.
+     *
+     * <p>An example would be to access some {@code 'myapp.host'} and {@code 'myapp.port'}:
+     * The underlying values are {@code 'oldserver'} and {@code '8080'}.
+     *
+     * <pre>
+     *     // get the current host value
+     *     ConfigAccessor&lt;String&gt; hostCfg config.resolve("myapp.host")
```
Emily-Jiang: this is incorrect.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -124,7 +134,60 @@
...
+     *              .cacheFor(60, TimeUnit.MINUTES);
+     *
+     *     // and right inbetween the underlying values get changed to 'newserver' and port 8082
+     *
+     *     // get the current port for the host
+     *     ConfigAccessor&lt;Integer&gt; portCfg config.resolve("myapp.port")
+     *              .as(Integer.class)
```
Emily-Jiang: do you mean = config.access

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -54,41 +55,68 @@ The `Config` will use the `Converter` with the highest `Priority` for each targe
...
+
+!TODO rule option A:
+* The target type {@code T} has a {@code public static T of(String)} method, or
+* The target type {@code T} has a {@code public static T valueOf(String)} method, or
+* The target type {@code T} has a public Constructor with a String parameter, or
+* The target type {@code T} has a {@code public static T parse(CharSequence)} method
+
+!TODO rule option B:
+* The target type `T` has a Constructor with a String parameter, or
+* the target type `T` has a Constructor with a CharSequence parameter, or
```
Emily-Jiang: This should be done under a different issue and PR.

api/src/main/java/org/eclipse/microprofile/config/ConfigSnapshot.java
```diff
@@ -0,0 +1,42 @@
...
+ * @see Config#snapshotFor(ConfigAccessor...)
+ * @see ConfigAccessor#getValue(ConfigSnapshot)
+ *
+ * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
+ * @author <a href="mailto:manfred.huber@downdrown.at">Manfred Huber</a>
+ * @author <a href="mailto:elexx@apache.org">Alex Falb</a>
+ * @author <a href="mailto:gpetracek@apache.org">Gerhard Petracek</a>
+ * @author <a href="mailto:rmannibucau@apache.org">Romain Manni-Bucau</a>
+ */
+public interface ConfigSnapshot {
```
tevans78: I think that `ConfigSnapshot` is the wrong name since it implies a instant in time sampling of the data. The Google definition of "Snapshot" (in computing) is “a record of the contents of a storage location or data file at a given time."

As I understand the intended concept, a better name might be `ConfigGrouping`?

api/src/main/java/org/eclipse/microprofile/config/package-info.java
```diff
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
```
tevans78: The package version lower down in this file should be bumped

tevans78: Should the package versions for `org.eclipse.microprofile.config.inject` and `org.eclipse.microprofile.config.spi` also be increased? @Emily-Jiang ?

tevans78: Probably this...
`org.eclipse.microprofile.config` = 1.1
`org.eclipse.microprofile.config.inject` = 1.1
`org.eclipse.microprofile.config.spi` = 1.4

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java
```diff
@@ -0,0 +1,253 @@
...
+     * @return This builder as a typed ConfigAccessor
+     */
+    ConfigAccessor<T> useConverter(Converter<T> converter);
+
+    /**
+     * Sets the default value to use in case the resolution returns null.
+     * @param value the default value
+     * @return This builder
+     */
+    ConfigAccessor<T> withDefault(T value);
```
Emily-Jiang: What is the use case here? The Javadoc is unclear. What resolution do you mean? Why not just rely on getOptionalValue to set the default?

struberg: because with getOptionalValue you have to do the `orElse` whereever you use that value. The idea behind `withDefault` is to have this at exactly a single place for each ConfigAccessor.

Issue #386, opened 2018-09-08
Add built-in converters for byte, short and char


gunnarmorling:
These should be added for the sake of consistently supporting all Java primitive types. I'll file a PR shortly.

As discussed on the ML: https://groups.google.com/forum/#!topic/microprofile/YxUp6Z87yz0

PR is here: https://github.com/eclipse/microprofile-config/pull/387

PR #387, opened 2018-09-10
#386 Adding built-in converters for byte, short and char

gunnarmorling:
As discussed here's a PR for completing the list of built-in converters. Are there other changes needed than that?

### DIFF
--- a/spec/src/main/asciidoc/converters.asciidoc
+++ b/spec/src/main/asciidoc/converters.asciidoc
@@ -31,16 +31,16 @@ This happens by providing ``Converter``s in the `Config`.
 
 The following ``Converter``s are provided by MicroProfile-Config by default:
 
-* `boolean` and `Boolean` , values for `true` (case insensitive) "true", "1", "YES", "Y" "ON".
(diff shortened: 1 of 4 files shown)


gunnarmorling:
Hey @Emily-Jiang, thanks for reviewing.

>  Can you sign off the PR so that you PR can pass ip-validation

Done.

> Can you add tests to CDIPlainInjectionTest to test the injection of byte, Byte, char, Charater working?

Reworked the last commit for that. Naturally, I haven't actually executed the TCK with these changes yet, as there's no compatible implementation. I'm not sure how you're dealing with this in general, but in the worst case we could adjust later on, should there be a bug in these new tests.

gunnarmorling:
Hey @Emily-Jiang, anything else I should do to get it merged? Thx!

struberg:
Reviewed and applied. Gunnar, you might also add yourself in the NOTICE file if you like.

gunnarmorling:
Thanks! Just checked, I'm listed in the NOTICE file already :)

Emily-Jiang:
@gunnarmorling thanks for your contribution! We will run the tcks against the impls in due course and verify whether they are correct. If not, further issue will be created to correct the tcks. I think some impls have this support already.

PR #391, opened 2018-12-05
Revert "Fixes typos"

Emily-Jiang:
This reverts commit 4fcbbafcdf7c6eaf32220bd895a2fd7c599c8057.


api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java
```diff
@@ -101,7 +101,7 @@
 @Retention(RUNTIME)
 @Target({METHOD, FIELD, PARAMETER, TYPE})
 public @interface ConfigProperty {
-    String UNCONFIGURED_VALUE="org.eclipse.microprofile.config.configproperty.unconfiguredvalue";
+    String UNCONFIGURED_VALUE="org.eclipse.microprofile.config.configproperty.unconfigureddvalue";
```
tevans78: I'd prefer there to be a comment in the code which explains why this is misspelt ... but that is a minor point and I will still approve this PR.

PR #388, opened 2018-09-23
#365 allow null to be injected

Emily-Jiang:
Signed-off-by: Emily Jiang <emijiang@uk.ibm.com>

PR #392, opened 2018-12-09
Refactor on #388 PR

Emily-Jiang:


tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigAccessorTest.java
```diff
@@ -0,0 +1,259 @@
...
+    public static WebArchive deploy() {
+        JavaArchive testJar = ShrinkWrap
+            .create(JavaArchive.class, "configValueTest.jar")
+            .addPackage(AbstractTest.class.getPackage())
+            .addClass(ConfigAccessorTest.class)
+            .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+            .addAsServiceProvider(ConfigSource.class, ConfigurableConfigSource.class)
+            .as(JavaArchive.class);
+
+        AbstractTest.addFile(testJar, "META-INF/javaconfig.properties");
```
jmesnil: @Emily-Jiang please change it to `META-INF/microprofile-config.properties` to make the test deploy properly

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigAccessorTest.java
```diff
@@ -0,0 +1,259 @@
...
+     */
+    @Test
+    public void testVariableReplacement() {
+        Assert.assertEquals(config.access("tck.config.variable.firstEndpoint", String.class).build().getValue(),
+                "http://some.host.name/endpointOne");
+
+        Assert.assertEquals(config.access("tck.config.variable.secondEndpoint", String.class).build().getValue(),
+                "http://some.host.name/endpointTwo");
+
+        // variables in Config.getValue and getOptionalValue do not get evaluated otoh
```
jmesnil: why are they not evaluated? for backwards compatibility?

Does it support nested variable replacement? e.g.  `${foo.${bar}}`
Does it support default value? e.g. `${foo:123}` -> returns `123` if the prop `foo` is not found
Does it support escaping? (so that the string `${` can appear without being an expression

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigAccessorTest.java
Emily-Jiang: fixed. thanks @jmesnil

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigAccessorTest.java
Emily-Jiang: They are not evaluated for backward compatibility. The variables were evaluated only via configaccess object.
${foo.${bar}} should be evaluated recursively.
At the moment, it does not support default value if a particular variable does not exist. I suggest For the case of ${foo}, if `foo` does not exist, we could either fail with exception or not to evaluate the variable.
Good point of the escaping aspect, we have not talked about this. I think it is a good idea to introduce escaping... how about \${... escape a escape character will be \\${-> \${, ${ will be still escaped

dmlloyd: Hi, I'm an outsider here but just wanted to comment that it would be nice if there was a way to be able to use variable expansion directly on the `Config` interface.  Couldn't this be something set on `org.eclipse.microprofile.config.spi.ConfigBuilder`?

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessorBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+     * Note that that the cache will get flushed if a {@code ConfigSource} notifies
+     * the underlying {@link Config} about a value change.
+     * This is done by invoking the callback provided to the {@code ConfigSource} via
+     * {@link org.eclipse.microprofile.config.spi.ConfigSource#onAttributeChange(java.util.function.Consumer)}.
+     *
+     * @param value the amount of the TimeUnit to wait
+     * @param timeUnit the ChronoUnit for the value
+     * @return This builder
+     */
+    ConfigAccessorBuilder<T> cacheFor(long value, ChronoUnit timeUnit);
```
tevans78: Should use Duration rather than a long and a ChronoUnit

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Returns the property name key given in {@link Config#access(String, Class)}.
+     * @return the original property name
+     */
+    String getPropertyName();
+
+    /**
+     * Returns the actual key which led to successful resolution and corresponds to the resolved value.
+     * This is useful when {@link ConfigAccessorBuilder#addLookupSuffix(String)} is used.
```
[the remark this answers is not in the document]
jmesnil: For reference, David is talking about https://www.playframework.com/documentation/1.5.x/ids

However, it seems that Play 2 has dropped out this feature and uses instead a simpler `devSettings`:  
https://www.playframework.com/documentation/2.6.x/ConfigFile#Extra-devSettings

api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java
```diff
@@ -135,4 +137,25 @@
      */
     @Nonbinding
     String defaultValue() default UNCONFIGURED_VALUE;
+
```
tevans78: I don't know why this diff isn't flagging it up as a change but there should be a typo in the value of UNCONFIGURED_VALUE, an extra 'd'...

```
String UNCONFIGURED_VALUE="org.eclipse.microprofile.config.configproperty.unconfigureddvalue";
```

This was previously removed but then re-instated by PR #391

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java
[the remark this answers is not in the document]
Emily-Jiang: I'll remove this method for now and raise a separate issue for further discussion.

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessorBuilder.java
Emily-Jiang: will update to Duration.

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessorBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+     *
+     * The algorithm to use in {@link ConfigAccessor#getValue()} is a binary count down.
+     * Every parameter is either available (1) or not (0).
+     * Having 3 parameters, we start with binary {@code 111} and count down to zero.
+     * The first combination which resolves to a result is being treated as result.
+     *
+     * @param suffixValue fixed String to be used as suffix
+     * @return This builder
+     */
+    ConfigAccessorBuilder<T> addLookupSuffix(String suffixValue);
```
Emily-Jiang: Will remove this method for now.

api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java
Emily-Jiang: revert back to the master branch.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -146,4 +164,57 @@ default int getOrdinal() {
...
+    /**
+     * The callback should get invoked if an attribute change got detected inside the ConfigSource.
+     *
+     * @param callback will be set by the {@link org.eclipse.microprofile.config.Config} after this
+     *                 {@code ConfigSource} got created and before any configured values
+     *                 get served.
+     * @return ChangeSupport informing the {@link org.eclipse.microprofile.config.Config} implementation about support for changes by this source
+     * @see ChangeSupport
+     */
+    default ChangeSupport onAttributeChange(Consumer<Set<String>> callback) {
```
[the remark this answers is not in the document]
Emily-Jiang: RegisterChangeListener
DeRegisterChangeListener

Emily-Jiang:
@jmesnil @OndroMih @tevans78 I have done the changes based on our discussion last week. please review the changes again. I would like to merge them before tomorrow's meeting, so that we can discuss how best to do prefix or suffix on property names.

PR #398, opened 2019-01-18
Refactor #384

jmesnil:
Update the onAttributeChange method to return a ChangeSupport class that
can be used to handle release of resources (in the close method).

When a Config is released, it should call ChangeSupport.close to ensure
that the ConfigSource has a chance to properly release any resources
held by the callbacks.

* Add TCK test ConfigAccessorTest#testOnAttributeChange

Signed-off-by: Jeff Mesnil <jmesnil@gmail.com>

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
@@ -26,11 +26,13 @@
  *      Extracted the Config part out of DeltaSpike and proposed as Microprofile-Config cf41cf130bcaf5447ff8
  *   2016-11-14 - Emily Jiang / IBM Corp
  *      Methods renamed, JavaDoc and cleanup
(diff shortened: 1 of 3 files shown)


jmesnil:
@Emily-Jiang @tevans78 @OndroMih following today's call, this is an alternative to handle attribute change.

The main change is that now the method return an object which can be `closed` to let the ConfigSource release any resources it helds.

I added a dummy prototype to check that the added test passed as expected in https://github.com/jmesnil/smallrye-config/commit/76fddcdef20df61db9640b523646680ad4307175

OndroMih:
Why did this go to eclipse:struberg-ConfigAccessor and not to master? @jmesnil, can we merge it to master?

OndroMih:
I assume it was meant to update the PR #392, which we merged meanwhile. This should be merged to master too, can you raise a new PR against master, @jmesnil ?

PR #401, opened 2019-02-08
Fix a few typos in the Converters section

gsmet:
Nothing fancy, noted while reading the Converters chapter.


genie-microprofile:

PR #405, opened 2019-02-14
Support variable expansion

jmesnil:
* Spec
  * Update specification with Variable Expansion chapter
  * Add a note about backwards compatibilty
* API
  * rename evaluateVariables() to expandVariables()
* TCK
  * Add VariableExpansionTest to the TCK to validate variable expansion

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/ConfigAccessorBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/ConfigAccessorBuilder.java
@@ -98,7 +98,7 @@ public interface ConfigAccessorBuilder<T> {
      * {@code "mycompany.some.url"} would be:
      * {@code "http://localhost:8081/some/path"}
      *
-     * <p><b>ATTENTION:</b> This defaults to {@code true}! That means variable replacement is enabled by default!</p>
+     * <strong>ATTENTION: This defaults to {@code true}! That means variable evaluation is enabled by default!</strong>
      *
      * @param evaluateVariables whether to evaluate variables in values or not
      *
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
@@ -95,6 +95,19 @@ public interface ConfigBuilder {
      */
     ConfigBuilder withConverters(Converter<?>... converters);
 
+    /**
+     * Determines whether variable evaluation is enabled.
+     *
+     * This enables variable evaluation globally for any configuration values retrieved using the
+     * {@link Config#getValue(String, Class)} and {@link Config#getOptionalValue(String, Class)} methods
+     * on `Config` built by this builder.
+     *
+     * This setting has no impact on {@link org.eclipse.microprofile.config.ConfigAccessorBuilder#evaluateVariables(boolean)}.
+     *
+     * @param evaluateVariables {@code true} to evaluate variables
+     * @return the ConfigBuilder
+     */
+    ConfigBuilder evaluateVariables(boolean evaluateVariables);
 
     /**
      * Add the specified {@link Converter} for the given type.
--- a/spec/src/main/asciidoc/variable-evaluation.asciidoc
+++ b/spec/src/main/asciidoc/variable-evaluation.asciidoc
@@ -1,0 +1,123 @@
+//
+// Copyright (c) 2016-2019 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+[[variable-evaluation]]
+== Variable Evaluation
+
+The value of a configuration property can contains a variable corresponding to another configuration property.
+This variable is _evaluated_ to be replaced by its own value when the original property is returned.
+
+For example, let's define a `server.url` property:
+
+[source]
+----
+server.url=http://${server.host}:{server.port}/endpoint
+----
+
+When the Config API returns the value of `server.url`, it evaluates any variable (such as `server.host` and `server.port`) to replace
+them with their value.
+
+If these properties are also defined:
+
+[source]
+----
+server.host=example.org
+server.port=8080
+----
+
+The `server.url` value returned by the Config API is `http://example.org:8080/endpoint`.
+
+[NOTE]
+====
+Backwards Compatibility
+
+Variable evaluation is not backwards compatible.
+Previous version of MicroProfile Config would not evalute the variables and would return the value `http://${server.host}:{server.port}/endpoint`.
+
+Implementation of MicroProfile Config MUST provide a way to disable variable evaluation to provide backwards compatibility.
+
+The property `mp.config.evaluateVariables` can be configured as a boolean in one of the 3 default `ConfigSource` (environment variables, system properties,
+`META-INF/microprofile-config.properties`). Its value will be used by the MicroProfile Config implementation to _globally_ enable or disable variable evaluation.
+In the absence of this property, variable evaluation is enabled.
+====
+
+---
+
+
+=== Variable Syntax
+
+The syntax to use a variable is `${key}` for a _required_ config property and `${<key>:<default_value>}` for
+an _optional_ config property.
+
+If the variable is required , by using `${key}`, a config property with the name `key` MUST be present in the Config to be evaluated, otherwise, the
+Config API will throw a `NoSuchElementException` (or an `Optional.empty()` value).
+
+For example, let's define the configuration properties:
+
+[source]
+----
+server.url=http://${server.host}:{server.port}/endpoint
+server.host=example.org
+----
+
+A call to `Config.getValue("server.url", String.class)` will throw a `NoSuchElementException` since the
+`server.port` property is not configured.
+Likewise, a call to `Config.getOptionalValue("server.url", String.class)` will return an `Optional.empty()`.
+
+==== Default value
+
+A variable can define a _default_ value to be evaluated if the corresponding config property is not present.
+
+For example, let's define the configuration properties:
+
+[source]
+----
+server.url=http://${server.host:example.org}:{server.port:8080}/endpoint
+----
+
+If neither `server.host` nor `server.port` config properties are present, the Config API will use the variable's default values and will return
+`http://example.org:8080/endpoint` for the `server.url`.
+
+It is possible to define an empty default value for a variable.
+If the key is not present in the Config, it will not be evaluated.
+
+For example, let's define the configuration properties:
+
+[source]
+----
+server.url=http://example.org:8080/endpoint#${server.anchor:}
+----
+
+When the `server.anchor` configuration property is not present, the Config API will return
+`http://example.org:8080/endpoint#` for the `server.url`.
+
+If the `server.anchor` is configured with the value `id123`, the Config API will return
+`http://example.org:8080/endpoint#id123` for the `server.url`.
+
+A default value does not support variable evaluation (a default value is _raw_).
+
+For example, let's define the configuration properties:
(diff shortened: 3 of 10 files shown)


jmesnil:
The feature breaks backward compatibility (as explained in the spec) but this feature is too useful to be disabled by default.
We can mandate that implementation provides backwards compatibility by disabling variable expansion (with a configuration property `mp-config.expand-variables`).

With that requirement, could we add this feature to MP Config 1.4 (without having to bump the major version)?

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessorBuilder.java
```diff
@@ -88,23 +88,23 @@
...
     /**
-     * Whether to evaluate variables in configured values.
+     * Whether to expand variables in configured values.
      * A variable starts with '${' and ends with '}', e.g.
      * <pre>
      * mycompany.some.url=${myserver.host}/some/path
      * myserver.host=http://localhost:8081
      * </pre>
-     * If 'evaluateVariables' is enabled, the result for the above key
+     * If 'expandVariables' is enabled, the result for the above key
```
Emily-Jiang: Not sure about the name of expandVariables.

jmesnil: I don't mind the name. The original PR was using either "evaluate variables" or "replace variables".
As long as we consistently use the same term through the code and spec, I'm fine with it.

spec/src/main/asciidoc/variable-expansion.asciidoc
```diff
@@ -0,0 +1,137 @@
...
+It is possible to have two configuration properties referencing each other with variables:
+
+[source]
+----
+prop1 = ${prop2}
+prop2 = ${prop1}
+----
+
+Such infinite variable expansion is not supported and the Config API will throw a `NoSuchElementException` (or
+return an `Optional.empty()`) value for both properties.
```
dmlloyd: This requires a possibly expensive "visited" set to be maintained during expansion.  Could there just be a simple numerical recursion limit instead?

spec/src/main/asciidoc/variable-expansion.asciidoc
```diff
@@ -0,0 +1,137 @@
...
+[NOTE]
+====
+Backwards Compatibility
+
+Variable expansion is not backwards compatible.
+Previous version of MicroProfile Config would not expand the variables and would return the _raw_ property `http://${server.host}:{server.port}/endpoint`.
+
+Implementation of MicroProfile Config MUST provide a way to disable variable expansion to provide backwards compatibility.
+
+The property `mp-config.expand-variables` can be configured as a boolean in one of the 3 default `ConfigSource` (environment variables, system properties,
```
jmesnil: TODO: rename it to `mp.config.expandVariables` to be consistent with other MP specs

spec/src/main/asciidoc/variable-expansion.asciidoc
dmlloyd: Also returning `empty` is potentially hazardous: I think it would be better to flag the error rather than treat the property as missing.  Treating it as missing might result in subtle problems later, whereas an expansion failure resulting in a different exception (say, `IllegalStateException`) would indicate the problem sooner (fail-fast).

spec/src/main/asciidoc/variable-expansion.asciidoc
dmlloyd: Could this be implemented as a default method on `ConfigSource`, whose default implementation is to evaluate that property?  This would allow for config source implementations that always or never expand properties.  For example, we've already run into trouble with property expansion in environment variables which contain `${..}` sequences.

spec/src/main/asciidoc/variable-expansion.asciidoc
jmesnil: @dmlloyd I will rephrase it. "Infinite" variable expansion is not correct, something along the lines of "if variable expansion can not be resolved in a limited steps, etc."

You're right about failing fast, raising an exception should be the same for optional or required values. I'll update the PR

spec/src/main/asciidoc/variable-expansion.asciidoc
jmesnil: The intent of this property is a *global* switch that applies at the Config level (either try to expand all properties or not).

If I follow you, you want to be able to configure variable expansion on a `ConfigSource` level.
E.g.

```
# config_source_1
mp.config.source.expandVariables=true
host=localhost
port=8080
server=http://${host}:${port}

# config_source_2
mp.config.source.expandVariables=false
server=http://${host}:${port}
```

If `config_source_1` has higher ordinal, `config.getValue("server", String.class)` would return `http://localhost:8080` (as its variables are expanded).
If `config_source_2` has higher ordinal, `config.getValue("server", String.class)` would return `http://${host}:${port}` (as its variables are *not* expanded).

@dmlloyd is that the behaviour you are expecting?

dmlloyd: No, I had just misinterpreted the description of the functionality.  I thought the property applied to each config source but the specification states that it must *be configured in* one of the default config sources to apply (globally).

That said, we did internally talk about the fact that we probably do not want to allow expansion within environment variable values as this could cause issues with certain environments.

spec/src/main/asciidoc/variable-expansion.asciidoc
dmlloyd: OK.  What about the idea of using a numerical recursion limit, to avoid having to maintain a visited set?

jmesnil: I think the mechanism to detect such recursion (visited set or recursion limit) should not be mandated by the spec and be up to the implementation.
TCK should just check that the implementation raises an exception for such use case.

I'll remove this section from the spec and just keep the TCK test.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
```diff
@@ -95,6 +95,17 @@
      */
     ConfigBuilder withConverters(Converter<?>... converters);
 
+    /**
+     * Determines whether variable expansion is enabled.
+     *
+     * This enables variable expansion globally for any `Config` built by this builder.
+     * However, it is possible to override this by using
```
tevans78: The word "override" may be misleading. This value should have no impact on the ConfigAccessorBuilder value. We should add a reference that this only defines the behaviour of Config.getValue and Config.getOptionalValue.

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessorBuilder.java
struberg: -1 on this change. 

I thought about 'evaluateVariables' vs 'expandVariables' for over a month now.
Both are ok-ish, but I don't see how evaluateVariables is wrong!
It explains exactly what is happening. Take a configured value and evaluate all ${} variables in it.
This is also a well known feature from MyFaces CODI, DeltaSpike-Config and Tamaya where exactly this term is being used. So if we want those users to easily migrate over then we should not change it if there is no really good argument.

jmesnil: as I said in 
https://github.com/eclipse/microprofile-config/pull/405#discussion_r259766498, I don't mind the name as long as it is consistently used throughout the API and spec

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
jmesnil: +1, I'll rephrase

spec/src/main/asciidoc/variable-evaluation.asciidoc
```diff
@@ -0,0 +1,123 @@
...
+[NOTE]
+====
+Backwards Compatibility
+
+Variable evaluation is not backwards compatible.
+Previous version of MicroProfile Config would not evalute the variables and would return the value `http://${server.host}:{server.port}/endpoint`.
+
+Implementation of MicroProfile Config MUST provide a way to disable variable evaluation to provide backwards compatibility.
+
+The property `mp.config.evaluateVariables` can be configured as a boolean in one of the 3 default `ConfigSource` (environment variables, system properties,
```
Emily-Jiang: -1 on specifying the location of the property. It should be allowed to be specified anywhere as a config, as long as it is available on application starting

dmlloyd: If I understand correctly, the property is supposed to apply on a per-source basis only, so that some sources can have expressions and some not.

Emily-Jiang:
Not per source but per application

dmlloyd:
> Not per source but per application

I guess this was meant to be in reply to https://github.com/eclipse/microprofile-config/pull/405#discussion_r270394818 ?

There are two points where expansion could be decidable.  One is per-source (for example, the environment often can contain things that look like expressions but would fail on expansion), and the other is the application.  I believe the specification allows both.  But the paragraph in question that you responded to seems to be referring to controlling expansion by source, not by application.

spec/src/main/asciidoc/variable-evaluation.asciidoc
jmesnil: @emilyjiang we can not specify this property that is using to configure the `Config` itself on any `ConfigSource` or we end up with a chicken-and-egg situation.

Let's say we have a 3rd-party Zookeeper ConfigSource that calls `ConfigProvider.getConfig()` to get its own configuration (to be able to connect to ZK).

If we look for any `ConfigSource` to get this `mp.config.evaluateVariables`, we will call ZooKeep ConfigSource that will call `ConfigProvider.getConfig()` to get its configuration (to be able to fetch its properties). In turn this will call the method responsible to get the `mp.config.evaluateVariables` property, etc.

Restricting to the 3 default ConfigSource avoid this situation. This property is a special case: it it used to configure the `ConfigBuilder` itself, not the application. It seems ok to me to limit its use.

spec/src/main/asciidoc/variable-evaluation.asciidoc
```diff
@@ -0,0 +1,123 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+[[variable-evaluation]]
+== Variable Evaluation
+
+The value of a configuration property can contains a variable corresponding to another configuration property.
```
gunnarmorling:
```suggestion
The value of a configuration property can contain a variable corresponding to another configuration property.
```

svenhaag:
Hi folks, why can't we have a more generic approach to this? I think of invoking (custom) interceptors/filters before returning a value. Each interceptor should be provided with a context that e.g. contains all (unresolved) config properties from `ConfigProvider#getConfig()`.

With this, a variable expansion would be just an implementation of this interface.

dmlloyd:
In SmallRye, we provide an API to allow a `ConfigSource` to act as a wrapper for another `ConfigSource`, which is interceptor-like.  This is, as you suggest, exactly how we implement variable expansion in Quarkus, and multiple other features as well.  We could do something similar in the spec.

spec/src/main/asciidoc/variable-expansion.asciidoc
pasumarthivijaykumar: > The intent of this property is a _global_ switch that applies at the Config level (either try to expand all properties or not).
> 
> If I follow you, you want to be able to configure variable expansion on a `ConfigSource` level.
> E.g.
> 
> ```
> # config_source_1
> mp.config.source.expandVariables=true
> host=localhost
> port=8080
> server=http://${host}:${port}
> 
> # config_source_2
> mp.config.source.expandVariables=false
> server=http://${host}:${port}
> ```
> 
> If `config_source_1` has higher ordinal, `config.getValue("server", String.class)` would return `http://localhost:8080` (as its variables are expanded).
> If `config_source_2` has higher ordinal, `config.getValue("server", String.class)` would return `http://${host}:${port}` (as its variables are _not_ expanded).
> 
> @dmlloyd is that the behaviour you are expecting?

Is this implemented in already and can you please let us know how to use it and details.?

radcortez:
This should be a mandatory feature for 2.0.

The question is, how to implement it.

At this time, I believe we should expose an ahead of time declaration of behaviour (either with hooks, interceptors, filters) to make this more flexible and also has suggested by @svenhaag and discussed in #524.

dmlloyd:
I think we should close this old PR and open one or more new `Use Case` issue(s) to cover this functionality.  WDYT?

dmlloyd:
Closing this PR, as we discussed previously/

Issue #397, opened 2019-01-16
Need a way to specify empty list/array in microprofile-config.properties


njr-11:
The MicroProfile Config spec currently specifies that Array Converters are built in, and provides examples of how to configure values that are multiple array elements (delimit by `,`).  However, it does not appear to document how to configure a value that is an empty array.

On one implementation, I tried out specifying an empty value, like this,
```
test.property.name=
```
but it ended up with a value that is an array of size 1 containing the empty string.

I also tried out specifying the value as the value of the org.eclipse.microprofile.config.inject.ConfigProperty.UNCONFIGURED_VALUE constant (lacks documentation, so not sure if this is proper usage. Also, it seems to have two different values depending on which version of the ConfigProperty class is used),
```
test.property.name.1=org.eclipse.microprofile.config.configproperty.unconfigureddvalue
test.property.name.2=org.eclipse.microprofile.config.configproperty.unconfiguredvalue

```
but this resulted in a size 1 array containing the specified constant value.

It's likely the above behavior is to be expected, but the requirement to be able to specify an empty list also seems valid.  In another MicroProfile spec, we are utilizing MicroProfile Config to allow the user to supply configuration property values, some of which are lists/arrays.  It ought to be possible for MicroProfile Config to have a standard way to set these list/array values to empty.

Emily-Jiang:
`test.property.name=` is the way to specify an empty string.
`test.property.name.1=org.eclipse.microprofile.config.configproperty.unconfigureddvalue`
`test.property.name.2=org.eclipse.microprofile.config.configproperty.unconfiguredvalue`
The above 2 lines are to specify two strings.

hutchig:
An Optional which stores a Collection which it indicates is not null but that has no elements. Using converters.

njr-11:
@hutchig I understand that a component can write its own Converter to interpret a value as an empty array/list/collection, although that doesn't really address the problem of how the end user should expect to disambiguate between String[] { "" } and String[] {} in microprofile-config.properties.  I was expecting that MicroProfile Config would standardize some sort of syntax for this.  It looks like you have an escape character, `\`, which can be used when list values include commas.  Could the escape character be used somehow to represent that an empty list is wanted?

njr-11:
[...]

PR #407, opened 2019-02-15
Specify handling of empty config property

jmesnil:
* [Spec] add section explaining how the Config API handles a config
  property with missing or empty value
* [Spec] mention that it is possible to set an empty value to reset a
  property from a lower-ordinal config source.
* [TCK] add EmptyConfigPropertyTest to test all cases of empty config
  property

### DIFF
--- a/spec/src/main/asciidoc/converters.asciidoc
+++ b/spec/src/main/asciidoc/converters.asciidoc
@@ -104,3 +104,44 @@ If no built-in nor custom `Converter` exists for a requested Type `T`, an implic
 
 If a `Converter` implements the `java.lang.AutoCloseable` interface  then the `close()` method will be called when the underlying `Config` is being released.
 
+=== Empty value
+
+A config property can be configured with an empty value (the empty String `""`).
+
+The config API interprets an empty string as a missing property and throws a `NoSuchElementException` when the output type is a `String`.
+
+[NOTE]
+This behaviour allows to _unconfigure_ a property. If a property is configured in a lower-ordinal ConfigSource with
+an undesirable value, you can set the property value to an empty string in a higher-ordinal ConfigSource to discard the lower-ordinal value.
+
+The Config API also interprets an empty string as an empty array when the output type is `String[]`.
+For consistency, the Config API also discards any leading and trailing commas `","` when the output type is a `String[]``
+
+[[empty_value_table]]
+.Examples of Config API
+[options="header"]
+|=======================
+| Property value | Output type | `Config` method | Output result
+| missing     | `String` | `getValue` | throws `NoSuchElementException`
+| `""` (empty string) | `String` | `getValue` | throws `NoSuchElementException`
+| `" "`       | `String` | `getValue` | `" "`
+| `"foo"`     | `String`   | `getValue` |  `"foo"`
+| missing     | `String[]` | `getValue` | `{ }` (empty array)
+| `""`        | `String[]` | `getValue` | `{ }`
+| `" "`       | `String[]` | `getValue` | `{ " " }`
+| `","`       | `String[]` | `getValue` | `{ }`
+| `",,"`      | `String[]` | `getValue` | `{ }`
+| `"foo"`     | `String[]` | `getValue` | `{ "foo" }`
+| `"foo,bar"` | `String[]` | `getValue` | `{ "foo", "bar" }`
+| `"foo,"`    | `String[]` | `getValue` | `{ "foo" }`
+| `",bar"`    | `String[]` | `getValue` | `{ "bar" }`
+| `",bar,"`   | `String[]` | `getValue` | `{ "bar" }`
+| missing     | `String` |`getOptionalValue` | `Optional.empty()`
+| `""`        | `String` | `getOptionalValue` | `Optional.empty()`
+| `" "`        | `String` | `getOptionalValue` | `Optional.of(" ")`
+| `"foo"`     | `String` | `getOptionalValue` | `Optional.of("foo")`
+| missing | `String[]` | `getOptionalValue` | `Optional.empty()`
+| `""` | `String[]` | `getOptionalValue` | `Optional.empty()`
+| `","` | `String[]` | `getOptionalValue` | `Optional.empty()`
+| `",,"` | `String[]` | `getOptionalValue` | `Optional.empty()`
+| `"foo,bar"` | `String[]` | `getOptionalValue` | `Optional.of({ "foo", "bar" })`
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyConfigPropertyBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyConfigPropertyBean.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.config.tck;
+
+import static org.eclipse.microprofile.config.tck.EmptyConfigPropertyTest.EMPTY_PROP;
+
+import java.util.List;
+import java.util.Set;
+
+import javax.enterprise.context.Dependent;
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@Dependent
+public class EmptyConfigPropertyBean {
+
+    @Inject
+    @ConfigProperty(name = EMPTY_PROP)
+    private String[] myStrings;
+
+    @Inject
+    @ConfigProperty(name = EMPTY_PROP)
+    private List<String> myStringList;
+
+    @Inject
+    @ConfigProperty(name = EMPTY_PROP)
+    private Set<String> myStringSet;
+
+    public String[] getMyStrings() {
+        return myStrings;
+    }
+
+    public List<String> getMyStringList() {
+        return myStringList;
+    }
+
+    public Set<String> getMyStringSet() {
+        return myStringSet;
+    }
+}
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyConfigPropertyTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyConfigPropertyTest.java
@@ -1,0 +1,206 @@
+/*
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package org.eclipse.microprofile.config.tck;
+
+import static org.eclipse.microprofile.config.tck.base.AbstractTest.addFile;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.fail;
+
+import java.util.NoSuchElementException;
+import java.util.Optional;
+
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
+import org.eclipse.microprofile.config.tck.configsources.KeyValuesConfigSource;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.testng.Arquillian;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class EmptyConfigPropertyTest  extends Arquillian {
+
+    @Deployment
+    public static WebArchive deploy() {
+        JavaArchive testJar = ShrinkWrap
+            .create(JavaArchive.class, "EmptyConfigPropertyTest.jar")
+            .addClass(EmptyConfigPropertyBean.class)
+            .addClass(KeyValuesConfigSource.class)
+            .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+            .as(JavaArchive.class);
+
+        addFile(testJar, "META-INF/microprofile-config.properties");
+
+        WebArchive war = ShrinkWrap
+            .create(WebArchive.class, "EmptyConfigPropertyTest.war")
+            .addAsLibrary(testJar);
+        return war;
+    }
+
+
+    private @Inject
+    Config config;
+
+    @Inject
+    private EmptyConfigPropertyBean bean;
+
+    static final String EMPTY_PROP = "tck.config.empty";
+    static final String MISSING_PROP = "tck.config.this.property.is.not.defined.anywhere";
+
+    @Test
+    public void testEmptyPropertyWithConfig() {
+        // | Property value | Output type | `Config` method | Output result
+        // | `""` (empty string) | `String` | `getValue` | throws `NoSuchElementException`
+        try {
+            config.getValue(EMPTY_PROP, String.class);
+            Assert.fail("An empty string is considered as missing and must raise a NoSuchElementException");
+        }
+        catch (NoSuchElementException e) {
+        }
+        try {
+            config.getValue(EMPTY_PROP, Double.class);
+            Assert.fail(("an empty String is considered missing regardless of the ouput type"));
+        }
+        catch (NoSuchElementException e) {
+        }
+
+        // but if it is accessed an array, it will returns an empty array
+        // | `""`        | `String[]` | `getValue` | `{ }`
+        assertEquals(config.getValue(EMPTY_PROP, String[].class).length, 0);
+        // regardless of the output type
+        assertEquals(config.getValue(EMPTY_PROP, Double[].class).length, 0);
+
+
+        // | `""`        | `String` | `getOptionalValue` | `Optional.empty()`
+        assertEquals(config.getOptionalValue(EMPTY_PROP, String.class), Optional.empty());
+        assertEquals(config.getOptionalValue(EMPTY_PROP, Double.class), Optional.empty());
+
+        // | `""` | `String[]` | `getOptionalValue` | `Optional.empty()`
+        assertEquals( config.getOptionalValue(EMPTY_PROP, String[].class), Optional.empty());
+        assertEquals( config.getOptionalValue(EMPTY_PROP, Double[].class), Optional.empty());
+    }
+
+    @Test
+    public void testMissingProperty() {
+        // | Property value | Output type | `Config` method | Output result
+        // | missing     | `String` | `getValue` | throws `NoSuchElementException`
+        try {
+            config.getValue(MISSING_PROP, String.class);
+            Assert.fail("A missing property must raise a NoSuchElementException");
+        }
+        catch (NoSuchElementException e) {
+        }
+        // | missing     | `String[]` | `getValue` | `{ }` (empty array)
+        assertEquals(config.getValue(MISSING_PROP, String[].class).length, 0);
+        // regardless of the output type
+        assertEquals(config.getValue(MISSING_PROP, Double[].class).length, 0);
+
+        // | missing     | `String` |`getOptionalValue` | `Optional.empty()`
+        assertEquals( config.getOptionalValue(MISSING_PROP, String.class), Optional.empty());
+        // | missing | `String[]` | `getOptionalValue` | `Optional.empty()`
+        assertEquals( config.getOptionalValue(MISSING_PROP, String[].class), Optional.empty());
+    }
+
+    @Test
+    public void testArrayCommasRules() {
+        Config config = KeyValuesConfigSource.buildConfig(
+            "arrayWithSpace", " ",
+            "arrayWithSingleComma", ",",
+            "arrayWithTwoCommas", ",,",
+            "arrayWithSingleValue", "foo",
+            "arrayWithTwoValues", "foo,bar",
+            "arrayWithTrailingComma", "foo,",
+            "arrayWithLeadingComma", ",bar",
+            "arrayWithLeadingAndTrailingCommas", ",bar,");
+
+        // | Property value | Output type | `Config` method | Output result
+        // | `" "`       | `String[]` | `getValue` | `{ " " }`
+        assertEquals(config.getValue("arrayWithSpace", String[].class), new String[] { " " });
+        // | `","`       | `String[]` | `getValue` | `{ }`
+        assertEquals(config.getValue("arrayWithSingleComma", String[].class).length, 0);
+        // | `",,"`      | `String[]` | `getValue` | `{ }`
+        assertEquals(config.getValue("arrayWithTwoCommas", String[].class).length, 0);
+        // | `"foo"`     | `String[]` | `getValue` | `{ "foo" }`
+        assertEquals(config.getValue("arrayWithSingleValue", String[].class), new String[] { "foo" });
+        // | `"foo,bar"` | `String[]` | `getValue` | `{ "foo", "bar" }`
+        assertEquals(config.getValue("arrayWithTwoValues", String[].class), new String[] { "foo", "bar" });
+        // | `"foo,"`    | `String[]` | `getValue` | `{ "foo" }`
+        assertEquals(config.getValue("arrayWithTrailingComma", String[].class), new String[] { "foo" });
+        // | `",bar"`    | `String[]` | `getValue` | `{ "bar" }`
+        assertEquals(config.getValue("arrayWithLeadingComma", String[].class), new String[] { "bar" });
+        // | `",bar,"`   | `String[]` | `getValue` | `{ "bar" }`
+        assertEquals(config.getValue("arrayWithLeadingAndTrailingCommas", String[].class), new String[] { "bar" });
+
+        // | `","` | `String[]` | `getOptionalValue` | `Optional.empty()`
+        assertEquals(config.getOptionalValue("arrayWithSingleComma", String[].class), Optional.empty());
+        // | `",,"` | `String[]` | `getOptionalValue` | `Optional.empty()`
+        assertEquals(config.getOptionalValue("arrayWithTwoCommas", String[].class), Optional.empty());
+    }
+
+    @Test
+    public void testEmptyPropertyWithConfigAccessor() {
+        try {
+            config.access(EMPTY_PROP, String.class).build().getValue();
+            Assert.fail("An empty string is considered as missing and must raise a NoSuchElementException");
+        }
+        catch (NoSuchElementException e) {
+
+        }
+        assertEquals(config.access(EMPTY_PROP, String.class).build().getOptionalValue(), Optional.empty());
+
+        assertEquals(config.access(EMPTY_PROP, String[].class).build().getValue().length, 0);
+        assertEquals(config.access(EMPTY_PROP, String[].class).build().getOptionalValue(), Optional.empty());
+    }
+
+    @Test
+    public void testEmptyPropertyWithInjectedConfigProperty() {
+        assertEquals(bean.getMyStrings().length, 0);
+        assertEquals(bean.getMyStringList().size(), 0);
+        assertEquals(bean.getMyStringSet().size(), 0);
+    }
+
+    @Test
+    public void testEmptyPropertyResetsLowerOrdinalProperty() {
+        Config config = ConfigProviderResolver.instance().getBuilder()
+            .withSources(
+                // lower-ordinal configures a prop
+                KeyValuesConfigSource.config(10,
+                    "my.prop", "foo"),
+                // higher-ordinal resets it with an empty string
+                KeyValuesConfigSource.config(20,
+                    "my.prop", ""))
+            .build();
+
+        try {
+            config.getValue("my.prop", String.class);
+            fail("The property must be reset by the empty string in the higher-ordinal config source");
+        }
+        catch (NoSuchElementException e) {
+        }
+
+        assertEquals(config.getOptionalValue("my.prop", String.class), Optional.empty());
+    }
+
+}
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValuesTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValuesTest.java
@@ -19,6 +19,12 @@
  */
 package org.eclipse.microprofile.config.tck;
 
+import static org.testng.Assert.assertEquals;
+
+import java.util.NoSuchElementException;
+
+import javax.inject.Inject;
+
 import org.eclipse.microprofile.config.Config;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.testng.Arquillian;
@@ -29,10 +35,6 @@ import org.jboss.shrinkwrap.api.asset.StringAsset;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.testng.annotations.Test;
-
-import javax.inject.Inject;
-
-import static org.testng.Assert.assertEquals;
 
 public class EmptyValuesTest extends Arquillian {
 
@@ -63,17 +65,31 @@ public class EmptyValuesTest extends Arquillian {
         assertEquals(emptyValuesBean.getStringValue(), "");
     }
 
-    @Test
+    @Test(expectedExceptions = NoSuchElementException.class)
     public void testEmptyStringProgrammaticLookup() {
-        System.setProperty(EMPTY_PROPERTY, "");
-        String stringValue = config.getValue(EMPTY_PROPERTY, String.class);
-        assertEquals(stringValue, "");
-        System.clearProperty(EMPTY_PROPERTY);
+        try {
+            System.setProperty(EMPTY_PROPERTY, "");
+            config.getValue(EMPTY_PROPERTY, String.class);
+        }
+        finally {
+            System.clearProperty(EMPTY_PROPERTY);
+        }
+    }
+
+    @Test(expectedExceptions =  NoSuchElementException.class)
+    public void testEmptyStringPropertyFromConfigFile() {
+        config.getValue(PROP_FILE_EMPTY_PROPERTY, String.class);
     }
 
     @Test
-    public void testEmptyStringPropertyFromConfigFile() {
-        String stringValue = config.getValue(PROP_FILE_EMPTY_PROPERTY, String.class);
-        assertEquals(stringValue, "");
+    public void testEmptyStringLookupAsArray() {
+        try {
+            System.setProperty(EMPTY_PROPERTY, "");
+            String[] result = config.getValue(EMPTY_PROPERTY, String[].class);
+            assertEquals(result.length, 0);
+        }
+        finally {
+            System.clearProperty(EMPTY_PROPERTY);
+        }
     }
 }
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/KeyValuesConfigSource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/KeyValuesConfigSource.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2016-2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.config.tck.configsources;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+public class KeyValuesConfigSource implements ConfigSource {
+
+
+    private final Map<String, String> properties = new HashMap<>();
+    private final int ordinal;
+
+    private KeyValuesConfigSource(Map<String, String> properties, int ordinal) {
+        this.properties.putAll(properties);
+        this.ordinal = ordinal;
+    }
+
+    @Override
+    public Map<String, String> getProperties() {
+        return Collections.unmodifiableMap(properties);
+    }
+
+    @Override
+    public String getValue(String propertyName) {
+        return properties.get(propertyName);
+    }
+
+    @Override
+    public int getOrdinal() {
+        return ordinal;
+    }
+
+    @Override
+    public String getName() {
+        return "KeyValuesConfigSource";
+    }
+
+    public static ConfigSource config(String... keyValues) {
+        return config(DEFAULT_ORDINAL, keyValues);
+    }
+
+    public static ConfigSource config(int ordinal, String... keyValues) {
+        if (keyValues.length %2 != 0) {
+            throw new IllegalArgumentException("keyValues array must be a multiple of 2");
+        }
+
+        Map<String, String> props = new HashMap<>();
+        for (int i = 0; i < keyValues.length; i += 2) {
+            props.put(keyValues[i], keyValues[i+1]);
+        }
+        return new KeyValuesConfigSource(props, ordinal);
+    }
+
+    public static Config buildConfig(String... keyValues) {
+        return ConfigProviderResolver.instance().getBuilder()
+            .withSources(KeyValuesConfigSource.config(keyValues))
+            .build();
+    }
+}
--- a/tck/src/main/resources/internal/META-INF/microprofile-config.properties
+++ b/tck/src/main/resources/internal/META-INF/microprofile-config.properties
@@ -160,3 +160,6 @@ tck.config.test.javaconfig.converter.class.array=org.eclipse.microprofile.config
 tck.config.variable.baseHost = some.host.name
 tck.config.variable.firstEndpoint = http://${tck.config.variable.baseHost}/endpointOne
 tck.config.variable.secondEndpoint = http://${tck.config.variable.baseHost}/endpointTwo
+
+# empty config property test
+tck.config.empty=


spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -104,3 +104,19 @@ If no built-in nor custom `Converter` exists for a requested Type `T`, an implic
 
 If a `Converter` implements the `java.lang.AutoCloseable` interface  then the `close()` method will be called when the underlying `Config` is being released.
 
+=== Empty value
+
+A config property can be configured with an empty value (the empty String `""`).
+Such config property will be interpreted by the Config API as follows:
+
+* `Config.getValue(key, String.class)` returns `""` (empty String)
```
dmlloyd: I disagree strongly with this behavior.  This makes it impossible to have a configuration source which overrides a present value with a non-present value.

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -104,3 +104,19 @@ If no built-in nor custom `Converter` exists for a requested Type `T`, an implic
...
+* `Config.getValue(key, String[].class)` returns a 0-sized `String[]` array
```
dmlloyd: Having an empty string be an empty string is implicitly contradictory with this requirement.  If `,` returns an array of `[ "", "" ]` then the empty string must return an array of `[ "" ]`.  The only logical and consistent solution is to have arrays (and lists) ignore empty strings and have an empty string be identical to "not present".

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -104,3 +104,19 @@ If no built-in nor custom `Converter` exists for a requested Type `T`, an implic
...
+* `Config.getValue(key, X.class)`
+** if X is an array type, it returns a 0-sized array of the given type `X`
+** else it returns either an object of the type `X` (if it can be converted from an empty string) or throws a `IllegalArgumentException`
+* `Config.getOptionalValue(key, String.class)` returns `Optional.of("")`
```
dmlloyd: Then this should be `Optional.empty()`.

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -104,3 +104,19 @@ If no built-in nor custom `Converter` exists for a requested Type `T`, an implic
...
+* `Config.getOptionalValue(key, X[].class)` returns an `Optional` of a 0-sized array of the given type `X`
```
dmlloyd: And this should also be `Optional.empty()`.

spec/src/main/asciidoc/converters.asciidoc
struberg: yes, this is against the possibility to detect 'nils'. This would break other features, like to 'unconfigure' values.

dmlloyd:
Here is my basic problem with the current and proposed rules.  An empty string is sometimes considered non-present, and sometimes considered a present empty string.  This inconsistency causes a few problems:

* Inconsistency of array element counting: the size of the array is *n* + 1 (where *n* is the number of commas) - *except* when *n* is zero, meaning an array or list can never be produced which contains exactly one empty string if this rule is followed in this way, which is pretty surprising.
* It becomes impossible for a higher priority configuration source to override a lower value with "not present", because the only "not present" value is `null` which indicates that a lower priority source must be consulted.
* It is not and should not be meaningful to applications to differentiate between a non-present value and an empty string; even with the best of intentions this would lead to user confusion.  It would always be better to use a separate `boolean` configuration switch in cases where one would be tempted to make such a distinction.

It would be far, far more sensible to always treat *all* empty values as an explicit "not present", which allows us to have much more consistent rules.

Input string | Output type | Current rules | Better rules
------------- | ------------- | ------------- | ----------------
missing | `String[]` | `{ }` | `{ }`
`""` | `String[]` | `{ }` | `{ }`
`","` | `String[]` | `{ "", "" }` | `{ }`
`",,"` | `String[]` | `{ "", "", "" }` | `{ }`
`"foo,bar"` | `String[]` | `{ "foo", "bar" }` | `{ "foo", "bar" }`
`"foo,"` | `String[]` | `{ "foo", "" }` | `{ "foo" }`
`",bar"` | `String[]` | `{ "", "bar" }` | `{ "bar" }`
`" "` | `String[]` | `{ " " }` | `{ " " }`
`"foo"` | `String` | `"foo"` | `"foo"`
`""` | `String` | `""` | `null`
missing | `String` | `null` | `null`
`"foo"` | `Optional<String>` | `Optional.of("foo")` | `Optional.of("foo")`
`""` | `Optional<String>` | `Optional.of("")` | `Optional.empty()`
missing | `Optional<String>` | `Optional.empty()` | `Optional.empty()`
`""` | `Optional<String[]>` | `Optional.of({})` | `Optional.empty()`
`","` | `Optional<String[]>` | `Optional.of({ "", "" })` | `Optional.empty()`
`",,"` | `Optional<String[]>` | `Optional.of({ "", "", "" })` | `Optional.empty()`
missing | `Optional<String[]>` | `Optional.empty` | `Optional.empty()`

spec/src/main/asciidoc/converters.asciidoc
jmesnil: The spec never mentions anything about "unconfiguring" the properties.
My understanding of the spec is that the configsource only allows to overrides values, not to "delete/unconfigure" them.

If a lower configsource defines `a=foo` and I override that with a higher configsource to be `a= ` (empty value). I expect the Config API to take my value and returns ``.

struberg: Probably we never wrote this down but I only explained this in the past. The original handing in DeltaSpike was to have an empty string 'erase' any configured value in a lower ordinal ConfigSource. This turned out to be quite useful. Especially now with `Optional<T>` it is otherwise possible to come back to the orElse. Do you have a better idea?

jmesnil:
@dmlloyd thanks for the table, it clearly identifies all the use cases (and should serve as the basic of a TCK test)

I thinks the following ones are not accurate wrt to the current rules (and disagree with the `""` better rules)

Input string | Output type | Current rules | Better rules
-- | -- | -- | --
missing | String[] | throws NoSuchElementException | { }
"" | String | "" | ""
missing | String | throws NoSuchElementException | throws NoSuchElementException

I have no issue with changing the rule so that a missing `String[]` returns an empty array instead of throwing an exception.
However I still think than empty string should be returned as an empty string and not `null`.
Likewise a missing String should still throw NoSuchElementException and not return `null`.

dmlloyd:
> @dmlloyd thanks for the table, it clearly identifies all the use cases (and should serve as the basic of a TCK test)
> 
> I have no issue with changing the rule so that a missing `String[]` returns an empty array instead of throwing an exception.
> However I still think than empty string should be returned as an empty string and not `null`.

Unfortunately this just cannot be made to work consistently, especially when custom converters are involved.  If an empty string is an empty string, then arrays created from empty strings must be `{ "" }` which is nonsense.  Otherwise we get inconsistent values.

> Likewise a missing String should still throw NoSuchElementException and not return `null`.

Ah you're right, that was my mistake.  I'll update the table accordingly.

spec/src/main/asciidoc/converters.asciidoc
dmlloyd: > The spec never mentions anything about "unconfiguring" the properties.
> My understanding of the spec is that the configsource only allows to overrides values, not to "delete/unconfigure" them.

This to me is a mistake and we should amend the specification.

> If a lower configsource defines `a=foo` and I override that with a higher configsource to be `a= ` (empty value). I expect the Config API to take my value and returns ``.

As I stated elsewhere, there is no possible meaningful distinction between empty and missing.  At *best* it encourages poor behavior and UX.

jmesnil: I agree with @dmlloyd proposal to modify the Optional case:


Input string | Output type | Current rules | Better rules
-- | -- | -- | --
"" | Optional<String> | Optional.of("") | Optional.empty()

This would make it possible to use the `Optional<T>` to "unconfigure" the value and use the `orElse`.

However, I'm not convinced and need to think more how changing the behaviour for the plain `getValue`.

If I understand you correctly, If I have a lower config source with `a=foo` that I want to configure, I'd have to provide a higher config source with `a=`.

In my code, I then have to take into account all the use cases:
* `a` is not configured anywhere, `Config.getValue(String)` throws `NoSuchElementException`
* `a` is configured in a config source, `Config.getValue(String)` return `foo`
* `a` is configured in a config source and "unconfigured" in a higher config source with an empty string. What does `Config.getValue()` return?
  * `` (empty string that I specified)
  *  `null` as David is proposing
  *  throws a `NoSuchElementException` (current rule and it is consistent with "unconfiguring" the property)

jmesnil:
@dmlloyd with your latest changes, I approved your changes.

It was bothering me to treat an `""` string as `null` but I have no issue with having an empty string throws a `NoSuchElementException`... Optics :)

This also is consistent with @struberg comment about unconfiguring properties.
With this change, an empty string value is treated as a missing config and can be used to unconfigure properties from lower config source.

However, this will change the behaviour of existing applications and must bump the MicroProfile Configuration to 2.0.

jmesnil:
@dmlloyd I've integrated your table in the spec: https://github.com/eclipse/microprofile-config/pull/407/files#diff-8faa438f11f150bc9d369db23f9044a1R120

@struberg I've added a note about resetting a property by setting it to an empty string in a higher-ordinal config source: https://github.com/eclipse/microprofile-config/pull/407/files#diff-8faa438f11f150bc9d369db23f9044a1R113

and I've based the TCK tests on the spec tables to check all these use cases

jmesnil:
This PR break compatibility as it now treats an empty string as a missing property and must require to bump the major version of the spec to 2.0

dmlloyd:
Since people are definitely using a missing value to signify that a default should be used, I think it may be worth adding a provision to the spec for making a "default values" configuration source like we do in Quarkus which has bottom priority.  It might also be a good idea to make it non-enumerable.  But this is something that can be discussed (there are pros and cons).

njr-11:
In your table, you have,

Input string | Output type | Method | Current rules | Better rules
------------- | ------------- | ------------- | ------------- | ----------------
`"foo,"` | `String[]` | `getValue` |  `{ "foo", "" }` | `{ "foo" }`
`",bar"` | `String[]` | `getValue` |  `{ "", "bar" }` | `{ "bar" }`

But what happens when the user specifies,

Input string | Output type | Method | Current rules | Better rules
------------- | ------------- | ------------- | ------------- | ----------------
`"foo,,bar"` | `String[]` | `getValue` |  `{ "foo", "", "bar" }` | `???`

It seems that for consistency with the previous two example, you would need to remove the empty string value from the middle of this list as well.  However, removing empty strings that the user has explicitly specified, whether at the beginning, end, or middle, seems wrong to me.  If you do that, how do they go about specifying an empty string when they actually want one?

Also, I want to point out that we're currently trying to write a spec that lets the user specify an empty list value in MicroProfile Config.  I opened #397 specifically for this because the MP Config spec was unclear on whether to interpret
```my.config.property=```
as String[] { "" } or String[] {}.  Due to the uncertainty, we ended up writing our spec to state that both of these values must interpreted as meaning the user wants a size 0 (empty) list.
However, If I understand the table correctly, this change makes the problem even worse, because it is saying that,
```my.config.property=```
will no longer have a value at all and will now be treated the same as if the property were completely missing from config.  To put it more succinctly, the proposed changes here do not resolve #397 - they actually make the problem worse by getting rid of the ability for the user to indicate that wanted a size 0 (empty) list.

**Comment by dmlloyd** [comment]
> In your table, you have,
> 
> Input string	Output type	Method	Current rules	Better rules
> `"foo,"`	`String[]`	`getValue`	`{ "foo", "" }`	`{ "foo" }`
> `",bar"`	`String[]`	`getValue`	`{ "", "bar" }`	`{ "bar" }`
> But what happens when the user specifies,
> 
> Input string	Output type	Method	Current rules	Better rules
> `"foo,,bar"`	`String[]`	`getValue`	`{ "foo", "", "bar" }`	`???`
> It seems that for consistency with the previous two example, you would need to remove the empty string value from the middle of this list as well.

Exactly, you'd get a two-element list.  This is also useful for expression cases like this:

```properties
foo.my.property=something,${foo.other.property:},something-else
```

> However, removing empty strings that the user has explicitly specified, whether at the beginning, end, or middle, seems wrong to me. If you do that, how do they go about specifying an empty string when they actually want one?

They can't have one under the current rules.  And note that under the current SmallRye implementation, trailing empty segments were already ignored and have been all along.

If a user wants to explicitly have empty strings, then we need a quoting mechanism.  But I think this would be a design/UX error in 90% of cases.

> Also, I want to point out that we're currently trying to write a spec that lets the user specify an empty list value in MicroProfile Config. I opened #397 specifically for this because the MP Config spec was unclear on whether to interpret
> `my.config.property=`
> as String[] { "" } or String[] {}. Due to the uncertainty, we ended up writing our spec to state that both of these values must interpreted as meaning the user wants a size 0 (empty) list.
> However, If I understand the table correctly, this change makes the problem even worse, because it is saying that,
> `my.config.property=`
> will no longer have a value at all and will now be treated the same as if the property were completely missing from config. To put it more succinctly, the proposed changes here do not resolve #397 - they actually make the problem worse by getting rid of the ability for the user to indicate that wanted a size 0 (empty) list.

The user would always get an empty list for an empty value; this would not get rid of that.  This gets rid of the logic that would result in a `String[] { "" }` array or list, which is not useful most of the time.

njr-11:
> The user would always get an empty list for an empty value; this would not get rid of that. This gets rid of the logic that would result in a `String[] { "" }` array or list, which is not useful most of the time.

That means I misinterpreted the meaning of `{ }` in your table.  Sorry about that.
However, your table still has _missing_ and `""` resolving to the same thing, which means we cannot disambiguate between the user's attempt to specify a size 0 list/array vs not specifying any override at all, and this will be a problem for our spec because we need to make that distinction,

Input string | Output type | Method | Current rules | Better rules
------------- | ------------- | ------------- | ------------- | ----------------
missing | `String[]` | `getValue` | `{ }` | `{ }`
`""` | `String[]` | `getValue` |  `{ }` | `{ }`

dmlloyd:
> > The user would always get an empty list for an empty value; this would not get rid of that. This gets rid of the logic that would result in a `String[] { "" }` array or list, which is not useful most of the time.
> 
> That means I misinterpreted the meaning of `{ }` in your table. Sorry about that.
> However, your table still has _missing_ and `""` resolving to the same thing, which means we cannot disambiguate between the user's attempt to specify a size 0 list/array vs not specifying any override at all, and this will be a problem for our spec because we need to make that distinction,

Right.  But I don't think you do need to make that distinction - I suspect what you're really trying to determine is whether the user explicitly wants no contexts to be propagated, or whether the default should be used (whatever that may be) - if I'm wrong about that, please correct me.

The spec still has a hole here that I want to fix as a part of this change.  In Quarkus for example the default configuration values exist as a config source.  So if the user does not give a key, calling `getValue()` on that key will return the default value.  The disadvantage to this approach though is that the default values have to be known at the time the configuration is created - this is something we can do in Quarkus but it's not practical for other scenarios; also it ignores the fact that, as a practical matter, default configuration values are really supported by use site today and making them all be declared up front is probably not even possible.  So I would propose a new API method, something like this:

```java
    /**
     * specified property name from the underlying {@link ConfigSource ConfigSources},
     * returning a default value if the property name is not found in any of them.
     * 
     * If this method gets used very often then consider to locally store the configured value.
     *
     * @param <T>
     *             The property type
     * @param propertyName
     *             The configuration propertyName.
     * @param propertyType
     *             The type into which the resolve property value should get converted
     * @param defVal
     *             The default value to use if no configuration source contains the configuration property name, or {@code null} to indicate that the property should be treated as non-present
     * @return the resolved property value as an object of the requested type.
     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type.
     * @throws java.util.NoSuchElementException if the property isn't present in the configuration.
     */
    <T> T getValue(String name, Class<T> aClass, T defVal);
```

This could also be extended to `getOptionalValue`.  Alternatively, `defVal` could be typed as a `String`.

This basically solves the problem of needing to fall back to a default if the key is not specified, while also allowing the user to specify no value explicitly which would override the default.

dmlloyd:
@jmesnil tells me that the new `ConfigAccessor` API already does this exact thing, so I think that this problem is already solved.

manovotn:
@njr-11 @dmlloyd I still feel you are talking about two slightly different scenarios.

So assuming I have a config `foo.bar=` and I invoke `getValue("foo.bar", String[].class)`, what do I get with the changes proposed here?
Because Smallrye currently gives me `NoSuchElementException` - treating this the same as if the key wasn't present at all.

From MP CP perspective (as it is designed now), having a config key equals empty value is perfectly fine and it doesn't mean "unconfigure" or "use default"; it translates into an empty string array.

dmlloyd:
> @njr-11 @dmlloyd I still feel you are talking about two slightly different scenarios.
> 
> So assuming I have a config `foo.bar=` and I invoke `getValue("foo.bar", String[].class)`, what do I get with the changes proposed here?

I would propose that you get an empty array, as shown in the chart, but OTOH with optional values I think it's probably better to get an `empty`, which is inconsistent, so... :man_shrugging:

> Because Smallrye currently gives me `NoSuchElementException` - treating this the same as if the key wasn't present at all.

You're right.

> From MP CP perspective (as it is designed now), having a config key equals empty value is perfectly fine and it doesn't mean "unconfigure" or "use default"; it translates into an empty string array.

So I guess it's a technical question about `getValue`.  We could detect arrays and treat them specially (though this arguably sucks, and wouldn't cover cases like custom collections).  We could go back to the old behavior which was to make the `Converter` instance decide what to do (though this is error prone). Or we could go with a behavior where empty values *always* translate into non-present which means `NoSuchElementException` for `getValue` and `empty` for `getOptionalValue`.

We would have to beef up the language around converters and what they are expected to do if we go with the second option, to make it clear that they should return `null` if they are singular or an empty collection if they are collection-oriented.  This would imply that `getOptionalValue` on collection-oriented values would always return an `Optional` with a value though, otherwise we're back to the original problem.

njr-11:
> We could detect arrays and treat them specially (though this arguably sucks, and wouldn't cover cases like custom collections). We could go back to the old behavior which was to make the `Converter` instance decide what to do (though this is error prone). Or we could go with a behavior where empty values _always_ translate into non-present which means `NoSuchElementException` for `getValue` and `empty` for `getOptionalValue`

Regardless of which approach is standardized here, we have a requirement for the user to be able to specify a size 0 (empty) array, the purpose of which is well summarized in your earlier comment,

> I suspect what you're really trying to determine is whether the user explicitly wants no contexts to be propagated, or whether the default should be used (whatever that may be) - if I'm wrong about that, please correct me.

If there are others asking for the ability to 'unconfigure' a property back to its defaults and there is trouble deciding between whether an empty string value should be used for that vs the size 0 (empty) array, would it be possible to solve the problem introducing a constant to instead be used for one of the these?  For example, the [UNCONFIGURED_VALUE](https://download.eclipse.org/microprofile/microprofile-2.0-javadocs-test/apidocs/org/eclipse/microprofile/config/inject/ConfigProperty.html#UNCONFIGURED_VALUE) constant in  org.eclipse.microprofile.config.inject.ConfigProperty which is currently available but lacks JavaDoc explaining its purpose?  Then you could do,
```
foo.bar=org.eclipse.microprofile.config.configproperty.unconfigureddvalue
```
as a very natural way of specifying that you want to override a config property such that is no longer configured and thus ends up with its default value.

dmlloyd:
> > We could detect arrays and treat them specially (though this arguably sucks, and wouldn't cover cases like custom collections). We could go back to the old behavior which was to make the `Converter` instance decide what to do (though this is error prone). Or we could go with a behavior where empty values _always_ translate into non-present which means `NoSuchElementException` for `getValue` and `empty` for `getOptionalValue`
> 
> Regardless of which approach is standardized here, we have a requirement for the user to be able to specify a size 0 (empty) array, the purpose of which is well summarized in your earlier comment,
> 
> > I suspect what you're really trying to determine is whether the user explicitly wants no contexts to be propagated, or whether the default should be used (whatever that may be) - if I'm wrong about that, please correct me.
> 
> If there are others asking for the ability to 'unconfigure' a property back to its defaults and there is trouble deciding between whether an empty string value should be used for that vs the size 0 (empty) array, would it be possible to solve the problem introducing a constant to instead be used for one of the these? For example, the [UNCONFIGURED_VALUE](https://download.eclipse.org/microprofile/microprofile-2.0-javadocs-test/apidocs/org/eclipse/microprofile/config/inject/ConfigProperty.html#UNCONFIGURED_VALUE) constant in org.eclipse.microprofile.config.inject.ConfigProperty which is currently available but lacks JavaDoc explaining its purpose? Then you could do,
> 
> ```
> foo.bar=org.eclipse.microprofile.config.configproperty.unconfigureddvalue
> ```
> 
> as a very natural way of specifying that you want to override a config property such that is no longer configured and thus ends up with its default value.

The problem I'm trying to solve here is that when a property has a default value, the user can't configure it to an *empty or missing* value.  In other words, I think the concept of a default value could be decoupled completely from the concept of a missing value.

So here's another table covering just these problem cases:

I want to | Using the existing approach | Using the proposed approach
------- | ------- | -------
Have a value in my config for an array | it's fine | it's fine
Have an array containing empty values | inconsistent for one-element arrays | not possible*
Have a default value for an array | it's fine | it's fine
Override an earlier value with another value | it's fine | it's fine
Override a default value with another value | it's fine | it's fine
Override a default value with an empty value | depends on the converter | it's fine
Override a default value with a non-present value | not possible | it's fine (same as empty value)
Override a specified value with an empty value | depends on the converter | it's fine
Override a specified value with a non-present value | not possible | it's fine (same as empty value)
Override a specified value with the default value | not possible* | not possible*

The * indicates that you can change the rules with custom converters.

Emily-Jiang:
wow. This is a long conversation. I don't agree with the suggestion though. I would like to discuss this in more details on a hangout.
At the moment:
my.property=
MP Config interpret it exists with an empty value.

In the past we were required to differentiate empty value with nulls. Now, some of you suggested it is wrong to differentiate them. I am not convinced.

By the way, the spec does not state an empty string means erasing a value. It is DeltaSpike feature not MP Config. When we try to support erasing a property, we might use a different approach.

dmlloyd:
> wow. This is a long conversation. I don't agree with the suggestion though. I would like to discuss this in more details on a hangout.
> At the moment:
> `my.property=`
> MP Config interpret it exists with an empty value.

Correct.

> In the past we were required to differentiate empty value with nulls. Now, some of you suggested it is wrong to differentiate them. I am not convinced.

To turn this around, can you come up with a valid reason to make them be different?  I contend that such reasons either result in usability deficits or else are hypothetical (i.e. not real-world), or else simply don't *necessitate* differentiation.  On the other hand there are definite concrete real-world advantages to keep them the same, as I've outlined above.

How would *you* decide which approach is best, if not by comparing use cases?

> By the way, the spec does not state an empty string means erasing a value. It is DeltaSpike feature not MP Config. When we try to support erasing a property, we might use a different approach.

Correct, I'm proposing changing the spec in a slightly incompatible way, based on our experiences with Quarkus.

> By the way, the issue was raised to clarify the behavior on how to convert to an array.
> For me:
> 
> `my.property=`
> will get an empty array.
> `my.property=,`
> will get an 1-size array with the empty as the first value.

The current implementation gives two empty values in this case AFAICT.

> What is the use case for the full table expansion?

What do you mean by "full table expansion"?

dmlloyd:
Any follow up on this @Emily-Jiang?  Or anyone else?  It would be good to get this resolved one way or another.

Emily-Jiang:
@jmesnil and I met at today's hangout. We made the following suggestion:

- differentiate empty string and non-existence of config property

specify userName= in a config source means that the property exists
`@Inject @ConfigProperty(name=”userName” defaultValue=”blah”) String user; `=> user=
if userName is not in any config sources, it means it does not exist
comply with the current spec behaviour
userName= means
- means empty or null?

- if means empty

`@Inject @ConfigProperty(name=”userName” defaultValue=”blah”) String user;` => user=’’
`@Inject @ConfigProperty(name=”userName” defaultValue=”blah”) String[] user;` => user.length=0

- if means null => leads to confusion

`@Inject @ConfigProperty(name=”userName” defaultValue=”blah”) String user;` => user=null
improve the spec on configsource.getValue() to return ConfigProperty.UnConfiguredValue
`@Inject @ConfigProperty(name=”userName” defaultValue=”blah”) String[] user;` => user=null
should be return back ConfigProperty.UnConfiguredValue => non existence; hmmm

- Conclusion: treat userName= as empty string

userName= => array size =1
userName=a,b => array size of 2
userName=a,b, => array size of 2

dmlloyd:
So how can a user express in their configuration that they wish to undefine a value?  How can a user specify an empty array in their configuration?  Under what circumstances does it even make sense to differentiate in application code between empty and missing values, particularly arrays?

Unless a definitive answer can be given to these questions, your conclusion is in error.

dmlloyd:
Also: why does backwards compatibility matter for undefined values but not for empty arrays?

jmesnil:
> So how can a user express in their configuration that they wish to undefine a value? 

@dmlloyd I just realised that it is already possible to "undefine" a value by using the [org.eclipse.microprofile.config.configproperty.unconfigureddvalue](https://github.com/eclipse/microprofile-config/blob/1519484a82d907969bf5a66795d86e6a994a21db/api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java#L106) value.

When the MicroProfileConfig implementation reads that value from the `ConfigSource`, it will be able to know that the property is "unconfigured" and overrides any lower-ordinal value.

I really think that the spec and api should do a much better job at describing the contract of `ConfigSource#getValue` so that it is clear from implementation which returned values are valid:
* `null` if the config source knows nothing about the property
* `org.eclipse.microprofile.config.configproperty.unconfigureddvalue` if the config source defines a property to "undefine"/"unconfigure" any lower-ordinal value.
* `""`(empty string) is a valid value devoid of any special meaning

> How can a user specify an empty array in their configuration?

He would be able to do so with an empty string.
We have to update the spec to fix the array behaviour so that it is possible to define all size of arrays (from empty to 1-size with empty string to n size).

> Under what circumstances does it even make sense to differentiate in application code between empty and missing values, particularly arrays?

One use case that has been reported to me was prefixes or suffixes. It should be valid to provide an empty prefix/suffix and this is different from a missing value (that would then use defaults).

> why does backwards compatibility matter for undefined values but not for empty arrays?

In either case, if we want to fix the existing inconsistencies, we have to break compatibility for either undefined values or empty arrays.
My own opinion on that is that the current behaviour for string is fine but the array behaviour is not.

However this change impacts users, configsource implementors and other microprofile specs that uses MicroProfile Config to define their configuration.
Maybe we should reach to the MicroProfile Architecture WG to decide what should be fixed (and required breaking changes)

dmlloyd:
> > So how can a user express in their configuration that they wish to undefine a value?
> 
> @dmlloyd I just realised that it is already possible to "undefine" a value by using the [org.eclipse.microprofile.config.configproperty.unconfigureddvalue](https://github.com/eclipse/microprofile-config/blob/1519484a82d907969bf5a66795d86e6a994a21db/api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java#L106) value.

Unfortunately this sentinel value is only recognized in the context of CDI.  Do we really expect users to insert this value into their configuration files?  I'd consider that a significant change.

> When the MicroProfileConfig implementation reads that value from the `ConfigSource`, it will be able to know that the property is "unconfigured" and overrides any lower-ordinal value.
> 
> I really think that the spec and api should do a much better job at describing the contract of `ConfigSource#getValue` so that it is clear from implementation which returned values are valid:
> 
> * `null` if the config source knows nothing about the property
> * `org.eclipse.microprofile.config.configproperty.unconfigureddvalue` if the config source defines a property to "undefine"/"unconfigure" any lower-ordinal value.
> * `""`(empty string) is a valid value devoid of any special meaning
> 
> > How can a user specify an empty array in their configuration?
> 
> He would be able to do so with an empty string.
> We have to update the spec to fix the array behaviour so that it is possible to define all size of arrays (from empty to 1-size with empty string to n size).

It's not really possible to do this in a consistent way.  Certainly not in a way that preserves compatibility.  See the tables above: these cases have not changed.

> > Under what circumstances does it even make sense to differentiate in application code between empty and missing values, particularly arrays?
> 
> One use case that has been reported to me was prefixes or suffixes. It should be valid to provide an empty prefix/suffix and this is different from a missing value (that would then use defaults).

What kind of prefix/suffix are you talking about?

> > why does backwards compatibility matter for undefined values but not for empty arrays?
> 
> In either case, if we want to fix the existing inconsistencies, we have to break compatibility for either undefined values or empty arrays.
> My own opinion on that is that the current behaviour for string is fine but the array behaviour is not.
> 
> However this change impacts users, configsource implementors and other microprofile specs that uses MicroProfile Config to define their configuration.
> Maybe we should reach to the MicroProfile Architecture WG to decide what should be fixed (and required breaking changes)

See the table above.  Can anyone give a real, legitimate reason to treat an empty value differently than a non-present value?  So far the answer is "no" AFAICT.

njr-11:
> See the table above. Can anyone give a real, legitimate reason to treat an empty value differently than a non-present value? So far the answer is "no" AFAICT.

Unless I'm misunderstanding the terminology again (and I suspect that I am because we've had this conversation before), the answer is yes, we have 5 different MP Config properties where we need to disambiguate empty from non-present in the MP Context Propagation spec, which is the scenario that caused this issue to be raised in the first place. See https://github.com/eclipse/microprofile-config/issues/397 where the MicroProfile Context Propagation spec impl has default lists of propagated/cleared/unchanged context types, and the user needs ought to be able to change this default to any array/list they wish, including empty.  When the user specifies an empty value,
```
mp.context.ThreadContext.propagated=
```
they are explicitly asking for no context types to be propagated. This is a different meaning than non-present, which would indicate that they had never set the property at all and therefore the Context Propagation impl should continue propagating the default set of context types.

dmlloyd:
> > See the table above. Can anyone give a real, legitimate reason to treat an empty value differently than a non-present value? So far the answer is "no" AFAICT.
> 
> Unless I'm misunderstanding the terminology again (and I suspect that I am because we've had this conversation before), the answer is yes, we have 5 different MP Config properties where we need to disambiguate empty from non-present in the MP Context Propagation spec, which is the scenario that caused this issue to be raised in the first place.

Right, you've answered the question of "what existing mechanism can be used to differentiate between the default value and an explicit no-value", whereas the question is "show a useful differentiation between an explicit no-value and an explicit empty-value".  Your use case, as we've discussed, is addressed by the proposal, by essentially telling the API "give me the value of this property, with this default value", and the API would give you an empty value if an explicit empty value was specified.

njr-11:
> Right, you've answered the question of "what existing mechanism can be used to differentiate between the default value and an explicit no-value",

On the contrary, there is no existing mechanism that can make a differentiation, but it sounds like you are proposing a mechanism for it here,

> Your use case, as we've discussed, is addressed by the proposal, by essentially telling the API "give me the value of this property, with this default value", and the API would give you an empty value if an explicit empty value was specified.

I would assert that if there wasn't a useful differentiation, then there would be no need for the proposal above, in which case we would simply have empty value mean empty array all of the time.

dmlloyd:
> > Right, you've answered the question of "what existing mechanism can be used to differentiate between the default value and an explicit no-value",
> 
> On the contrary, there is no existing mechanism that can make a differentiation, but it sounds like you are proposing a mechanism for it here,

That's not contrary.  You're using the existing mechanism of null versus empty difference to decide when to manually pull in a default.

> > Your use case, as we've discussed, is addressed by the proposal, by essentially telling the API "give me the value of this property, with this default value", and the API would give you an empty value if an explicit empty value was specified.
> 
> I would assert that if there wasn't a useful differentiation, then there would be no need for the proposal above, in which case we would simply have empty value mean empty array all of the time.

You can't just "simply" do that because then you end up in odd situations such as having `,` expand to a one-element array (how does that work?) and other such weirdness.  It's not a simple problem and there's no "simply" sort of answer for it.

The crux of the opposition to the proposal boils down to backwards compatibility - which AFAIK has literally never mattered in MicroProfile before (in major releases) - and apparent confusion about the problem itself.  The tables above should make it abundantly clear, particularly the use case table which shows what use cases were not previously satisfied and how this approach fixes the issue.  So maybe it's more productive to address that: what use cases from the table are invalid, if any?  What rules from the "Better rules" table are wrong, and why?  Is there disagreement that the proposal meets the use cases from the table?

jmesnil:
It seems we have 2 different perspectives which boils down to the semantic attached to an empty string.
On one side, an empty string means that the property is missing.
On the other side, an empty string is a regular value whose meaning is application specific.

Both approaches have their pros and cons and are already used in some implementation or microprofile specifications.
I don't see a way to be able to cover the 2 use cases at the same time.

Meanwhile, a useful "new" feature (not defined in the spec) is the ability to /unconfigure/ a property (that may come from lower-ordinal config source)  by setting a specific value.
If we use the empty string for that feature, then we can no longer treat it as a regular value from the application.
Another way to provide this feature would be to user another specific value. For example, the https://github.com/eclipse/microprofile-config/blob/master/api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java#L106[ConfigProperty#UNCONFIGURED_VALUE] that is already part of the API (but setting a value to org.eclipse.microprofile.config.configproperty.unconfigureddvalue is really ugly, a shorter value would be preferable).

I see 2 options:

* [ ] treat an empty string as a missing property
   * (+) simple way to unconfigure property§
   * (+) simplifies array conversion and make it consistent
   * (-) forbids application to get the empty string as a regular value
* [ ] treat an empty string as a regular string
  * (+) leaves open the interpretation of the empty string to the application
  * (-) does not solve array conversion that would still require special use case and break changes
  * (-) requires another way to unconfigure properties§

(§) given the confusion in this PR that would require a good documentation to let users know about this feature

All other use cases that David has extensively listed will follow from that decision.

I'm going back and forth between the 2 approaches. I value the ability to let application decides what an empty string instead of enforcing its meaning.
But having a consistent API is better (even if that makes it a bit less intuitive) that one with many edge cases...

dmlloyd:
> If we use the empty string for that feature, then we can no longer treat it as a regular value from the application.

This is exactly why I've been asking if there is a legitimate use case for that behavior.  So far none has been raised (other than using that as a way to manually select a default value, which should be accomplished in other ways).  If it is true that there is no such use case, then that eliminates the one negative of the proposed behavior.

FroMage:
Guys, I'd be totally fine with `mp.context.ThreadContext.propagated=` meaning "use default value" and `mp.context.ThreadContext.propagated={}` meaning "set to empty array"

It's inconsistent between zero-sized arrays and non-zero-sized arrays (because those don't require braces) but it's also much better than the current status-quo (it allows both zero-sized values and resetting to default) and the `org.this.is.impossible.to.sell.unconfigured` sentinel.

You can also get 100% of young programmers behind this scheme by allowing `∅` and :hand: as aliases for `{}`

njr-11:
+1 to having one value for empty and a different value for default.  That is far more intuitive, and will solve the problem that we originally raised about the spec needing to standardize the right way of specifying empty.  My recommendation would be to standardize a new value for default (for example `mp.context.ThreadContext.propagated=<default>`, and define the empty string to mean empty `mp.context.ThreadContext.propagated=` as it is already being used, but I'll leave it to the MP Config experts to decide.

This would have the following advantages:

- Values are more explicit and consistent with the behavior you get.  Empty string makes sense as a value regardless of whether the type is string or array, whereas `{}` is only intuitive for empty array and would be an awkward value for specifying empty string.
- Breaking change in behavior is limited to the single more unlikely case that someone is already setting a value of `<default>` (or whatever is chosen to represent it).  If done the other way, then there are two breaking changes to deal with: one where users are already setting an empty string value, and another where users are already setting a `{}` value.

dmlloyd:
> You can also get 100% of young programmers behind this scheme by allowing `∅` and  as aliases for `{}`

Maybe we need to support 👏 as a separator!

dmlloyd:
> +1 to having one value for empty and a different value for default. That is far more intuitive, and will solve the problem that we originally raised about the spec needing to standardize the right way of specifying empty. My recommendation would be to standardize a new value for default (for example `mp.context.ThreadContext.propagated=<default>`, and define the empty string to mean empty `mp.context.ThreadContext.propagated=` as it is already being used, but I'll leave it to the MP Config experts to decide.

How would one define an array with a single empty element?

njr-11:
> How would one define an array with a single empty element?

Configuring an array with a single empty element doesn't sound all that useful in my opinion, and I'd recommend waiting until someone establishes a good use case for it before standardizing a way to achieve it.  That said, @OndroMih points out a good way it could be specified in https://github.com/eclipse/microprofile-config/issues/397#issuecomment-455560508

dmlloyd:
Let me rephrase: why is an empty string value useful, when an array with one empty element not useful?

dmlloyd:
Rephrasing again: don't bother suggesting implementation approaches before everyone is in agreement with the use cases.  The suggestion that empty string values are useful presumes an implementation choice.  It doesn't invalidate any of the demonstrated use cases.

njr-11:
> Let me rephrase: why is an empty string value useful, when an array with one empty element not useful?

Here's a scenario that comes to mind.  Suppose an application has a bunch of default values for fields in a form that the user might or might not modify as they step through various panels, and those defaults can be overridden by MP Config properties.  If some are String values and you wanted to override to blank one of them out so that no value appears on the form, then being able to set the MP Config property value to empty String would be the natural way of accomplishing that.

I'm sure you could come up with a scenario for wanting an array containing only the empty string value, but I can't easily think of one that isn't more contrived.

That said, this sort of discussion ought to really be happening between experts/committers within the MP Config space.  I'm only here as a user of the MP Config, providing opinions on what seems the most intuitive to me as a more basic user.  If I'm suggesting things that are architecturally inconsistent or unimplementable, feel free to disregard what I say.  The spec that I'm working on will be able to interoperate with whatever solution you choose so long as there is some spec-defined way for the user to unambiguously configure empty array as a value.

FroMage:
One alternative that comes to mind for syntax to reset a property to its default could be:

```
-mp.context.ThreadContext.propagated
```

This would make it an entirely different syntax to setting a property and remove ambiguity.

manovotn:
> One alternative that comes to mind for syntax to reset a property to its default could be:
> 
> ```
> -mp.context.ThreadContext.propagated
> ```
> 
> This would make it an entirely different syntax to setting a property and remove ambiguity.

That's really good idea, I like how self-explanatory that seems.
It would impose a restriction on initial character of keys, but starting your key with `-` sounds dubious anyway, so it should do not harm.

FroMage:
It also is kinda regular with the badly needed include and profile directives:

```
@include=other.properties
%prod.@include=for-prod.properties
```

(I don't mean to start discussing those in this issue: I merely point out that using a prefix for non-definition operations can be extended to other uses).

dmlloyd:
In the end, syntaxes like this will tie in to the config source implementation, which is really a separate question from how the specification handles such values.  So, I'm not opposed to specifying that the properties-backed configuration source could support this kind of syntax to indicate an explicitly empty/missing value, but I do think that such questions should be addressed *after* this proposal is addressed.

Emily-Jiang:
Sorry for the absence from this conversation. Finally caught up with the thread discussion. Here is what I think we ought to solve:
1. We do need to differentiate empty from non-existence: 

- [ ] use keyword in property name (pretty error prune), 

- [ ] use values (might be ok, maybe something like {{NON-EXISTENCE}} meaning not existing, otherwise, there is a value. Empty means empty.) Do we need to invent a way to skip this special value.

- [ ] For arrays, @jmesnil and I talked about this. 
`names=, ` -> array size 1

`names=`-> array size 1
`names=,,` -> array size 2
basically strip off the ending space.
Let's finalise this in tomorrow's hangout. Please join in the discussion.

dmlloyd:
I will definitely be there.  The key thing I will expect that we will discuss is justifying this assertion "we do need to differentiate empty from non-existence", because my belief that this statement is false is the basis for the proposal.

dmlloyd:
Specifically speaking, let's bring in use cases and see if they can be solved using the proposed approach.

Emily-Jiang:
I consulted both Config JSR EG and MicroProfile googlegroup mailing list with the following email:
Recently, on MP config hangout, we cannot reach conclusion on a couple of use cases. I would like to bring them up here to get a wider conversation purely from a user perspective.

1. As an end user, do you need to set an empty value as a config value and wish to get the empty value from your application? If yes, please provide your detailed use case.
2. As an end user, do you need to way to erase a property, though it was specified in a config source? If yes, please provide your detailed use case again.

When you reply, please use a detailed example.

Thanks in advance!

Emily
========================
Responses are as follows:
Dear Emily,

    here are my 50 Cents on this:

    Concerning question 1: yes, I like empty values

    For me it makes a difference if a property is present but empty, or if it is null.

    There are plenty of use cases for empty properties (like custom prefixes or suffixes).

    If a property is null in contrary, I would like to be able to provide a default value myself, which I wouldn't do, if it is empty.


    Concerning question 2: no idea

    It depends on the definition of "configuration". If "configuration" is always deemed to be "read only", than the erasing of a property simply makes no sense.

    I "configuration" to the contrary is something that is "mutable", than erasing must be possible. Nevertheless it gets tricky here, if a configuration item is defined on several sources and some sources are "ready only" (as e.g. environment variables) whereas other are "modifiable" by the application (as e.g. system properties or files). It must than be ensured, that if a simple "= null" (aka removal) of a single property does not lead to a consecutive search on the next invocation, meaning that an explicit "remove" list must be maintained, which increases programmatic complexity.

    I would go for a "read only" API but offer a simple "clone-like" mechanism into a similar data structure that is mutable and which may also be able to handle deletes. But for the 95% of cases the overhead makes imho no sense.


hth, Philip
======================
I think these behaviors depend of ConfigSource implemented and we should keep it open for the implementation decide. Looking in my example here(https://rhuanrocha.net/2018/12/21/microprofile-config-creating-a-custom-configsource/) the config values are immutable (are read from source only once), but I can create a ConfigSource that refresh these values when called. I'm talking about value inside ConfigSource, but if you are talking about the config value inside a class attribute, I think it don't need to way to erase a property.

Regards,
Rhuan Rocha
=====================

Emily-Jiang:
I spent more time to think about this and took into consideration of the above comments and feedback from end users. I am with @jmesnil @struberg on the using empty as a special way to mean the property does not exist has advantages and disadvantages. I like what @jmesnil suggested to add a way for users to reset the property back to its default. I think this is the best.

dmlloyd:
Sorry, I'm confused about what you're proposing because I cannot figure out what suggestion by @jmesnil you are referring to. Are you suggesting that the proposal can be accepted as it currently stands? Or is there some case you would add or change?

Emily-Jiang:
What I mentioned is not in this PR. It was in one of @jmesnil comments.
I discussed this again at JCrete with Dimitris Andreadis, Ivar Grimstad and Werner Keil. The general consensus for the use case of my.users= specified in a config source is:
The retrieved value is empty string
`@Inject @ConfigProperty(name="my.users") String names;` -> empty string

if a different type such as int or Integer is specified, a Conversion Exception should be thrown. 

I think we have discussed this long enough. I suggest to validate this PR to reflect the above.
As for the following:
`@Inject @ConfigProperty(name="my.users") String[] names;` -> array size 1 with empty value in the first element.
I think this will solve the original issue raised, which was to seek clarification for the above statement.
If `my.users=,`
`@Inject @ConfigProperty(name="my.users") String[] names;` -> array size 1 with empty value in the first element.

dmlloyd:
Unfortunately this introduces quite a lot of chaos into the implementation.

Proposed rules:

* If the value is empty, it is missing
* If a list value is empty, it is missing
* Users can remove a lower-priority config value with a new value or missing

Your rules:

* If the value is empty,
   * If the value has a natural empty representation, use that (Strings for example)
   * If not, give either a conversion exception or a missing exception
* If a list contains an empty value,
   * If the value has a natural empty representation, preserve it as a blank spot in the list *unless* that value happens to be the last value in the list, in which case it's left off
   * If not, give either a conversion exception or a missing exception
* Users cannot cause a value present in a lower-priority source to be missing

Can you explain why these rules are better?  It seems like every time we talk, you agree that the proposed rules make sense, yet you keep coming back to these rules which are more complex and yet cover fewer use cases AFAICT.

dmlloyd:
Also you keep talking about consensus, but have you noticed that we have reached many (conflicting) consensuses?  It all depends on which subset of people you have in the group, and whether they understand what is being proposed.  Those opposed have largely assumed that the proposal doesn't solve some use case, whereas the use case chart above would easily have answered these questions.  The way it is continually described consistently has conflated "what" (adding support for overriding values with empty, without sacrificing any known use case) with "how" (the specific treatment of optional and empty string, which changes the way some use cases are executed).  So, imagine my frustration when you say a decision has been reached.

If you are expressing that you've reached a point of frustration because of the length of this process, I can only say that it's a natural result of the combination of a general lack of interest (one way or another) and confusion about what is being proposed.  It is frankly necessary that both of these problems be solved in order to proceed rationally.

OndroMih:
The distinction for 1) only makes sense if the config sources can make the distinction. And if they can do it, they either have to do the conversion themselves or they need to pass this information to upper levels. If we do the conversion without knowing whether the value is set to empty or not set at all, the non-default conversion can be done in a custom converter. 

If we expect that ConfigSources do the conversion, then it's up to each ConfigSource how the conversion is done. Though we can specify what it means when a ConfigSource returns null and "". We can also specify how the **default** ConfigSources treat `my.users=` values, e.g. whether they return null or "".

If we want to cover most cases with a simple solution and allow extending it to all other cases, we should provide all the information to converters so that a custom converter can decide what to do with an empty/null value.

A possible solution:

* specify that when a ConfigSource returns null, it means the value doesn't exist. If it returns "" then the value exists but is empty. Some sources would return null even for empty values, that's up to the config source. I believe there are very few custom sources that would return "" for unspecified values so we can ignore that case.

* then each converter would either receive an empty value or null and could decide what to do with it, regardless of the default behavior we agree on.

If we go with the simpler default (an empty value leads to "missing") as described by @dmlloyd in https://github.com/eclipse/microprofile-config/pull/407#issuecomment-513225352, a simple converter could return empty strings instead:

String convert(String value) {
    if (value == null) {
        return null;
    } else if (value.isEmpty()) {
        return "";
    } else {
        return value;
    }
}

Alternatively, we can specify a special confg prefix which would trigger this behavior automatically even just for specific config keys. E.g. if the app specifies `mp-config-empty-value-to-empy-string.my.users=true` then `my.users=` would result in "", otherwise in null. However, I think this is too complicated to specify and should be left implementation specific. We can just provide a hint for implementations how to provide more flexibility to users in a recommended way.

OndroMih:
So, my question is: regardless of the final default solution, is there something that prevents using a custom converter to switch to the other behavior or even customize it completely?

If not, I prefer the behavior that is simpler to specify. Here I agree with @dmlloyd  (https://github.com/eclipse/microprofile-config/pull/407#issuecomment-513225352) that returning null is easier to specify than returning "" or throwing exceptions for incompatible types.

dmlloyd:
Ondro, there's a lot of confusion in this thread so let me lay it out simply.  The solution you propose is very close to what I have proposed.

This proposal is really about two things: adding the ability to unset properties, and aligning on exactly one representation for empty values for all specification-provided converters.  All other use cases which are supported today would remain possible and supported.

In terms of UX it is proposed to do it this way:
* Users would unset properties by setting them to the empty string
* The value `Optional.empty()` is used to represent all empty values, even for types which do not have a "natural" empty type. 

The following additional behavioral changes are necessary:
* In arrays and lists, empty initial and middle values will be left out.  Today, AFAIK only empty trailing values are left out.
* All config sources would return `""` for present-but-empty and `null` for missing like you say.
* All spec-provided converters would return `null` when the input is an empty value.
* User provided converters would, as you suggest, be able to return alternative "empty" values, of course, because ultimately they have control over what they return.

I would also suggest that to ensure a uniform behavior, the converter for a given type should always be interrogated to determine what to do when a value is empty by passing it an empty string.  This would give user-provided converters complete control over what they return.  However many converters today will fail with an `IllegalArgumentException` when presented with an empty string.  This complication would either have to be accepted as a compatibility change, or else implementations should capture this exception when it is thrown in response to an empty string and automatically use a `null`/empty value in this case.

Converters should be specified to never be given a `null` value for any reason.

One thing that is not yet clear is how to apply the new consistent rules to `@Inject`+`@ConfigProperty` configuration injections.  A simple solution would be to require that injected values which can be empty simply should be of type `Optional<X>`.  However this is also more impacting in terms of compatibility.  It is also possible to specify that injected values should use special converters for strings, arrays, sets, and lists which yield their "native" empty representations.  But the disadvantage is that it would then be impossible to automatically require a value for these cases.

njr-11:
> In terms of UX it is proposed to do it this way:
> 
>     * Users would unset properties by setting them to the empty string
>     * The value `Optional.empty()` is used to represent all empty values, even for types which do not have a "natural" empty type.

I agree that looking at this from the perspective of user experience is a good approach to take.  The part of the user experience that stands out as missing here is how the user sets properties to an empty list.  I have a MicroProfile spec that instructs users to specify an empty list as the property value in MP Config when they want to override the implementation's default built-in value with empty.  It is a completely different meaning from 'missing', which would mean to accept the default built-in value.  In current versions of MP Config, this is possible with the empty string value.  The proposal under this issue which seems to be gaining consensus involves repurposing the empty string value, but without any clear way for the user to specify that they really wanted an empty list value.  I suspect the intent here is that when you want empty list, write your own converter that re-interprets the value in that way, but that is an awful user experience.  The user who configures the property and the 'user' which is the code that accepts the configured value are two completely different users. A well-designed solution shouldn't require that they have prearranged knowledge of eachother's interpretation of values to make this work.  Even if that approach were considered acceptable, it involves swapping away one provided function (ability to unset) to get another (set to empty list).  If unset is as important as those behind this issue are claiming it to be, then those ought to be two completely distinct things, so that a user who has the ability to specify an empty list value for one property simultaneously has the ability to unset a different property of the same type.

dmlloyd:
> > In terms of UX it is proposed to do it this way:
> > ```
> > * Users would unset properties by setting them to the empty string
> > * The value `Optional.empty()` is used to represent all empty values, even for types which do not have a "natural" empty type.
> > ```
> 
> I agree that looking at this from the perspective of user experience is a good approach to take. The part of the user experience that stands out as missing here is how the user sets properties to an empty list. I have a MicroProfile spec that instructs users to specify an empty list as the property value in MP Config when they want to override the implementation's default built-in value with empty. It is a completely different meaning from 'missing', which would mean to accept the default built-in value.

I'm going to stop you right there because this is the point where people seem to have confusion.  You've made an implementation assumption that detecting a missing value as empty is how one would know to use a default value, but that is not correct.  Using the config accessor API you should specify what the default value is in the case (either as a string or as the actual target object type) where the default value is something other than empty.  This way you would simply read the property and use the result as-is: if it's empty, that's because the user intended it to be empty.  If it's the default value, either the user didn't specify the property or else the default was used (both cases should be treated with one code path if possible).  If it's some other value, then that value should be used.  No branching is necessary.

In the less common event that the default value is not representable, you would have to use (and test for) a sentinel default value object instead in the implementation.  But I expect this case to be less common than a simple default value.

njr-11:
> You've made an implementation assumption that detecting a missing value as empty is how one would know to use a default value, but that is not correct. Using the config accessor API you should specify what the default value is in the case (either as a string or as the actual target object type) where the default value is something other than empty.

I'm not intending to make any implementation assumptions, especially not the one stated.  Not making implementation assumptions means that we don't know if the user who consumes the value will do so via the ConfigAccessor or other MP Config API methods like [Config.getValue](https://download.eclipse.org/microprofile/microprofile-config-1.3/apidocs/org/eclipse/microprofile/config/Config.html#getValue-java.lang.String-java.lang.Class-) or [Config.getOptionalValue](https://download.eclipse.org/microprofile/microprofile-config-1.3/apidocs/org/eclipse/microprofile/config/Config.html#getOptionalValue-java.lang.String-java.lang.Class-). 

When the user invokes `config.getValue(propName, String[].class)` on an empty-value property, they can only get back one value: either the value from the lower priority config source due to the property being 'unset' in a higher priority one, or they can get back the empty array value representing the empty value that they configured.  (Whatever value they get back I hope will not be implementation-specific and vary across implementations - that would be the worst experience for the user). Similarly, with `config.getOptionalValue(propName, String[].class)` on an empty value property, the user can only be given a single value.  Is it the value from the lower priority config source due to the property being 'unset' in a higher priority one?  Is it the empty array?  Is it `Optional.empty()` - but how could that be consistent with what `config.getValue` returns?

I know this discussion has gone on for a while, and I'm torn between continuing to provide helpful - if that's even the right word - feedback on the user perspective here, or whether to just remain quiet until a solution complete with JavaDoc and everything else is finished, and then assess how it fits into our spec usage in MicroProfile.

dmlloyd:
> > You've made an implementation assumption that detecting a missing value as empty is how one would know to use a default value, but that is not correct. Using the config accessor API you should specify what the default value is in the case (either as a string or as the actual target object type) where the default value is something other than empty.
> 
> I'm not intending to make any implementation assumptions, especially not the one stated. Not making implementation assumptions means that we don't know if the user who consumes the value will do so via the ConfigAccessor or other MP Config API methods like [Config.getValue](https://download.eclipse.org/microprofile/microprofile-config-1.3/apidocs/org/eclipse/microprofile/config/Config.html#getValue-java.lang.String-java.lang.Class-) or [Config.getOptionalValue](https://download.eclipse.org/microprofile/microprofile-config-1.3/apidocs/org/eclipse/microprofile/config/Config.html#getOptionalValue-java.lang.String-java.lang.Class-).

Right, but those would be the wrong methods to call if you have a default value, unless you follow the (recommended) practice of using a low-priority configuration source for your default values.  (The reason this is recommended is due to property expansion: most of our users seem to expect that their expressions can refer to values that have defined defaults, and get the default value if the value isn't defined, but this is only achievable using a configuration source for default values).

> When the user invokes `config.getValue(propName, String[].class)` on an empty-value property, they can only get back one result: either a NoSuchElementException overriding the lower priority config source due to the property being 'unset' in a higher priority one, or they can get back the empty array value representing the empty value that they configured. (Whatever value they get back I hope will not be implementation-specific and vary across implementations - that would be the worst experience for the user).

Agreed that making this dependent on implementation would be a bad user experience.  Users would call this method if empty is not an acceptable value.

> Similarly, with `config.getOptionalValue(propName, String[].class)` on an empty value property, the user can only be given a single result. Is it an `Optional.empty()` overriding the lower priority config source due to the property being 'unset' in a higher priority one? Or is it the empty array? How can we not be trading off one for the other here?

Saying that the user must call one of these two methods *is* in fact assuming an implementation which cannot meet the use cases.  To meet these kinds of use cases, the config accessor API becomes useful.  However I also would propose adding methods to `Config` such as:

```java
<T> T getValueWithDefault(String key, Class<T> type, T defaultValue);
<T> T getValueWithStringDefault(String key, Class<T> type, String defaultValue);
<T> T getValueWithDefault(String key, Converter<T> converter, T defaultValue);
<T> T getValueWithStringDefault(String key, Converter<T> converter, String defaultValue);
<T> Optional<T> getOptionalValueWithDefault(String key, Class<T> type, T defaultValue);
<T> Optional<T> getOptionalValueWithStringDefault(String key, Class<T> type, String defaultValue);
<T> Optional<T> getOptionalValueWithDefault(String key, Converter<T> converter, T defaultValue);
<T> Optional<T> getOptionalValueWithStringDefault(String key, Converter<T> converter, String defaultValue);
```

Note that by allowing the converter to provide a custom empty behavior, all of the `getOptional*` methods can be implemented using a single generic `Optional` converter.

> I know this discussion has gone on for a while, and I'm torn between continuing to provide helpful - if that's even the right word - feedback on the user perspective here, or whether to just remain quiet until a solution complete with JavaDoc and everything else is finished, and then assess how it fits into our spec usage in MicroProfile.

It's always worth the one-time investment to get things right IMO.  However I will say that I'm on vacation right now so my responses may be slow.

dmlloyd:
This has been superceded by the issue #446 and two options #444 and #445.  @jmesnil would you please close this PR?

PR #412, opened 2019-02-26
Add ConfigurableConfigSource.class to tck archive

tevans78:

ConfigurableConfigSource was referenced via the ConfigSource Service Loader but the class was missing from the archive.


genie-microprofile:

Issue #393, opened 2018-12-14
Default ConfigSources section list seems to be misplaced


xstefank:
In the [default ConfigSources section](https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc#default-configsources) list of three default property locations it seems that bullet point for `META-INF/microprofile-config.properties` is misplaced.

jmesnil:
fixed by #435

PR #435, opened 2019-08-02
[#380] Fix default ConfigSources list

### LINKED ISSUE #380: Text repeated in configsource spec doc
In the section about environment variable mapping there is the following repeated text:

This ConfigSource searches 3 environment variables for a given property name (e.g. com.ACME.size): This ConfigSource searches 3 environment variables for a given property name (e.g. com.ACME.size):

jmesnil:
List all 3 default ConfigSource first and then add a subsection about
the mapping rule specific to the Env variables.

This fixes #380.


jmesnil:
this also fixes #393

Issue #442, opened 2019-09-06
Get Config master branch back to Config 1.3 


Emily-Jiang:
We discussed today and agreed that we want to revert back to Config 1.3 and then add features gradually.

jmesnil:
I'll create a PR to revert master branch to 1.3.

I'll also look at the PRs that have been merged since 1.3 and cherry-picked the ones that are additional TCKs tests / spec clarification etc.

The idea behind this is to go back to a releasable state for MicroProfile Config 1.x without losing any work that has been done to clarify the existing spec and API

dmlloyd:
This issue can be closed now I think?

PR #443, opened 2019-09-06
Revert to 1.3 + cherry-picks

jmesnil:
As discussed in the meeting today, we agreed to revert to 1.3 in order to have a stable releasable master branch and add new feature one at a time with appropriate use cases / tests / spec.

I did not want to lose any changes we did after 1.3 that consolidates the API and clarified the spec so what I did was
1. Revert to 884c1d30f3c7a13440e981ad00556833574fba8c (the maven commit to prepare 1.4-SNAPSHOT) just after 1.3 tag[1]
2. cherry-pick any commit after 1.3 that clarified the spec / API without introducing any functional changes.
3. I added a commit at the end to bring back some stuff from the reverted commits that were still interesting to keep (javadoc changes, better sentences in the spec)

In addition to the diff from master, another diff to look at is the [diff to 1.3](https://github.com/eclipse/microprofile-config/compare/1.3...jmesnil:revert_to_1.3) which shows what will actually be going in the new release of MicroProfile Config

[1] I used these commands to do this revert commit:

```
git reset --hard 884c1d30f3c7a13440e981ad00556833574fba8c
git reset --soft HEAD@{1}
git commit
```


jmesnil:
@Emily-Jiang the ECA tool is complaining about missing sign-offs but I only cherry-picked from commits that were already in the project

Emily-Jiang:
@jmesnil You need to fix your ECA. Have you signed the new version of agreement?

Emily-Jiang:
I had a 2nd thought. How best to preserve the ConfigAccessor changes with the original contribution as it is very likely we want to add back in? Will it be easily just deleting ConfigSnapshot and cacheFor? Thoughts?

dmlloyd:
> I had a 2nd thought. How best to preserve the ConfigAccessor changes with the original contribution as it is very likely we want to add back in? Will it be easily just deleting ConfigSnapshot and cacheFor? Thoughts?

Definitely in a separate PR.    See #445 for one such option.

jmesnil:
@Emily-Jiang The ECA tool has issues with other sign-offs than mine (see https://accounts.eclipse.org/legal/eca/validation/13334).

I think some commits I cherry-picked were directly applied to master and then were not validated by the tool. But all these commits are already in the project history so I see no issue with merging them again

dmlloyd:
This fixes #442.

dmlloyd:
IMO this should be merged ASAP.  @Emily-Jiang WDYT?

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -121,77 +113,20 @@
...
-     * @param configValues the list of {@link ConfigAccessor} to be accessed in an atomic way
-     *
-     * @return a new {@link ConfigSnapshot} which holds the resolved values of all the {@code configValues}.
-     */
-    ConfigSnapshot snapshotFor(ConfigAccessor<?>... configValues);
-
-
-    /**
-     * Return all property names used in any of the underlying {@link ConfigSource ConfigSources}.
+     * Return a collection of property names.
```
Emily-Jiang: I prefer the previous line.

api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java
```diff
@@ -137,25 +136,4 @@
      */
     @Nonbinding
     String defaultValue() default UNCONFIGURED_VALUE;
-
```
Emily-Jiang: In this file, the contributor Tomas should be removed as well.

Emily-Jiang:
In the error report, it says: Anonymous (jmesnil@****.com) is not covered by the necessary legal agreements in order to proceed. Anyway, can you look at your branch and remove the contributors that committed to the deleted code?

jmesnil:
@Emily-Jiang I fixed the contributors list.

Re the ECA tool, the failure for "Anonymous (jmesnil@****.com)" is caused by a direct commit I did by mistake on GitHub UI (I thought that it'd create a PR but merged to master directly instead): https://github.com/eclipse/microprofile-config/commit/3c04b3a430543643d0ef2df0b0dde8cbad17b980

Likewise, the error for you is related to https://github.com/eclipse/microprofile-config/commit/624426937d1b1f7a5704115024cb411e9cc05b73 that was not signed-off

(we should rely configure GitHub to prevent any direct commit to master to avoid these errors)

jmesnil:
> I had a 2nd thought. How best to preserve the ConfigAccessor changes with the original contribution as it is very likely we want to add back in? Will it be easily just deleting ConfigSnapshot and cacheFor? Thoughts?

There is also variableExpansion and attribute change support, etc. that are still not specified/test properly.

Going back to 1.3 + clarifications allow us to move forward again by adding feature one at a time with a clear understanding and agreement on them.

Keeping master and deleting bits here and there would add entropy to the current state that is not stable enough.
My intent with that PR is to be stable (and releasable again) so that new features can be then added properly

dmlloyd:
As I've said in other contexts, the `ConfigAccessor` I present as an option is almost completely different to the one that was originally in the `master` branch.

I am however OK with the idea of bringing over the original copyright and authors tags.

Emily-Jiang:
As today's hangout, we agreed to merge this PR. The ECA validation errors displayed were the removed contributions.

PR #444, opened 2019-09-06
Empty values option 1

### LINKED ISSUE #446: Consistent empty/missing behavior

# Use cases

### Force empty values

### Consistent handling

### Natural representation

### Default values

# Tasks

# For more information

dmlloyd:
This is the first option for an API to support the new consistent rules for empty value handling, by adding method overloads to the `Config` interface.  Supercedes #407.  Fixes #446.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -34,6 +34,7 @@ package org.eclipse.microprofile.config;
 import java.util.Optional;
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.Converter;
 
 /**
  * <p>
@@ -44,23 +45,28 @@ import org.eclipse.microprofile.config.spi.ConfigSource;
  * <p>If multiple {@link ConfigSource ConfigSources} are specified with
  * the same ordinal, the {@link ConfigSource#getName()} will be used for sorting.
  * <p>
- * The config objects produced via the injection model <pre>@Inject Config</pre> are guaranteed to be serializable, while
+ * The config objects produced via the injection model {@code @Inject Config} are guaranteed to be serializable, while
  * the programmatically created ones are not required to be serializable.
  * <p>
- * If one or more converters are registered for a class of a requested value then one of the registered converters 
- * which has the highest priority is used to convert the string value retrieved from config sources. 
+ * If one or more converters are registered for a class of a requested value, then the registered converter
+ * which has the highest priority is used to convert the string value retrieved from config sources.
  * The highest priority means the highest priority number.
- * For more information about converters, see {@link org.eclipse.microprofile.config.spi.Converter}
+ * For more information about converters, see {@link org.eclipse.microprofile.config.spi.Converter}.
+ * <p>
+ * Alternatively, an explicit converter can be given.  The converter might be independent or it might wrap a converter
+ * returned from the {@link #getConverter(Class)} method.
  *
  * <h3>Usage</h3>
  *
  * For accessing the config you can use the {@link ConfigProvider}:
  *
  * <pre>
- * public void doSomething(
+ * public void doSomething() {
  *   Config cfg = ConfigProvider.getConfig();
  *   String archiveUrl = cfg.getString("my.project.archive.endpoint", String.class);
  *   Integer archivePort = cfg.getValue("my.project.archive.port", Integer.class);
+ *   // ...
+ * }
  * </pre>
  *
  * <p>It is also possible to inject the Config if a DI container is available:
@@ -72,7 +78,14 @@ import org.eclipse.microprofile.config.spi.ConfigSource;
  * }
  * </pre>
  *
- * <p>See {@link #getValue(String, Class)} and {@link #getOptionalValue(String, Class)} for accessing a configured value.
+ * <p>
+ * The {@code getValue(*)} and {@code getOptionalValue(*)} methods can be used for simple access of configuration values.
+ * <p>
+ * To access a configuration value providing a default, the {@code getValueWithDefault(*)} and {@code getOptionalValueWithDefault(*)}
+ * methods allow a typed default to be specified, whereas the {@code getValueWithDefaultString(*)} and {@code getOptionalValueWithDefaultString(*)}
+ * methods allow an unconverted default value string to be given.  If the value is not found in the configuration, then
+ * instead of returning an empty value or throwing {@code NoSuchElementException}, these methods will return the
+ * default value given on the method call.
  *
  * <p>Configured values can also be accessed via injection.
  * See {@link org.eclipse.microprofile.config.inject.ConfigProperty} for more information.
@@ -82,7 +95,7 @@ import org.eclipse.microprofile.config.spi.ConfigSource;
  * @author <a href="mailto:rsmeral@apache.org">Ron Smeral</a>
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  * @author <a href="mailto:gunnar@hibernate.org">Gunnar Morling</a>
- *
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
 @org.osgi.annotation.versioning.ProviderType
 public interface Config {
@@ -91,20 +104,105 @@ public interface Config {
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@link ConfigSource ConfigSources}.
      * <p>
-     * If this method gets used very often then consider to locally store the configured value.
+     * If this method gets used very often, consider locally storing the configured value.
      *
-     * @param <T>
-     *             The property type
-     * @param propertyName
-     *             The configuration propertyName.
-     * @param propertyType
-     *             The type into which the resolve property value should get converted
-     * @return the resolved property value as an object of the requested type.
-     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type.
+     * @param <T> the result type
+     * @param propertyName the configuration property name (must not be {@code null})
+     * @param propertyType the type into which the resolved property value should be converted (must not be {@code null})
+     * @return the resolved property value as an object of the requested type
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
      * @throws java.util.NoSuchElementException if the property isn't present in the configuration, or is present
-     *  but has an empty value.
+     *  but has an empty value
      */
     <T> T getValue(String propertyName, Class<T> propertyType);
+
+    /**
+     * Return the resolved property value with the specified type for the
+     * specified property name from the underlying {@link ConfigSource ConfigSources}, using the given
+     * default value if the property is not found.
+     * <p>
+     * If this method gets used very often, consider locally storing the configured value.
+     *
+     * @param <T> the result type
+     * @param propertyName the configuration property name (must not be {@code null})
+     * @param propertyType the type into which the resolved property value should be converted (must not be {@code null})
+     * @param defaultValue the default value to return if no property with the given name is found
+     * @return the resolved property value as an object of the requested type
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws java.util.NoSuchElementException if the property has an empty value
+     */
+    <T> T getValueWithDefault(String propertyName, Class<T> propertyType, T defaultValue);
+
+    /**
+     * Return the resolved property value with the specified type for the
+     * specified property name from the underlying {@link ConfigSource ConfigSources}, using the given
+     * default value if the property is not found.  The default value will be converted into the target type
+     * only if the property is not found.
+     * <p>
+     * If this method gets used very often, consider locally storing the configured value.
+     *
+     * @param <T> the result type
+     * @param propertyName the configuration property name (must not be {@code null})
+     * @param propertyType the type into which the resolved property value should be converted (must not be {@code null})
+     * @param defaultValue the default value to convert and return if no property with the given name is found (must not be {@code null})
+     * @return the resolved property value as an object of the requested type
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws java.util.NoSuchElementException if the property has an empty value, or the property is not present
+     *      and the default value string is empty
+     */
+    <T> T getValueWithStringDefault(String propertyName, Class<T> propertyType, String defaultValue);
+
+    /**
+     * Return the resolved property value with the specified converter for the
+     * specified property name from the underlying {@link ConfigSource ConfigSources}.
+     * <p>
+     * If this method gets used very often, consider locally storing the configured value.
+     *
+     * @param <T> the result type
+     * @param propertyName the configuration property name (must not be {@code null})
+     * @param converter the converter to use to convert the value (must not be {@code null})
+     * @return the resolved property value as converted by the given converter
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws java.util.NoSuchElementException if the property isn't present in the configuration, or is present
+     *  but has an empty value
+     */
+    <T> T getValue(String propertyName, Converter<T> converter);
+
+    /**
+     * Return the resolved property value with the specified converter for the
+     * specified property name from the underlying {@link ConfigSource ConfigSources}, using the given
+     * default value if the property is not found.
+     * <p>
+     * If this method gets used very often, consider locally storing the configured value.
+     *
+     * @param <T> the result type
+     * @param propertyName the configuration property name (must not be {@code null})
+     * @param converter the converter to use to convert the value (must not be {@code null})
+     * @param defaultValue the default value to return if no property with the given name is found
+     * @return the resolved property value as converted by the given converter
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws java.util.NoSuchElementException if the property has an empty value
+     */
+    <T> T getValueWithDefault(String propertyName, Converter<T> converter, T defaultValue);
+
+    /**
+     * Return the resolved property value with the specified converter for the
+     * specified property name from the underlying {@link ConfigSource ConfigSources}, using the given
+     * default value if the property is not found.  The default value will be converted into the target type
+     * only if the property is not found.
+     * <p>
+     * If this method gets used very often, consider locally storing the configured value.
+     *
+     * @param <T> the result type
+     * @param propertyName the configuration property name (must not be {@code null})
+     * @param converter the converter to use to convert the value (must not be {@code null})
+     * @param defaultValue the default value to convert and return if no property with the given name is found (must not be {@code null})
+     * @return the resolved property value as converted by the given converter
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws java.util.NoSuchElementException if the property has an empty value, or the property is not present
+     *      and the default value string is empty
+     */
+    <T> T getValueWithStringDefault(String propertyName, Converter<T> converter, String defaultValue);
 
     /**
      * Return the resolved property value with the specified type for the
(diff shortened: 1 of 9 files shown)


genie-microprofile:

genie-microprofile:

dmlloyd:
Before moving the PR out of draft:

- [ ] Choose this approach and close #445 
- [ ] Ensure that a conforming implementation is possible; revise as needed
- [ ] Complete and test TCK additions necessary to test conformance
- [ ] Rebase after #443 or equivalent is merged

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -92,3 +92,50 @@ If no built-in nor custom `Converter` for a requested Type `T`, an implicit Conv
...
+| `"foo,bar"` | `String[]` | `getValue` | `{ "foo", "bar" }`
+| `"foo,"`    | `String[]` | `getValue` | `{ "foo" }`
+| `",bar"`    | `String[]` | `getValue` | `{ "bar" }`
+| `",bar,"`   | `String[]` | `getValue` | `{ "bar" }`
+| missing     | `String` |`getOptionalValue` | `Optional.empty()`
+| `""`        | `String` | `getOptionalValue` | `Optional.empty()`
+| `" "`        | `String` | `getOptionalValue` | `Optional.of(" ")`
+| `"foo"`     | `String` | `getOptionalValue` | `Optional.of("foo")`
+| missing | `String[]` | `getOptionalValue` | `Optional.empty()`
+| `""` | `String[]` | `getOptionalValue` | `Optional.empty()`
```
njr-11: This section of documentation seems contradictory.  The section begins by stating "A config property can be explicitly configured with an empty value (the empty String `""`)", which is something I'm in favor of.  But it later goes on to state that the returned Optional then ends up without a value: `Optional.empty()`.  Per the documentation of java.util.Optional, this clearly indicates missing, rather than a value that is an empty String/String[]/whatever.  This makes usage of the Optional interface counter-intuitive.  Consider what will happen when MicroProfile users write code per the intended meaning of java.util.Optional that is defined within java.util.Optional's own JavaDoc. When users write,
```
   optional.ifPresent(value -> consumeValue(value));
```
they expect the Consumer to run with the value that was configured: empty value or otherwise.  However, that won't happen if MP Config halfway repurposes Optional.empty, in which case it would behave as if there were no value present at all - not an empty value.

Similarly, when users write,
```
   propValue = optional.orElse(new String[] { DEFAULT_ITEM1, DEFAULT_ITEM2 });
```
they expect to actually get an empty String array value assigned if their user configured to have an empty string array.  However, if MP Config repurposes `Optional.empty()` as proposed here then they'll end up with the supplied parameter instead.

In the JavaDOC API for java.util.Optional, `Optional.empty()` very clearly indicates missing, both in how it is documented (stating that ["No value is present for this Optional"](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#empty--)), as well as in the behavior of the various operations that are available on java.util.Optional.  I strongly caution against altering its meaning when used within MP Config.  Doing so would stray from all other expected usage and leave MicroProfile with a confusing and difficult-to-use spec that is inconsistent with Java and will be the source of numerous bugs.

This basically gets at the underlying problem that I've had trouble with all along in the earlier discussions.  Missing values and values for instances that are empty are fundamentally two distinct things and cannot share a single representation without causing inconsistency and a source of confusion/error.   An Optional that is empty (has no value) is different than an Optional that has a value which is an empty instance of something: the empty string, the empty set, an empty array or list, ...

dmlloyd: > This section of documentation seems contradictory. The section begins by stating "A config property can be explicitly configured with an empty value (the empty String `""`)", which is something I'm in favor of. But it later goes on to state that the returned Optional then ends up without a value: `Optional.empty()`. Per the documentation of java.util.Optional, this clearly indicates missing, rather than a value that is an empty String/String[]/whatever.

The method is literally called `empty()`, which to me pretty clearly means "empty".  So we might have to agree to disagree on this one.

> This makes usage of the Optional interface counter-intuitive. Consider what will happen when MicroProfile users write code per the intended meaning of java.util.Optional that is defined within java.util.Optional's own JavaDoc. When users write,
> 
> ```
>    optional.ifPresent(value -> consumeValue(value));
> ```
> 
> they expect the Consumer to run with the value that was configured: empty value or otherwise.

I don't know that this would be intuitive.  Consider for example the case where `value` is not a `String`, but rather, say, a `Double` or a `LocalDate` or something.

> Similarly, when users write,
> 
> ```
>    propValue = optional.orElse(new String[] { DEFAULT_ITEM1, DEFAULT_ITEM2 });
> ```
> 
> they expect to actually get an empty String array value assigned if their user configured to have an empty string array. However, if MP Config repurposes `Optional.empty()` as proposed here then they'll end up with the supplied parameter instead.

That's what default value is for.  I don't think it's that confusing to say "if you have a default value, use the default value method".

> In the JavaDOC API for java.util.Optional, `Optional.empty()` very clearly indicates missing, both in how it is documented (stating that ["No value is present for this Optional"](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#empty--)), as well as in the behavior of the various operations that are available on java.util.Optional. I strongly caution against altering its meaning when used within MP Config.

We will now define an empty input string as being "explicitly missing", which is completely within the documented definition of the method.  Again you've latched on to the behavior of `String`, which has its own natural empty representation.  Most types do not.  If I give a property value like `my.int.value=`, I am clearly and unambiguously saying "this value is now explicitly cleared by my request".  Whereas by specifying a default value, I am clearly saying "if no value is given for this property, it has this value", where the default value will generally be non-empty.  So in this case it is not missing at all.

Once we look at property expansion, which is likely to immediately follow this issue, the traditional idea of "missing" becomes even more nonsensical.  When a property is expanded, it may reference other properties which in turn may be expanded, and so on.  These properties might not be present in the traditional sense, but they may be associated with default values.  The only way to accomplish this is by having a bottom-priority config source which contains all the default values for properties which have them.  Once you do this however, then the idea of detecting "missing" by `Optional` no longer works, no matter how intuitive or unintuitive it may appear to you.  It simply ceases to be an option, because any "missing" value will appear to be "present" by the `Optional` API due to the default value resolution.  The proposed specification change, however, integrates flawlessly with this case.

In any case this is the only viable specification option that has been found or that seems possible for fixing the behavioral inconsistencies without introducing other inconsistencies or blocking the way to property expansion.  An alternative proposal would be welcome, if it satisfies all the same cases.  If you open such a PR, reference this issue.  However, I simply do not believe it is possible to solve the underlying issues without a common representation of empty values, and distinct means for handling empty and missing.  And doing this will, as far as I can tell, inevitably require that empty and missing-with-no-default must necessarily be represented by the same value; in other words, you'll end up in the same place once you have accounted for all the hard constraints.

> Doing so would stray from all other expected usage and leave MicroProfile with a confusing and difficult-to-use spec that is inconsistent with Java and will be the source of numerous bugs.

That is just pure hyperbole, which I completely disagree with, so I'm going to leave that aside.

> This basically gets at the underlying problem that I've had trouble with all along in the earlier discussions. Missing values and values for instances that are empty are fundamentally two distinct things and cannot share a single representation without causing inconsistency and a source of confusion/error. An Optional that is empty (has no value) is different than an Optional that has a value which is an empty instance of something: the empty string, the empty set, an empty array or list, ...

Right, this is the essential problem of having multiple empty representations: every converter ends up with different behavior.  In addition, of course, some types have no empty representation, so sometimes empty values will fail and sometimes they will not, making it impossible to use an empty input string as a way to unset a property, which is the only intuitive solution to that problem.  So, we unify on exactly one representation of empty.  The examples should be fairly easy to follow, so I really don't anticipate much justifiable confusion.

I'm sorry that this specification change does not meet your liking.  However, it cannot be denied that the change does work with all of the existing and suggested use cases as well as making it possible to support property expansion.  In this case I think it is important to design for correctness first.  If there is some possible change that might improve usability in your eyes, without sacrificing correctness, then I'm eager to hear about it.  But otherwise, we have to accept this change and move forward.

As such I will not accept or respond to any more subjective or hyperbolic assessments of the change.  Any objective ideas for usability enhancements can be proposed as pull requests to one or the other of the branches on these pull requests (https://github.com/dmlloyd/microprofile-config/tree/empty-values-option-1 or https://github.com/dmlloyd/microprofile-config/tree/empty-values-option-2) and they will be considered fairly as long as they do not cause a regression that invalidates any of the preestablished use cases or behaviors.

njr-11: > The method is literally called `empty()`, which to me pretty clearly means "empty". So we might have to agree to disagree on this one.

Neither of us gets to decide for ourselves what [Optional.empty()](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#empty--) means.  This isn't a matter of opinion that we can agree to disagree on.  java.util.Optional is a standard, it is JavaDoc'ed, and it is already implemented with a behavior that won't change based on any arguments we make in this pull.

> If there is some possible change that might improve usability in your eyes, without sacrificing correctness, then I'm eager to hear about it. But otherwise, we have to accept this change and move forward.
> 
> As such I will not accept or respond to any more subjective or hyperbolic assessments of the change. Any objective ideas for usability enhancements can be proposed as pull requests to one or the other of the branches on these pull requests

I realize, as someone who lacks any experience in the implementation of MicroProfile Config, that my comments are only suggestions.  It sounds like acceptance of this pull, or the other one, with the same identical section, is already a foregone conclusion.  To that end, the best I can do, having conceded that the MP config spec will be changing behavior, is to help you align the language in the documentation with what these changes are actually doing.  I would rather see the MicroProfile Config spec be meaningful and well understood in the behavior that you actually get from it, rather than for it to portray that it is giving me a behavior which I requested but am not actually getting.

To accomplish that, please rewrite this section, previously titled "Empty value", to instead be:

```
=== Missing value

A config property can be explicitly configured to indicate that a property is missing by specifying the empty String (`""`).  The property
is considered to be missing if no default value exists for that property.

The `Config.getValue()` method will normally throw a `NoSuchElementException` when a configuration value
is missing.  Empty values (such as the empty string `""` or a zero-length array) cannot be returned
unless a custom converter is specified which provides a specific value for
the case where an empty value is desired.

[NOTE]
This behaviour allows users to _unconfigure_ a property. If a property is configured in a lower-ordinal ConfigSource with
an undesirable value, you can set the property value to an empty string, which means missing, in a higher-ordinal ConfigSource to discard the lower-ordinal value.

...
```

The remaining portion of this section can then remain as is.

The MicroProfile spec that I made the original request for empty array on behalf of, will then need to work around this change in behavior by defining its own constant for users to specify when they want empty array.

dmlloyd: > > The method is literally called `empty()`, which to me pretty clearly means "empty". So we might have to agree to disagree on this one.
> 
> Neither of us gets to decide for ourselves what [Optional.empty()](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#empty--) means. This isn't a matter of opinion that we can agree to disagree on. java.util.Optional is a standard, it is JavaDoc'ed, and it is already implemented with a behavior that won't change based on any arguments we make in this pull.

Well, surely we agree that the word "empty" means "empty", regardless of subjective context.  There otherwise is no objective relationship that I can discern between `Optional`'s documented behavior and the use to which it is put here: any use of the word "empty" in the `Optional` documentation seems to refer to the `Optional` itself, which is defined only as "A container object which may or may not contain a non-`null` value".  In fact the word "missing" never actually seems to appear in the `Optional` JavaDoc, so I'm not quite sure what you're referring to here.

> > If there is some possible change that might improve usability in your eyes, without sacrificing correctness, then I'm eager to hear about it. But otherwise, we have to accept this change and move forward.
> > As such I will not accept or respond to any more subjective or hyperbolic assessments of the change. Any objective ideas for usability enhancements can be proposed as pull requests to one or the other of the branches on these pull requests
> 
> I realize, as someone who lacks any experience in the implementation of MicroProfile Config, that my comments are only suggestions. It sounds like acceptance of this pull, or the other one, with the same identical section, is already a foregone conclusion. To that end, the best I can do, having conceded that the MP config spec will be changing behavior, is to help you align the language in the documentation with what these changes are actually doing. I would rather see the MicroProfile Config spec be meaningful and well understood in the behavior that you actually get from it, rather than for it to portray that it is giving me a behavior which I requested but am not actually getting.
> 
> To accomplish that, please rewrite this section, previously titled "Empty value", to instead be:

Sure, this change seems quite reasonable.  I'll make the changes.

> The MicroProfile spec that I made the original request for empty array on behalf of, will then need to work around this change in behavior by defining its own constant for users to specify when they want empty array.

That is not true.  If you have a look at the description of #446, you should be able to extrapolate how to accomplish this use case (n.b. I'm understanding the use case to be "have a non-empty default value with a special meaning which might not be otherwise representable as a string value") without requiring a special constant for users to specify.  The point of the proposed solution is to be able to meet use cases like this one without requiring the user to go through any unusual acrobatics.

Of course you're also free to have a special constant, if you feel that gives a better user experience for that particular case.  In particular it allows the user to override a lower priority value with a value representing "whatever the default is", which is not otherwise possible.  But it is definitely not required otherwise.

Note that having a representable default value - even if the literal string isn't known until the program runs - will allow the value to participate in property expansion, which otherwise it would not be able to do.

dmlloyd: I've made the changes, and I marked you (@njr-11) as co-author using the email address you've used on other MicroProfile commits.  Please let me know if this is OK.

njr-11: > That is not true. If you have a look at the description of #446, you should be able to extrapolate how to accomplish this use case (n.b. I'm understanding the use case to be "have a non-empty default value with a special meaning which might not be otherwise representable as a string value") without requiring a special constant for users to specify. The point of the proposed solution is to be able to meet use cases like this one without requiring the user to go through any unusual acrobatics.

Here is what I see in #446 for this scenario,

> // approach 1 (using implementation option 1)
String value = config.getOptionalValueWithDefault("prop6.not.specified", String.class, "default-value").orElse("");
// approach 2 (using implementation option 2)
String value = config.getAccessor("prop6.not.specified")
        .convertedForType(String.class)
        .withDefault("default-value")
        .optional()
        .getValue()
        .orElse("");

In the case of approach 1, the getOptionalValueWithDefault method can never return Optional.empty() because "default-value" is not null.  (See the JavaDoc for getOptionalValueWithDefault).  That means `orElse` is a no-op here.

In the case of approach 2, the Optional that is returned by `ConfigAccessor<Optional<String>>.getValue()` similarly either has a value of "default-value" or the configured value (absent any default) and so the `orElse` is again a no-op.

To get this working properly, getOptionalValueWithDefault [1] and getValue [2] would need to behave differently and instead return Optional.empty(), ignoring the non-null default, but that would bust the case where the user specified "" to unconfigure a lower-order config source and let the value default.

This goes back to the problem that "" cannot mean 2 different things at once.  With MP Config now claiming "" for unconfiguring/missing value, it would simply be wrong for another MicroProfile spec to claim that it must mean something else.  That's why the new constant will be needed.

dmlloyd: > > That is not true. If you have a look at the description of #446, you should be able to extrapolate how to accomplish this use case (n.b. I'm understanding the use case to be "have a non-empty default value with a special meaning which might not be otherwise representable as a string value") without requiring a special constant for users to specify. The point of the proposed solution is to be able to meet use cases like this one without requiring the user to go through any unusual acrobatics.
> 
> Here is what I see in #446 for this scenario,
> 
> > // approach 1 (using implementation option 1)
> > String value = config.getOptionalValueWithDefault("prop6.not.specified", String.class, "default-value").orElse("");
> > // approach 2 (using implementation option 2)
> > String value = config.getAccessor("prop6.not.specified")
> > .convertedForType(String.class)
> > .withDefault("default-value")
> > .optional()
> > .getValue()
> > .orElse("");
> 
> In the case of approach 1, the getOptionalValueWithDefault method can never return Optional.empty() because "default-value" is not null. (See the JavaDoc for getOptionalValueWithDefault). That means `orElse` is a no-op here.

That's not correct.  The default value is returned if no configuration source specifies a that configuration key.  The empty optional is returned if the configuration source with the highest priority for a given key explicitly specifies an empty value for that key (e.g. `keyname=`).  That's what this sentence means: "The return value will be empty if the value is empty or if the property is not found and the default value given is `null`".  If the default value is not null, then the other behavior applies: "...using the given default value if the property is not found".  Specifying an empty value in this case does not cause the value to be "not found", rather it is found to have an empty value, rather than whatever the default value is.

It all works, honest!

> In the case of approach 2, the Optional that is returned by `ConfigAccessor<Optional<String>>.getValue()` similarly either has a value of "default-value" or the configured value (absent any default) and so the `orElse` is again a no-op.

This is the same as above: if the key is not specified, the default value is used; otherwise, if the value is empty, the empty optional is returned.

Does that make sense?

njr-11: > Specifying an empty value in this case does not cause the value to be "not found", rather it is found to have an empty value, rather than whatever the default value is.

Ok - I was missing that detail about "found to have an empty value" vs "not found".  I still have trouble grasping how it can on the one hand be found to have a value, but on the other hand, that value isn't the value that is considered present within the Optional.  Are you certain users will be able to understand the minute details needed to properly use these methods?

dmlloyd: I would have thought so.  But, any ideas as to how to further clarify the language would be welcome.  I'll spend some time today trying to improve the verbiage if I can.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -82,55 +95,245 @@
...
+     *
+     * @param <T> the result type
+     * @param propertyName the configuration property name (must not be {@code null})
+     * @param propertyType the type into which the resolved property value should be converted (must not be {@code null})
+     * @param defaultValue the default value to return if no property with the given name is found
+     * @return the resolved property value as an object of the requested type
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws java.util.NoSuchElementException if the property has an empty value
+     */
+    <T> T getValueWithDefault(String propertyName, Class<T> propertyType, T defaultValue);
```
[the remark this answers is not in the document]
dmlloyd: It cannot be removed.  `NoSuchElementException` does not mean not-present, it means empty.  We need some exception to represent this case, and this is the exception used already today (with 1.3 and earlier); changing that would be drastically incompatible.  Whereas with this behavior almost every existing case will continue to work the same.  I recognize it is not ideal, but if one just reads the documentation it should be clear what it means: it's thrown when the property has an empty value.

The property type has to be given because the default value can be any subtype, including null which is a subtype of everything.  It would be possible to extrapolate the type in some cases but then the user might end up with a different converter than the one they were wanting.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -82,55 +95,245 @@
...
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws java.util.NoSuchElementException if the property has an empty value, or the property is not present
+     *      and the default value string is empty
+     */
+    <T> T getValueWithStringDefault(String propertyName, Converter<T> converter, String defaultValue);
+
+    /**
+     * Return the resolved property value with the specified type for the
+     * specified property name from the underlying {@link ConfigSource ConfigSources}.  The
+     * return value will be {@linkplain Optional#empty() empty} if the value is empty.
```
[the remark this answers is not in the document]
dmlloyd: A zero-length string is not the same as empty.  For example, when converting an array, the value `",,,,,"` is empty.  Why?  Multiple reasons.  If I'm doing property expansion (coming soon) and give `"${foo:},${bar:}"` and neither are defined, I want an empty list.  And if `","` yielded an array with two blank values, then `""` logically would yield an array with one blank value.  Not good!  So, empty value segments in lists are empty; a list containing only empty value segments is empty.  A zero length string is not the only empty case.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -82,55 +95,245 @@
...
+     *
+     * @param <T> the result type
+     * @param propertyName the configuration property name (must not be {@code null})
+     * @param propertyType the type into which the resolved property value should be converted (must not be {@code null})
+     * @param defaultValue the default value to return if no property with the given name is found
+     * @return the resolved property value as an {@code Optional} of the requested type (not {@code null})
+     *
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     */
+    <T> Optional<T> getOptionalValueWithDefault(String propertyName, Class<T> propertyType, T defaultValue);
```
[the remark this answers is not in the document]
dmlloyd: If this method is removed, then use cases cannot be met, so it is not an option unless a replacement is given that does meet the lost use cases (your own use case is included in that by the way).

A "default value" is defined as "the value given when the property is not present".  Using this method will yield an empty optional if the value is explicitly empty, or the given default if the value is missing.  Such a method is necessary if a user can be allowed to explicitly remove properties, no matter how you do it.

dmlloyd:
My apologies if I'm misreading you, but it seems as if you still do not understand the semantics of what happens, nor the reasons for it.  The changes you proposed will actively break the implementation and cause the associated use cases to not be achievable.  Is there some possible way I could explain the how & why of what is in here, so that we can make productive forward progress?  Is there some clarification of the documentation I could do to help out?

I understand you have a certain idea of what some of the terminology *should* mean.  But I think that unless you understand the meanings of the terms as they're laid out by the specification, *before* we propose changing the terminology and *before* we propose changing the semantics, I think we're just going to be running in circles.

We have to approach this in an organized fashion if we're ever going to get anywhere.

dmlloyd:
I'm not saying that no change is possible, but I do think we have to be speaking the same language in order to communicate.

api/src/main/java/org/eclipse/microprofile/config/Config.java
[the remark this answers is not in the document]
njr-11: > `NoSuchElementException` does not mean not-present, it means empty. We need some exception to represent this case, and this is the exception used already today (with 1.3 and earlier); changing that would be drastically incompatible.

I was trying to point out that you don't need to raise the NoSuchEelementException from this method at all.  It's a newly introduced method by this pull, so there is no incompatibility with existing usage.  Given that the point of the method is to let the user provide a default value, there should always be a value to return (absent a conversion error, which you already cover with IllegalArgumentException) - either an explicitly configured value that evaluates to something (non-empty) or otherwise the default.  I thought this was the behavior you were wanting to achieve with "" and other expressions that evaluate to no value specified.

What I've been trying to emphasize as an important design point is that a single configuration property value, given a specific type, ought to have a single meaning.   The conclusion that I thought we had reached was that the zero-length string (""), and by extension other values that evaluate to 'no value specified', would have the meaning of unconfigure.  If that is what we are presenting to the user, then the expected behavior for a method to which you supply a default would be to use the default when no value is configured and should be identical to what you get if the value is unconfigured.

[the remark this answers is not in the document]
dmlloyd: But then you re-introduce the problem that you can't override a default value with an empty value.  So if I have a property which, for example, represents a list of contexts to propagate, and has a default value of `"foo,bar"`, but the user wants to explicitly propagate no contexts, the user could specify `""` and reasonably expect no contexts to propagate.  But if, as you propose, either a non-empty value or the default value is returned, the result would be `"foo,bar"`, invalidating the purpose of the change; the user would not be able to specify "no contexts" without a special sentinel value, which doesn't work for all types and thus would require custom converters in some cases but not others.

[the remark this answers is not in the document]
njr-11: Yes, I understand that, that is the exact the scenario that I need to preserve for MicroProfile Context Propagation.  However, I'm conceding that we need to sacrifice that because otherwise MicroProfile Config becomes too complex and doesn't make any sense to the average user.  We will have to deal with it, either by introducing our own constant for empty value/list, or convincing MP Config to standardize one.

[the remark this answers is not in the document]
dmlloyd: I really don't think the change can be done without having a first-order concept of an "empty value", and a first-order concept of a "missing key", and a first-order concept of a "default value", and defining behaviors in those terms:

* An *empty value* is a value that would convert to nothing (for example, imagine a property of type `Double` that was given as an empty string); empty values can be given as explicit values for configuration keys
* A *default value* is a value for a configuration key that implicitly exists in a "virtual" lowest-priority configuration source, i.e., it's only used when no other configuration source gave a value
* A *missing key* is a configuration key that is not specified in any config source (including as a default value); in absence of a *default value*, *missing keys* resolve to *empty values*

Since we don't want to force every type to have an "empty" representation - especially because some simply can't - one might consider to use `null` to mean "empty" for all types; however, regardless of one's feelings about `null`, `getValue` is not expected to return it historically so that would be a major breaking change, and in fact is the exact reason that `getOptionalValue` exists, returning (as it does) its `null`-wrapping payload.

But because of this, `Optional` simply cannot represent "missing", because if it did, what would represent "empty"?  Sure we could add an `Emptiable<T>` and then have an `Optional<Emptiable<T>>` variation, but isn't that worse?  Now the user has to figure out what to do in the empty case, or the missing case, or the missing-and-empty case, and what would `Emptiable<Optional<T>>` mean anyway?  Or `Emptiable<String>`, or `Emptiable<OptionalInt>`?  You'd end up with as many rules as you have permutations of types, values, and converters.

At least the rules as proposed are consistent and clear, if one reads the docs.  They're even intuitive, especially once you consider questions like:

* What do I want to get back if I call `getValue` and pass in a `Converter<OptionalInt>` and the value is missing?  Or explicitly empty?  What if the value defaults to "10" but can also be explicitly cleared?  Shouldn't the result be `OptionalInt.empty()`?  How does that compare to if I pass in a `Converter<Optional<Integer>>`?  Shouldn't they behave the same i.e. get back an `Optional.empty()`?
* Compare to this case: what do I want to get if I call `getValue` with a `Converter<String>` and the value is empty?  An empty string?
* But then... if I call `getValue` with a `Converter<List<String>>`... now I come back to the janky list case where I get a list of blanks (of possibly unexpected size) in some cases.  And then, using the same logic, a `Converter<List<Double>>` gets... an exception!  Why?  Because we don't represent empty specially, so the conversion fails for lists such as `"123.0,${foo:},456.0"` in the absence of a `foo` property, because `Double` has no empty representation and so cannot be elided from the list.

Without the ability to represent empty specially, generic enclosing converters (like those needed to return arrays or lists of things, or optional-ize them, etc.) become impossible because they can never, ever consistently handle empties, because empty handling varies by type.  Therefore we have a special representation of emptiness: at a converter level, this is `null`.  This allows a generic list converter to delegate conversion of its elements, removing empty segments and producing an expected result. But now we have a new problem: what if the list ends up empty?  Should we return an empty list?  No!  Because now the list converter in turn cannot be encapsulated, closing the door on more complex configuration value types.  Instead, the empty list returns `null` as well, as must *every other* converter type that would produce an empty value, allowing converters to nest arbitrarily.

In other words, types that have special empty values cannot be nested within other types.  But `Optional` serves that purpose perfectly: if the value is empty, then you can use `orElse` to use whatever empty representation you like, be it an empty list, empty string, etc.

It all makes sense once you accept the above three definitions.  Then you know that every key has a value - even if it's an *empty value* - and `getOptionalValue` lets you return an empty value but `getValue` does not, returning `NoSuchElementException` in this case.  Since empty values behave identically to pre-1.4 missing values, it makes sense to continue the same semantics moving forward.  In addition, by specifying a custom converter, customized empty values can also be used, though in most cases `Optional` is sufficient.  And by adding a way to specify a use-site default, one can conveniently get a value that is used when the key is *missing*.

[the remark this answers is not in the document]
njr-11: While I wouldn't have defined default value as you do, I can see how doing so is helpful to the solution you are aiming for.  Otherwise, I can follow the logic up until,

> But because of this, `Optional` simply cannot represent "missing", because if it did, what would represent "empty"?

This is where I get lost in the logic, because, as someone who is more familiar with java.util.Optional than MP Config, Optional does have a natural representation of "missing", which is `Optional.empty()`.  From previous comments, I know you don't agree with that and I don't want to start the argument again.  But if you assume for a moment, despite the unfortunate overloading of terminology, that `Optional.empty()` is the natural way of representing "missing/not present", that leaves the question of how does an empty value get represented, which would be "" for String, size 0 array for arrays, and a failure to create the Optional at all if not representable.

We already decided, I wouldn't push that approach any further.  I'll certainly acknowledge that your solution offers more functionality and flexibility, and is very creative in achieving that.  I don't see it as intuitive, though, and have tried to explore some ways of improving on that.  Unfortunately, it seems the conclusion we are reaching is that doing so would start to limit the function you are aiming to achieve, and so isn't acceptable.

At this point, I would highly recommend that this should be reviewed by MP Config contributors/experts, rather than by myself, who is just a user of MP Config.  What you have proposed may very well be the right strategic direction for MP Config to take.  I'm not the right person to make that sort of judgement.  All of the feedback that I have to offer, I have already offered.  You and the other contributors to MP Config should just make the decision.

Emily-Jiang:
I think this approach complicates Config API significant with some occasional used method. We can discuss a bit more on the other option. I prefer to close this one and discuss on the other one.

dmlloyd:
That's my feeling as well.  I'll close this one and then we can continue the discussion on #445 and the issue #446.  If anyone has a strong objection to closing this, we can discuss reopening it.

PR #445, opened 2019-09-06
Empty values option 2

### LINKED ISSUE #446: Consistent empty/missing behavior

dmlloyd:
This is the second option for an API to support the new consistent rules for empty value handling, by creating a fluent accessor API instead of using method overloads.  Supercedes #407.  Fixes #446.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java
@@ -1,0 +1,183 @@
+/*
+ ******************************************************************************
+ * Copyright (c) 2009-2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *   2011-12-28 - Mark Struberg & Gerhard Petracek
+ *      Contributed to Apache DeltaSpike fb0131106481f0b9a8fd
+ *   2016-07-07 - Mark Struberg
+ *      Extracted the Config part out of DeltaSpike and proposed as Microprofile-Config 8ff76eb3bcfaf4fd
+ *
+ *******************************************************************************/
+package org.eclipse.microprofile.config;
+
+
+import java.util.NoSuchElementException;
+import java.util.Optional;
+
+import org.eclipse.microprofile.config.spi.Converter;
+
+/**
+ * An accessor for a configuration value.  Accessors are always immutable.  To customize
+ * the behavior of a property access, methods of this class may be called (typically in a chain)
+ * to derive successively further customized accessor instances.
+ * <p>
+ * For example, to retrieve an optional value of type {@code Integer} using a default value:
+ * <pre>
+ *  Optional&lt;Integer&gt; val = config.access("my.int.value")
+ *    .convertedForType(Integer.class)
+ *    .withDefault(123)
+ *    .optional()
+ *    .getValue();
+ * </pre>
+ * <p>
+ * Intermediate accessors may be cached for reuse; however, retaining references to accessors can be expected to
+ * cause an increase in memory footprint so this should only be done in cases where there is a net gain in code
+ * simplicity or performance.  Some environments may have capability of inlining method calls in the chain and deleting
+ * short-lived allocations; retaining intermediate accessor references might defeat this capability leading to a
+ * degradation in overall performance for some specific workloads in such environments.
+ * <p>
+ * There are some constraints on the ways in which accessors may be derived.
+ * Accessors cannot retrieve values, provide type-specific defaults, or become optional unless the accessor has an
+ * associated type or converter.  Optional accessors cannot be used to create derived accessors with a different type
+ * or with type-specific defaults.  Accessors with a type-specific default value cannot be used to create
+ * derived accessors with a different type.
+ * <p>
+ * Accessor instances are serializable if (and only if) the associated configuration, the associated converter (if any),
+ * and the associated type-safe default value (if any) are <em>all</em> serializable.
+ * <p>
+ * Accessor instances are not required to cache configuration values.  As such, if the value is needed at multiple
+ * points in time, the result of {@link #getValue()} should generally be stored someplace for reuse.
+ *
+ * @param <T> the type of value being accessed
+ *
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
+ * @author <a href="mailto:gpetracek@apache.org">Gerhard Petracek</a>
+ * @author <a href="mailto:tomas.langer@oracle.com">Tomas Langer</a>
+ */
+public interface ConfigAccessor<T> {
+
+    /**
+     * Get the associated configuration value.
+     *
+     * @return the configuration value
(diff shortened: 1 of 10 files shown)


genie-microprofile:

genie-microprofile:

dmlloyd:
Before moving the PR out of draft:

- [ ] Choose this approach and close #444 
- [ ] Ensure that a conforming implementation is possible; revise as needed
- [ ] Complete and test TCK additions necessary to test conformance
- [ ] Rebase after #443 or equivalent is merged

Emily-Jiang:
@dmlloyd You need to sort your ECA. The other thing - can you associate this PR with an issue. It might be better to open an issue to clearly specify the problem. It might make sense to copy some statement from PRs. When we release, we list issues. The closed unmerged PRs will be discarded.

dmlloyd:
ECA is sorted.  I'll open an issue.

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -92,3 +92,50 @@ If no built-in nor custom `Converter` for a requested Type `T`, an implicit Conv
...
+| `"foo,bar"` | `String[]` | `getValue` | `{ "foo", "bar" }`
+| `"foo,"`    | `String[]` | `getValue` | `{ "foo" }`
+| `",bar"`    | `String[]` | `getValue` | `{ "bar" }`
+| `",bar,"`   | `String[]` | `getValue` | `{ "bar" }`
+| missing     | `String` |`getOptionalValue` | `Optional.empty()`
+| `""`        | `String` | `getOptionalValue` | `Optional.empty()`
+| `" "`        | `String` | `getOptionalValue` | `Optional.of(" ")`
+| `"foo"`     | `String` | `getOptionalValue` | `Optional.of("foo")`
+| missing | `String[]` | `getOptionalValue` | `Optional.empty()`
+| `""` | `String[]` | `getOptionalValue` | `Optional.empty()`
```
njr-11: Some comment applies here that I added under #444,
https://github.com/eclipse/microprofile-config/pull/444/files/#r322320335

Emily-Jiang:
This approach is closer to what we have in master. With this, I think it might make better sense just deleting the snapshot and then reserve the copyright. It is not right to use the original idea but remove the original contribution. All contributors are listed under ConfigAccessor needs to be added. I think it is cleaner to delete unwanted methods and then rework. In this case, we can move faster and there won't be any copyright complication. Thoughts?

Emily-Jiang:
@dmlloyd You need to use -s to sign off. "David Lloyd (david.lloyd@****.com) did not include the "Signed-off-by footer" which is required for all commits made by a contributor."

dmlloyd:
The original accessor was builder-based and doesn't share a lot of similarities with the original in usage AFAICT.  But I'm fine with adding contribution info, if the original author(s) of the original accessor approach want it.

I'll add the sign-off once the checklist above is complete.

dmlloyd:
I have a plausible implementation of the accessor API here: https://github.com/dmlloyd/smallrye-config/tree/accessor so that should be sufficient to demonstrate the workability of the API.

dmlloyd:
Also I don't like how the spec language is a bit inconsistent so I'll probably add commits to clean that up, either as part of this or as a separate PR.  But I've filed #447 in any event.

dmlloyd:
I've updated this version (locally) to rename `getAccessor` to `access` (similarly to the original/existing version).  But I won't push it up until #443 is resolved one way or another because I will need to rebase the whole thing (either way).

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java
```diff
@@ -0,0 +1,174 @@
...
+     * is called on an accessor that is already associated with a type or converter, the previous type or converter is
+     * discarded and the new accessor will not reference it.
+     *
+     * @param clazz the type of the value to return (must not be {@code null})
+     * @param <U> the target type
+     * @return the derived accessor
+     * @throws IllegalStateException if the accessor is optional, or if the accessor is already associated with a
+     *      type-specific default value, or if there is no converter available for the given type
+     */
+    <U> ConfigAccessor<U> convertedForType(Class<U> clazz) throws IllegalStateException;
```
Emily-Jiang: I think this pattern leads to the creation of ConfigAccessor being mutable or thrown away ConfigAccessor. This is why we introduced the builder pattern in the first place.

dmlloyd: I would expect a properly tuned JDK to elide the vast majority of these allocations or else detect that they do not escape.  The config accessor is not mutable (by specification) so I'm not sure what you mean by saying it's mutable.

The "fluent" style of API is widely used already.  Using a builder on the other hand is much more unwieldy from a user's perspective, for something that will be thrown away immediately.

dmlloyd: Here are some other things to consider:

* Using the fluent immutable API means the result is perfectly type safe.  The builder approach means that no mutations are allowed which would change the type of the result (otherwise ugly tricks like unsafely re-typing the object would have to be employed).  If you forbid re-typing, the type or converter must be passed in to `access` and cannot be changed later; this also means the accessor result must have separate methods for every result type (for example, optional values would need their own method, as would collection values (if those were ever added), etc.).
* Using the builder API, you end up with an extra, strange `build` step.  These two things together make the API more clunky and decrease its flexibility: `config.access("some.prop", Foo.class).build().getOptionalValue()` versus `config.access("some.prop").withType(Foo.class).optional().getValue()`.  It's not entirely obvious that to the user that calling `build` is necessary, whereas with the immutable option, IDE completion would make it clear what operations are possible in a fairly unambiguous way based on the result type and possible method completions.
* The fluent immutable API also reflects the immutability of the `config` object itself (not reflecting the mutability of the backing data, but the actual configuration object and *its* configuration from an API perspective).

jmesnil: I agree with David, this pattern is fine and make things type safe and consistent.

```
cs = config.access("some.prop").withType(Foo.class).WithDefault("blah")
cs.getValue() -> "blah"
cs.WithDefault("Yes!") // <- this does not mutate the cs instance
cs.getValue() -> "blah"

// If users want to provide another default, he has to take the returned ConfigAccessor:
cs.withDefault("Yes!").getValue() -> "Yes!"
```

In most cases, users won't even keep a reference on the `ConfigAccessor` and will directly take the returned value for `getValue()`.

struberg: It's very similar to the original design and mixes up builder and ConfigAccessor. It's basically the old ConfigValue. We have been there over a year ago.

struberg:
@dmlloyd This is very, very similar approach to what was originally proposed over 3 years ago.
Please compare with https://github.com/struberg/javaConfig/blob/configValue/api/src/main/java/javx/config/ConfigValue.java

The design is derived from the work we've done in Apache DeltaSpike in 2011
https://github.com/apache/deltaspike/blob/master/deltaspike/core/api/src/main/java/org/apache/deltaspike/core/api/config/ConfigResolver.java#L611

api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java
Emily-Jiang: I prefer the builder pattern, which was in the master before. With the build pattern, the usage will be a functional style. You build one configAccessor and use it.

Emily-Jiang:
@struberg The current code was influenced by the previous ConfigAccessor code, which was derived from the ConfigJSR conversation (DeltaSpike influcences).

dmlloyd:
> @dmlloyd This is very, very similar approach to what was originally proposed over 3 years ago.

So I assume this means you don't have any objections to this approach.

> Please compare with https://github.com/struberg/javaConfig/blob/configValue/api/src/main/java/javx/config/ConfigValue.java

One difference I see here is that it uses unsafe casting to use the same mutable instance, as opposed to returning a new typesafe instance.

> The design is derived from the work we've done in Apache DeltaSpike in 2011
> https://github.com/apache/deltaspike/blob/master/deltaspike/core/api/src/main/java/org/apache/deltaspike/core/api/config/ConfigResolver.java#L611

Yes, this is very similar to my proposed approach.

dmlloyd:
Plan is to reopen this as new issues under the new organizational scheme.

PR #449, opened 2019-09-27
Do not iterate parent class loaders to find ConfigProviderResolver impls

### LINKED ISSUE #390: Resolver must not use load from parent first
Follow up of https://github.com/eclipse/microprofile-reactive-streams/issues/111,
loading any spi from parent first assume the impl in a parent loader is higher priority which is wrong in most environment (to not say all):

1. EE: users must be able to override container impl so it is child first
2. OSGi: using anything else than the TCCL will be broken
3. flat classpath: no issue but parent lookup is useless

Long story short you can always assume the TCCL is right.
I guess the case it tries to solve is an ear and lookup from a child webapp but here again user can want to override the "lib" part of the ear and it is a user choice to do that so better to stick to user call. Also note the containers ensure the TCCL is right so no need to try to be more clever than containers here IMHO.

dmlloyd:
Fixes #390

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
@@ -136,28 +136,19 @@ public abstract class ConfigProviderResolver {
             return null;
         }
 
-        // start from the root CL and go back down to the TCCL
-        ClassLoader parentcl = AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
-            @Override
-            public ClassLoader run() {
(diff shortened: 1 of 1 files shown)


genie-microprofile:

genie-microprofile:

dmlloyd:
I have a follow-up PR after this one that resolves #450.

Emily-Jiang:
Please ignore my previous comments as I missed the issue. Sorry. I'll continue the review and provide a update soon.

dmlloyd:
Can this be merged?  I have another follow-on PR that has been waiting for this one for weeks.

Issue #440, opened 2019-08-14
Exclude EL API transitive dependency


kenfinnigan:
Explicitly exclude `el-api` from `cdi-api` dependency

See https://github.com/eclipse/microprofile/issues/56

PR #461, opened 2019-10-17
#440 exlude EL transitive dependency

Emily-Jiang:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,6 @@ test-output
 *.iml
 *.iwl
 *.ipr
+.vscode/
 # Ignore a release.conf for perform_release/* script usage
 release.conf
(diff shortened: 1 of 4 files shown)


kenfinnigan:
I think the exclusion needs to be done in the `dependencyManagement` of the root `pom.xml` as the dependency is also used in the TCK as well

Emily-Jiang:
@kenfinnigan thanks for the info! I thought we only want to exclude from API initially. I have now excluded from the parent.

kenfinnigan:
Unrelated to your change, I just noticed that https://github.com/eclipse/microprofile-config/blob/master/tck/pom.xml#L60 defines the version for the dependency so I think that means it doesn't pick up the dependencyManagement version.

Should the version be removed so it uses the one in the root pom?

Emily-Jiang:
@kenfinnigan I think it will get picked up as the version matches. However, to prevent from future changes that make the versions out of sync, I will remove the version from tck.

PR #467, opened 2019-10-22
Use only constant global class loaders to initialize the global resolver

### LINKED ISSUE #450: ConfigProviderResolver wrongly uses TCCL in a static context
The `ConfigProviderResolver` class resolves a singleton instance based on whatever the context class loader was set to when the method was called.  This essentially means that the behavior of the singleton global resolver will depend entirely on when and how that method is first called.

The TCCL should not be consulted when constructing the global resolver instance.

The only correct solution would be to use a globally available, constant (static) class loader.  The following are available, and could be tried in order:

* The class's own class loader (this works in an environment such as WildFly where module dependencies are explicitly declared, or in flat classpath situations)
* The JDK application class loader (this works in an environment where the implementation is visible from the application class loader)
* The JDK platform class loader (this is technically already be visible from the application class loader so it does not need to be explicitly tried)

dmlloyd:
Fixes #450

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
@@ -20,8 +20,7 @@
 
 package org.eclipse.microprofile.config.spi;
 
-import java.security.AccessController;
-import java.security.PrivilegedAction;
+import java.util.Iterator;
 import java.util.ServiceLoader;
 
 import org.eclipse.microprofile.config.Config;
@@ -62,8 +61,8 @@ public abstract class ConfigProviderResolver {
      * This ConfigBuilder will initially contain no {@link ConfigSource}. The other {@link ConfigSource} will have
      * to be added manually or discovered by calling {@link ConfigBuilder#addDiscoveredSources()}.
      *
-     * This ConfigBuilder will initially contain default {@link Converter Converters}. Any other converters will need to 
-     * be added manually. 
+     * This ConfigBuilder will initially contain default {@link Converter Converters}. Any other converters will need to
+     * be added manually.
      *
      * The ConfigProvider will not manage the Config instance internally
      * @return a fresh ConfigBuilder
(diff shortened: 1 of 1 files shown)


genie-microprofile:

Emily-Jiang:
@genie-microprofile please test

dmlloyd:
Updated to fix wrong issue reference.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
```diff
@@ -105,39 +103,25 @@ public static ConfigProviderResolver instance() {
                 if (instance != null) {
                     return instance;
                 }
+                // first try my class loader
+                ConfigProviderResolver instance = loadSpi(ConfigProviderResolver.class.getClassLoader(), null);
+                // next try the system class loader
+                instance = loadSpi(null, instance);
```
Emily-Jiang: Your change will bypass the ThreadContextClassloader, as you try to use the classloader from ConfigProviderResolver or the system classloader, which might not the right classloader.

Emily-Jiang: @struberg @OndroMih Can you please check against your implementation?

dmlloyd: That is correct. There *is* no correct thread context classloader in a static initializer, because there is no way to guarantee who the caller is - it can vary between runs even.  It is never correct to reference TCCL or any thread local value from a static initializer.

OndroMih: It makes sense, @dmlloyd. When I look at it now, @Emily-Jiang, I think it's enough to have a single ConfigProviderResolver instance per JVM as it doesn't hold any state. In the end, the concrete implementation of `ConfigProviderResolver` can load the Config instance using the thread context classloader as it's done in Payara already: https://github.com/payara/Payara/blob/master/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java. I would have to review it more thoroughly but I think that Payara doesn't expect that the implementation of ConfigProviderResolver is loaded by any specific classloader.

OndroMih: I'm really not sure if it makes sense to load the resolver by both the resolver's classloader and the system classloader. I'm afraid that if the resolver impl is in the system classloader and if resolver's classloader delegates to the system classloader, the same impl would be found in both classloaders and that would end with an exception "Multiple ConfigResolverProvider implementations found". Is it really necessary to try loading also with the system classloader? In which relevant case the resolver's classloader doesn't delegate to the system classloader so that the resolver impl isn't found and needs to be looked up by the system classloader? I would delete the call to loadSpi using the system classloader and only rely on the resolver's classloader.

dmlloyd: I would be happy to revise this patch to only consider the implementation's own class loader.  It would simplify the change.

OndroMih: We've tested this PR with the current Payara implementation and all seems to work. It passes the MP TCK. So we approve this PR.

kenfinnigan:
good to merge?

PR #472, opened 2019-11-10
Deprecate releaseConfig() and replace it with unregisterConfig().

### LINKED ISSUE #471: It is impossible to release a configuration for a specific class loader
This is another design flaw with `ConfigProviderResolver`.  If a configuration is registered on two class loaders, it is not specified which one (or both) is released by `ConfigProviderResolver.releaseConfig`.

Note that no behavior is correct here: if you release just one, then the one that was released might be wrong from the caller's perspective.  If you release both, then whatever registered the other one will be disrupted.

Either the entire application-is-a-class-loader edifice should be removed, or else at least deprecate `releaseConfig(Config)` and replace it with `releaseConfig(ClassLoader)` instead.

dmlloyd:
Fixes #471

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
@@ -62,8 +62,8 @@ public abstract class ConfigProviderResolver {
      * This ConfigBuilder will initially contain no {@link ConfigSource}. The other {@link ConfigSource} will have
      * to be added manually or discovered by calling {@link ConfigBuilder#addDiscoveredSources()}.
      *
-     * This ConfigBuilder will initially contain default {@link Converter Converters}. Any other converters will need to 
-     * be added manually. 
+     * This ConfigBuilder will initially contain default {@link Converter Converters}. Any other converters will need to
+     * be added manually.
      *
      * The ConfigProvider will not manage the Config instance internally
      * @return a fresh ConfigBuilder
@@ -91,8 +91,22 @@ public abstract class ConfigProviderResolver {
      *
      * If the given Config is associated within an Application then it will be unregistered.
      * @param config the config to be released
+     *
+     * @deprecated This method is not adequately specified and can lead to problems when a configuration is registered
+     * to multiple class loaders; please use {@link #unregisterConfig(ClassLoader)} instead.
      */
+    @Deprecated
     public abstract void releaseConfig(Config config);
+
+    /**
+     * Unregister the configuration associated with the given class loader.  If no configuration is registered,
+     * {@code null} is returned, otherwise the previously-registered configuration is returned.
+     *
+     * @param classLoader the class loader of the configuration
+     * @return the configuration which was previously registered to the class loader, or {@code null} if none was
+     *      registered
+     */
+    public abstract Config unregisterConfig(ClassLoader classLoader);
 
     /**
      * Creates a ConfigProviderResolver object


dmlloyd:
It is in fact required, because unless the specification designates a specific owner and lifecycle for config objects (and it doesn't), then there's no reason why a specific config couldn't have been registered more than once by different parties.

For example, the same config might be in use by two applications; when one application is destroyed, its class loader must be released, but there's no reason to disrupt the other application.

kenfinnigan:
I also agree with this being an undefined behavior gap that needs to be closed.

struberg:
I think we defined somewhere that a Config must at maximum be registered with a single 'Application'. This has to do with the lifecycle of ConfigSources.

Remember our discussions about ConfigSources which someone wants to use in multiple Apps? Like e.g. a shared DbConfigSource for the whole Container? And that in this case, you need to have a kind of proxy ConfigSource, 1 per App, which use internal reference counting? Please go ask @jmesnil for further infos.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java
```diff
@@ -91,9 +91,23 @@ protected ConfigProviderResolver() {
      *
      * If the given Config is associated within an Application then it will be unregistered.
      * @param config the config to be released
+     *
+     * @deprecated This method is not adequately specified and can lead to problems when a configuration is registered
```
struberg: No this statement is based on a fundamentally wrong assumption. A Config must at maximum be registered with a single Application (-> 1 ClassLoader)

dmlloyd: Here's the problem: I can't find anything in the specification that says this.

When you're dealing in specifications, everything in the specification is normative, which is to say "it's the truth".  Everything *not* in the specification is an opinion, even if it has historical significance.  It is not correct to say that the assumption is wrong, because until this proposed change, there's nothing in the specification to indicate whether it is wrong or right (hence the change).

So in order to resolve the issue, we have to update the specification in *some* way to make it consistent.  It may be your opinion that certain APIs were design with certain intent, but while that information is useful to feed into the decision process, it is secondary to other concerns, such as compatibility.

When I talk about "compatibility" I'm not just talking about it in the traditional sense of source or binary or even behavioral compatibility.  I'm talking about whether there is value in essentially cutting off or disabling existing uses and use cases that were established due to the interpretation of the specification *as it is written*, rather than *as it was originally intended to be*.  So while the correct remedy *might* be to "fix" the specification to meet the original intent, it also might *not* be the correct remedy given the ways in which the specification has already been adopted for usage.

So what I'm trying to say is that if, whether by intent or by error, something is left out of the specification, then that omission is now normative or "correct" by definition, and once it's out in the world one must assume that this normative behavior now cannot just be "fixed up" after the fact without consequences.  And in viewing those consequences one must consider existing users and use cases, and whether these can be justifiably invalidated (I believe the answer to this question is nearly always going to be "no" because use cases generally do not *become* invalid; they either are or are not, and sometimes we might make a mistake in that judgement).

dmlloyd:
Closing for now. Will revisit the original use case later.

struberg:
Thanks David!
Yes, we might miss some documentation here or there. We considered the 'multi-app' use case as so advanced that we discussed it but did not add it to the docs it seems. It's really an edge case which is mostly important for integrators and container vendors.

Issue #457, opened 2019-10-16
Incorrect description in the README


kenfinnigan:
The project README has the following statement in the third paragraph:

> Under some circumstances, some data sources may change dynamically. The changed values should be fed into the client without the need for restarting the application. This requirement is particularly important for microservices running in a cloud environment. The MicroProfile Config approach allows to pick up configured values immediately after they got changed.

As the current specification doesn't describe the ability to handle dynamically changing values, it is probably a good idea to remove this statement.

Emily-Jiang:
Both spec and Javadoc on ConfigProperty annotation mentions the retrieval of the dynamic config property. 

> //Injects a Provider for the value of myprj.some.dynamic.timeout property to
    //resolve the property dynamically. Each invocation to Provider#get() will
    //resolve the latest value from underlying Config.
    //The existence of configured values will get checked during startup.
    //Instances of Provider<T> are guaranteed to be Serializable.
    @Inject
    @ConfigProperty(name="myprj.some.dynamic.timeout", defaultValue="100")
    private javax.inject.Provider<Long> timeout;

We left how the runtime handles the dynamism unspeced on purpose.

kenfinnigan:
I'd missed that reference in searching yesterday.

Not sure if there are other issues related to this, but I think the spec needs to move away from leaving it undefined otherwise it becomes a huge gray area where implementations can do completely different things and users get unexpected behavior if they migrate between them

Emily-Jiang:
Thanks for your feedback, Ken! Points taken. Let's add a couple of sentences in the spec to clearly spell it out.

ljnelson:
Related: #369, #370, #371, #395, #431

Emily-Jiang:
Since we have not reached a conclusion, I'll move this issue out of the MP Config 1.4 milestone.

dmlloyd:
Also related: #432 #304

Emily-Jiang:
fixed by #478

PR #478, opened 2019-11-22
#457 add a section to make dynamic config source more visible

Emily-Jiang:

### DIFF
--- a/spec/src/main/asciidoc/configsources.asciidoc
+++ b/spec/src/main/asciidoc/configsources.asciidoc
@@ -149,8 +149,12 @@ Please note that a single `ConfigSource` should be either registered directly or
 
 === Dynamic ConfigSource
 
-A dynamic ConfigSource is a `ConfigSource` where its content can be updated, such as updating a property name or value or adding or deleting config properties 
-from it. For the property lookup, the method `config.getValue()` or `config.getOptionalValue()` retrieves the up-to-date value.
+As a `ConfigSource` is a view of configuration data, its data may be changing, or unchanging. 
+If the data is changing, and a `ConfigSource` can represent its changes, we call that `ConfigSource` a dynamic `ConfigSource`, since at any two moments two operations on it may reflect two different sets of underlying configuration data. 
+If instead the data is unchanging, we call the `ConfigSource` a static `ConfigSource`, since at any two moments two operations on it will reflect only one set of underlying (unchanging) configuration data. 
+A caller cannot know whether a `ConfigSource` is dynamic or static. 
+
+For the property lookup, the method `config.getValue()` or `config.getOptionalValue()` retrieves the up-to-date value.
 Alternatively, for the injection style, the following lookup should be used to retrieve the up-to-date value.
 [source, text]
 ----
--- a/spec/src/main/asciidoc/configsources.asciidoc
+++ b/spec/src/main/asciidoc/configsources.asciidoc
@@ -148,6 +148,7 @@ public class ExampleYamlConfigSourceProvider
 Please note that a single `ConfigSource` should be either registered directly or via a `ConfigSourceProvider`, but never both ways.
 
 === Dynamic ConfigSource
+
 A dynamic ConfigSource is a `ConfigSource` where its content can be updated, such as updating a property name or value or adding or deleting config properties 
 from it. For the property lookup, the method `config.getValue()` or `config.getOptionalValue()` retrieves the up-to-date value.
 Alternatively, for the injection style, the following lookup should be used to retrieve the up-to-date value.
@@ -157,14 +158,14 @@ Alternatively, for the injection style, the following lookup should be used to r
     @ConfigProperty(name="myprj.some.dynamic.timeout", defaultValue="100")
     private javax.inject.Provider<Long> timeout; 
 ----
-Whether a `ConfigSource` supports this dynamic behavior or not is dependent on how it's implemented.
-For an instance, the default `ConfigSource` microprofile-config.properties and Environment Variables are not dynamic 
+Whether a `ConfigSource` supports this dynamic behavior or not depends on how it's implemented.
+For instance, the default `ConfigSource` microprofile-config.properties and Environment Variables are not dynamic 
 while System Properties are dynamic by nature. MicroProfile Config Implementation can decide whether 
 a `ConfigSource` can be dynamic or not.
 
 === Cleaning up a ConfigSource
 
-If a `ConfigSource` implements the `java.lang.AutoCloseable` interface  then the `close()` method will be called when the underlying `Config` is being released.
+If a `ConfigSource` implements the `java.lang.AutoCloseable` interface then the `close()` method will be called when the underlying `Config` is being released.
 
 === ConfigSource and Mutable Data
 
--- a/spec/src/main/asciidoc/configsources.asciidoc
+++ b/spec/src/main/asciidoc/configsources.asciidoc
@@ -160,7 +160,8 @@ Alternatively, for the injection style, the following lookup should be used to r
 Whether a `ConfigSource` supports this dynamic behavior or not is dependent on how it's implemented.
 For an instance, the default `ConfigSource` microprofile-config.properties and Environment Variables are not dynamic 
 while System Properties are dynamic by nature. MicroProfile Config Implementation can decide whether 
-a `ConfigSource` can be dyanamic or not.
+a `ConfigSource` can be dynamic or not.
+
 === Cleaning up a ConfigSource
 
 If a `ConfigSource` implements the `java.lang.AutoCloseable` interface  then the `close()` method will be called when the underlying `Config` is being released.


spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -147,6 +147,17 @@ public class ExampleYamlConfigSourceProvider
...
+A dynamic ConfigSource is a `ConfigSource` where its content can be updated, such as updating a property name or value or adding or deleting config properties 
+from it. For the property lookup, the method `config.getValue()` or `config.getOptionalValue()` retrieves the up-to-date value.
+Alternatively, for the injection style, the following lookup should be used to retrieve the up-to-date value.
+[source, text]
+----
+    @Inject
+    @ConfigProperty(name="myprj.some.dynamic.timeout", defaultValue="100")
+    private javax.inject.Provider<Long> timeout; 
+----
+
```
kenfinnigan: Maybe also mention that whether a `ConfigSource` supports this dynamic behavior or not is dependent on how it's implemented.

Need to make clear there's no guarantee as to whether any `ConfigSource` does, or does not, support this dynamic behavior and that it's a case by case basis

Emily-Jiang:
@dmlloyd Since the spec already supported dynamic behaviour and having tck to verify this. I think it is a good idea to clarify this. I did add a few sentences as per @kenfinnigan 's comments. As for the thread-safety, I think we were more in the context of whether all properties appear in `getPropertyNames` as you rightly pointed out some sensitive properties might be filtered out on purpose.

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -157,7 +157,10 @@ Alternatively, for the injection style, the following lookup should be used to r
     @ConfigProperty(name="myprj.some.dynamic.timeout", defaultValue="100")
     private javax.inject.Provider<Long> timeout; 
 ----
-
+Whether a `ConfigSource` supports this dynamic behavior or not is dependent on how it's implemented.
+For an instance, the default `ConfigSource` microprofile-config.properties and Environment Variables are not dynamic 
+while System Properties are dynamic by nature. MicroProfile Config Implementation can decide whether
```
ljnelson: "MicroProfile Config Implementation [sic] can decide whether a `ConfigSource` can be dynamic or not": I find this hard to understand.  `ConfigSource`s can be packaged with applications.  How can a given MicroProfile Config implementation "decide" whether a `ConfigSource` packaged with an application is dynamic or not?

kenfinnigan: It somewhat depends on whether a `ConfigSource` implementation using a backing source that could be updated.

ljnelson: @kenfinnigan Not trying to be difficult, but again: how is, say, an OpenLiberty MicroProfile Config implementation shipped with an application server supposed to "know" that my `ConfigSource` implementation—the one in my application deployed to that server—is dynamic?  It cannot.

kenfinnigan: Ah, see your point now.

So yes, an MP Config implementation can only "know" whether the `ConfigSource`s it provides are dynamic or not. There is no way for it to "know" about any `ConfigSource` that is included within the application code directly

ljnelson: Correct.  In other words: what is the contract of [`ConfigSource#getPropertyNames()`](https://javadoc.io/static/org.eclipse.microprofile.config/microprofile-config-api/1.3/org/eclipse/microprofile/config/spi/ConfigSource.html#getPropertyNames--)?  It is so underspecified as to be almost useless, _viz._ is the returned `Set` an immutable snapshot?  A mutating snapshot?  Is it thread safe (i.e. what must I do to iterate over it)?  If I call the method two times must I get back equal `Set`s?  Identical `Set`s?  And so on.  (And that's before even considering the [`ConfigSource#getProperties()`](https://javadoc.io/static/org.eclipse.microprofile.config/microprofile-config-api/1.3/org/eclipse/microprofile/config/spi/ConfigSource.html#getProperties--) method, which suffers from the same kind of lack of specificity.)

kenfinnigan: I would agree that the way it's documented, and the fluidity around whether a `ConfigSource` support dynamic updates or not, means that it's impossible to write meaningful TCK tests to verify any of this as it's undefined

ljnelson: Right, and not just TCK tests: this is a public API exposed to the end user.  She cannot use it because she does not know how to use the returned `Set`.  Must she synchronize on it?  Or not?  Can she cache it?  And so on.  We are in agreement here at a high level that there is _much_ more work to be done here, I think.

kenfinnigan: I agree that there is still lots to be done.

In terms of this PR, I'm wondering if we need to go back to discussing improving the definitions of the API first.

Emily-Jiang: @ljnelson I was going to add a PR to say the getPropertyNames() can be iterated in a thread safe manner based on the suggestion from @dmlloyd 
@kenfinnigan there is a TCK to verify the dynamic aspect of system property.

ljnelson: Yes, that's a good start.  An even better start would be to make sure that the iteration is _fully_ specified.  See the [Concurrent Collections section in the package description of the `java.util.concurrent` package documentation](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/package-summary.html#package.description) for a particularly good example.

dmlloyd:
Is this still current, given the recent spec updates?

Emily-Jiang:
I think this is still relevant as the spec still supports dynamic configsources. It worths the clarification.

spec/src/main/asciidoc/configsources.asciidoc
Emily-Jiang: Follow up on what I mentioned there are tcks to assert dynamic config lookup.
ConfigProviderTest.testDynamicValueInPropertyConfigSource()
ConfigProviderTest.testGetPropertyNames()
ConfigProviderTest.testPropertyConfigSource() 
EmptyValuesTest.testEmptyStringProgrammaticLookup()

Emily-Jiang:
I would like to merge this PR by 15th July as I have answered all questions and have got 2 approvals. Please log your comments asap.

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -147,6 +147,20 @@ public class ExampleYamlConfigSourceProvider
...
+from it. For the property lookup, the method `config.getValue()` or `config.getOptionalValue()` retrieves the up-to-date value.
+Alternatively, for the injection style, the following lookup should be used to retrieve the up-to-date value.
+[source, text]
+----
+    @Inject
+    @ConfigProperty(name="myprj.some.dynamic.timeout", defaultValue="100")
+    private javax.inject.Provider<Long> timeout; 
+----
+Whether a `ConfigSource` supports this dynamic behavior or not is dependent on how it's implemented.
+For an instance, the default `ConfigSource` microprofile-config.properties and Environment Variables are not dynamic
```
radcortez, line 161: > For an instance -> For instance

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -147,6 +147,20 @@ public class ExampleYamlConfigSourceProvider
...
+[source, text]
+----
+    @Inject
+    @ConfigProperty(name="myprj.some.dynamic.timeout", defaultValue="100")
+    private javax.inject.Provider<Long> timeout; 
+----
+Whether a `ConfigSource` supports this dynamic behavior or not is dependent on how it's implemented.
+For an instance, the default `ConfigSource` microprofile-config.properties and Environment Variables are not dynamic 
+while System Properties are dynamic by nature. MicroProfile Config Implementation can decide whether 
+a `ConfigSource` can be dyanamic or not.
```
radcortez, line 163: > dyanamic -> dynamic

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -147,6 +147,20 @@ public class ExampleYamlConfigSourceProvider
...
 === Cleaning up a ConfigSource
```
radcortez, line 164: > === Cleaning up a ConfigSource -> Needs a break line before to format properly.

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -147,6 +147,20 @@ public class ExampleYamlConfigSourceProvider
...
 
 If a `ConfigSource` implements the `java.lang.AutoCloseable` interface  then the `close()` method will be called when the underlying `Config` is being released.
```
radcortez, line 166: > interface  then -> interface then (extra space)

spec/src/main/asciidoc/configsources.asciidoc
Emily-Jiang: done

Emily-Jiang:
@radcortez I have done all of the tidy up including the format. Please double check and approve if no more comments.

radcortez:
We probably need for @ljnelson to check the latest update to see if it addresses his comments.

Emily-Jiang:
@ljnelson can you double check to see whether this PR has addressed your issue?

ljnelson:
It is written:

> A dynamic ConfigSource is a `ConfigSource` where its content can be updated, such as updating a property name or value or adding or deleting config properties from it.

With respect, that is not grammatically correct.  Nor does it indicate who is doing the updating: a naïve user might very well see "its content can be updated" and start looking for an `update()` method on `org.eclipse.microprofile.config.spi.ConfigSource`, or might look for a class named `DynamicConfigSource` or something.

I think instead you're trying to say something about the dynamism of the data that underlies any given `ConfigSource`, not the dynamism of the `ConfigSource` itself, right?

In other words, I think you might be trying to say something as simple as this (and indeed maybe the whole PR comes down to a couple of sentences):

"A `ConfigSource` is a view of configuration data.  That data may be changing, or unchanging.  If the data is changing, and a `ConfigSource` can represent its changes, we call that `ConfigSource` a _dynamic_ `ConfigSource`, since at any two moments two operations on it may reflect two different sets of underlying configuration data.  If instead the data is unchanging, we call the `ConfigSource` {something else that I don't think anyone has named yet; maybe _static_}, since at any two moments two operations on it will reflect only one set of underlying (unchanging) configuration data.  A caller cannot know whether a `ConfigSource` is dynamic or static."

That is just a suggestion.  (I sense some sudden urgency to this PR for some reason so I won't stand in the way of whatever you want to do.)

radcortez:
@ljnelson the urgency is because there is planned RC release of MP platform by 1st week of August, so all individuals specs should have their RC's ready as well.

I like your text better, so I think we can modify it to better express the view. Cheers!

Emily-Jiang:
@ljnelson I have updated the sentences in question with your suggested text with slight modification. Please double check to see whether you have any further comments.

Emily-Jiang:
@ljnelson I'll merge this PR by COB today if I don't hear more feedback from you today.

Issue #404, opened 2019-02-13
Clarify if @ConfigProperty injected values are bean passivating enabled.


rdebusscher:
There is not much info in the specification around the fact if an injected `@ConfigProperty` is passivating enabled or not.

For example

    @Inject
    @ConfigProperty(name = "config.foo.value")
    private String value;

Will fail upon deployment in some runtimes when it it used in a passivating scoped bean like `SessionScoped`.

There is only a mentioning about the Provider variant
    
     Instances of Provider<T> are guaranteed to be Serializable.

So we should mention the restrictions of using `@ConfigProperty` in a passivating scoped bean.

Emily-Jiang:
I prefer not to add this info into the spec. Please comment if you think otherwise.

ljnelson:
I think the issue here is (if I'm reading it right): using `@ConfigProperty` in any bean prevents that bean from being passivated.  That means no bean can be passivated if it uses MicroProfile Config.  That seems important to call out.

Emily-Jiang:
Using `@ConfigProperty` does not cause any issue, as it is just a qualifier. The real issue is `@Inject` as it declares injection point as explained in my previous comment.

ljnelson:
Isn't all of this important to spell out to end users?

Emily-Jiang:
Let's add a sentence in this spec to mention CDI spec  (Section 6.6.2) as per https://github.com/eclipse/microprofile-config/issues/404#issuecomment-543213159

rdebusscher:
And we need a TCK test to enforce it for the implementations,  I guess.

Inject into a CDI scoped bean?

Emily-Jiang:
@rdebusscher as I explained this is not introduced by Config spec but by CDI. I am not sure we need a TCK for this. The sentence is just for CDI spec awareness though.

PR #479, opened 2019-11-22
#404 Add some info regarding passivation capable injection point

Emily-Jiang:

### DIFF
--- a/spec/src/main/asciidoc/configexamples.asciidoc
+++ b/spec/src/main/asciidoc/configexamples.asciidoc
@@ -23,7 +23,7 @@
 == Config Usage Examples
 
 
-An application can obtain it's configuration programmatically via the `ConfigProvider`.
+An application can obtain it's configuration programmatically via the `ConfigProvider`. 
 In CDI enabled beans it can also get injected via `@Inject Config`.
 An application can then access its configured values via this `Config` instance.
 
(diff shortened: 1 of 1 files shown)


Emily-Jiang:
@rdebusscher can you check to see whether you are happy with the PR?

rdebusscher:
@Emily-Jiang  Seems that I made a comment on the issue and not here https://github.com/eclipse/microprofile-config/issues/404#issuecomment-558336454

Will the TCK update be in a separate PR or in this one?

The text update itself is sufficient for me.

Emily-Jiang:
@rdebusscher as per my explanation, Config does not change any mandated behaviour from CDI. The behaviour you observed is purely CDI. I don't think we should add tests here to test CDI. Adding the sentence is just to provide some pointer to the corresponding section in CDI.

rdebusscher:
@Emily-Jiang I understand it is CDi related, but not all implementations behave properly (not checked anymore since I opened the issue). If we do not enforce it through some TCK test, there might be maybe a difference in behaviour forever.

Emily-Jiang:
Which container did you check? As per my comments earlier, all container with CDI complianced should behave the same. The reason I don't think we need to add tests is that Config did not play any part in this game. It is only to do with `@Inject`.

Issue #431, opened 2019-07-18
What must the set of property names returned by Config#getPropertyNames() consist of?


ljnelson:
The javadocs for `Config#getPropertyNames()` say, in their entirety, and I see no other language related to this topic in the specification documents:

> Return a collection of property names.
> **Returns:**
> the names of all configured keys of the underlying configuration.

It is hard to know what to make of this.  Does this mean:
1. The set of property names returned by `Config#getPropertyNames()` must consist only of the combination of the sets of property names returned by each `ConfigSource` present in the return value of `Config#getConfigSources()` and nothing else
2. The set of property names returned by `Config#getPropertyNames()` may contain a subset of the combination of the sets of property names returned by each `ConfigSource` present in the return value of `Config#getConfigSources()`
3. The set of property names returned by `Config#getPropertyNames()` can be anything the implementor wants, regardless of whether it is represented in part or in whole by any `ConfigSource` notionally contained by the `Config`
…?  It must be one of these.  The specification does not say, and there is no TCK test.

I am assuming that (1) is the proper answer (modulo any information related to #370).  However, given that the specification says nothing and there is no TCK<sup>1</sup>, (3) is the _de facto_ answer at the moment.

If (1) is the proper answer, should a TCK test exist for this?

If (1) is not the proper answer, why does this method exist?

<sup>1</sup> There is https://github.com/eclipse/microprofile-config/blob/828ee2fee4ccbef6701943bb96d7be187408fd63/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java#L170-L183 which indirectly tests only that `Config#getPropertyNames()` must contain at least one key returned by the `getPropertyNames()` methods of the `ConfigSource` instances returned by `getConfigSources()` at the moment of the `Config#getPropertyNames()` call.

ljnelson:
Related: #370, #371, #395

Emily-Jiang:
@ljnelson It should be the following interpretation:

> The set of property names returned by Config#getPropertyNames() must consist only of the combination of the sets of property names returned by each ConfigSource present in the return value of Config#getConfigSources() and nothing else

ljnelson:
@Emily-Jiang Thanks very much.  Shall I contribute a TCK test for this?

Emily-Jiang:
@ljnelson please feel free to contribute. Thank you!

Emily-Jiang:
[...]

PR #480, opened 2019-11-22
#431 rework the configsource api for clarification

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
@@ -29,7 +29,6 @@
 package org.eclipse.microprofile.config.spi;
 
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -28,6 +28,8 @@
  *******************************************************************************/
 package org.eclipse.microprofile.config.spi;
 
+import java.util.HashMap;
+import java.util.LinkedHashMap;
```
kenfinnigan: Don't think the `LinkedHashMap` import is used

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -81,7 +83,11 @@
...
      */
-    Map<String, String> getProperties();
+    default Map<String, String> getProperties() {
+        Map<String, String> props = new HashMap<>();
+        getPropertyNames().stream().forEach((prop) -> props.put(prop, getValue(prop)));
+        return props;
+    }
 
     /**
      * Gets all property names known to this config source, without evaluating the values.
```
ljnelson: I assume "without evaluating the values" should really be "potentially without evaluating the values".

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
Emily-Jiang: removed.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
Emily-Jiang: updated

PR #481, opened 2019-11-29
#431 - undo this change as it has binary incompatible changes

Emily-Jiang:


Emily-Jiang:
@genie-microprofile please test

Emily-Jiang:
If we provide both default implementation, will this potentially make people feel they don't need to implement either of the methods?

dmlloyd:
It should be fairly evident that they have to implement *one* of the methods, else their class will generally be empty.  If the class is documented properly and `getProperties` is deprecated, it should be clear enough what is happening.

If we want to be able to remove `getProperties` in 2.0 (which we do IMO), then getting it deprecated now is important.

Emily-Jiang:
I think Payara does need getProperties for their tool. We just provide a default impl and move on. I don't have strong feeling on whether to provide two defaults or not. If we do, in 2.0, we should remove the getPropertyNames to be abstract. Let me do a commit to return an empty set for the getPropertyNames and see what other people think.

dmlloyd:
It might even be best to just return an empty set for property names and empty map for properties.  That way you won't end up with any surprise recursion.

tevans78:
As noted here (https://github.com/eclipse/microprofile-config/pull/482#issuecomment-562658437), the changes to the default methods in ConfigSource are not backwards compatible, so reverting is a good idea.

tevans78:
One minor improvement (that could be a follow-up PR) ... add a null guard to prevent the default impl of getPropertyNames from throwing an NPE if null is returned from getProperties().
e.g.
```
    default Set<String> getPropertyNames(){
    	Map<String, String> properties = getProperties();
    	return properties == null ? Collections.emptySet(): properties.keySet();
    }
```

dmlloyd:
> One minor improvement (that could be a follow-up PR) ... add a null guard to prevent the default impl of getPropertyNames from throwing an NPE if null is returned from getProperties().

This should definitely be a follow-up: we've gone this long without the extra `null` check, so it is not suddenly a new issue.

dmlloyd:
I propose instead that `getProperties` should be deprecated and the default implementation in #482 be added, and no other changes done at the same time.  We're getting killed by non-incremental changes: something needed will be bundled with 10 things that are not needed so no progress is ever made.  I think adding just the one default impl and the deprecation docs will be very non-controversial.

Let's keep each change *small*, *specific*, and *incremental* please.

Emily-Jiang:
I propose to merge this one to go back to square one again. This just undoes the changes recently committed. 
Let's discuss this again under #431. Note: Please don't add any more comments on this as this is just a level set back to Config 1.3 again for ConfigSource class.

PR #482, opened 2019-12-05
#431 further changes to maintain backward compatiblility

Emily-Jiang:

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
@@ -30,7 +30,6 @@ package org.eclipse.microprofile.config.spi;
 
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
@@ -153,7 +152,8 @@ public interface ConfigSource {
         if (configOrdinal != null) {
             try {
                 return Integer.parseInt(configOrdinal);
-            } catch (final NumberFormatException ignored) {
+            } 
+            catch (final NumberFormatException ignored) {
 
             }
         }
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -83,63 +84,76 @@
...
 
     /**
-     * Gets all property names known to this config source, potentially without evaluating the values.
+     * Gets all property names known to this config source, potentially without
+     * evaluating the values.
      *
-     * For backwards compatibility, there is a default implementation that just returns the keys of {@code getProperties()}
-     * slower ConfigSource implementations should replace this with a more performant implementation
+     * For backwards compatibility, there is a default implementation that just returns
+     * an empty set. This method should be overriden.
```
ljnelson: Minor, but I would prefer this documentation to read something like the following:

    The default implementation of this method returns an empty {@link Set}.
    {@link ConfigSource} implementations should override this method.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -83,63 +84,76 @@
...
     }
 
     /**
-     * Gets all property names known to this config source, potentially without evaluating the values.
+     * Gets all property names known to this config source, potentially without
+     * evaluating the values.
      *
-     * For backwards compatibility, there is a default implementation that just returns the keys of {@code getProperties()}
-     * slower ConfigSource implementations should replace this with a more performant implementation
+     * For backwards compatibility, there is a default implementation that just returns
```
ljnelson: Javadoc improvements: normally every paragraph after the first sentence should be wrapped in `<p>` tags.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
Emily-Jiang: done

tevans78:
This change is not backward compatible with exsiting older ConfigSource implementations which were relying on the default implementation of getPropertyNames. A call to Config.getPropertyNames() would previously returned a complete set based on the keys from ConfigSource.getProperties(). Now, if ConfigSource.getPropertyNames() is not overridden, that set will not be complete.

Emily-Jiang:
Not going to pursue this one any further, so I closed it.

PR #468, opened 2019-10-25
Deprecate ConfigProviderResolver.setInstance()

dmlloyd:
Fixes #364

Signed-off-by: David M. Lloyd <david.lloyd@redhat.com>

PR #488, opened 2020-01-13
Prevent incorrect caching of config provider resolver instance

### LINKED ISSUE #265: Overriding the ConfigProviderResolver doesn't work right

dmlloyd:
Fixes #265


dmlloyd:
If #468 is not accepted, then this patch must be accepted in order to ensure correct operation of `ConfigProvider.getInstance()` under all circumstances.

Issue #486, opened 2020-01-10
javadoc generation for the API fails when using Java 11


tevans78:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for MicroProfile Config 1.4:
[INFO] 
[INFO] MicroProfile Config ................................ SUCCESS [  3.340 s]
[INFO] MicroProfile Config API ............................ FAILURE [  6.398 s]
[INFO] MicroProfile Config TCK ............................ SKIPPED
[INFO] MicroProfile Config Specification .................. SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  11.292 s
[INFO] Finished at: 2020-01-10T17:43:59Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.10.4:jar (attach-javadocs) on project microprofile-config-api: MavenReportException: Error while generating Javadoc: 
[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in http://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.
[ERROR] 
[ERROR] Command line was: /Library/Java/JavaVirtualMachines/adoptopenjdk-11-openj9.jdk/Contents/Home/bin/javadoc @options @packages
[ERROR] 
[ERROR] Refer to the generated Javadoc files in '/Users/tevans/Liberty/microprofile/microprofile-config/api/target/apidocs' dir.
[ERROR] 
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :microprofile-config-api
```

tevans78:
Works if javadoc generation is skipped (`-Dmaven.javadoc.skip=true`)

Emily-Jiang:
It seems it is a JDK [issue](https://bugs.openjdk.java.net/browse/JDK-8212233).

[...]

PR #490, opened 2020-01-16
#486 make MP Config compilable with Java11

Emily-Jiang:

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
@@ -155,15 +155,13 @@ public class ConfigProviderTest extends Arquillian {
         ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
         try (ObjectOutputStream out = new ObjectOutputStream(byteArrayOutputStream)) {
             out.writeObject(config);
-        } 
-        catch (IOException ex) {
+        } catch (IOException ex) {
             Assert.fail("Injected config should be serializable, but could not serialize it", ex);
         }
         Object readObject = null;
         try (ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(byteArrayOutputStream.toByteArray()))) {
             readObject = in.readObject();
-        } 
-        catch (IOException | ClassNotFoundException ex) {
+        } catch (IOException | ClassNotFoundException ex) {
             Assert.fail("Injected config should be serializable, but could not deserialize a previously serialized instance", ex);
         }
         MatcherAssert.assertThat("Deserialized object", readObject, CoreMatchers.instanceOf(Config.class));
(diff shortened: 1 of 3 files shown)


Emily-Jiang:
This closes #486

Emily-Jiang:
@dmlloyd please take a look

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
```diff
@@ -155,13 +155,15 @@ public void testInjectedConfigSerializable() {
         ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
         try (ObjectOutputStream out = new ObjectOutputStream(byteArrayOutputStream)) {
             out.writeObject(config);
-        } catch (IOException ex) {
```
kenfinnigan: Are the formatting changes in this and `ConverterTest` required for JDK 11?

dmlloyd: I was thinking the same, but just assumed it was due to checkstyle.  But if not, then the formatting changes should be left out probably...

kenfinnigan: +1

Emily-Jiang: It was to sastisfy the checkstyle.

kenfinnigan:
Why the checkstyle rule change?

I don’t think we want open braces on the next line

pzygielo:
Which modification is the key to make MPC Java11 compilable? Is CS upgrade and related changes necessary to accomplish this goal? On my machine upgrading just the `maven-javadoc-plugin` does the job.

Emily-Jiang:
@kenfinnigan the new version of checkstyle requires the closing bracket is on the new line.
@pzygielo I updated the javadoc version together with others to move up the version.

pzygielo:
@Emily-Jiang 
>  I updated the javadoc version together with others to move up the version.

I know, but it is out of the scope of commit message. It seems like other changes are required, which they are not.

kenfinnigan:
@Emily-Jiang then the rule needs to be adjusted. We shouldn't be changing the code style due to upgrading the version of checkstyle

Emily-Jiang:
> Why the checkstyle rule change?
I don’t think we want open braces on the next line

The checkstyle has a backward changes from 2.17 to 3.1. I had a look at the [option](https://checkstyle.sourceforge.io/property_types.html#rcurly) for right curly braces.
alone, alone_or_singleline, same
In the version 2.17, we used 'alone`, which does not enforce alone.
I tried to change to same. If I do, I need to change a few more files.

Basically, we cannot maintain the inconsistent state anymore. @kenfinnigan it is not open braces, but close braces. I think it is ok.

kenfinnigan:
Ah yeah, close braces. My preference is not to have a closing brace on a different line than any `catch`.

I'm leaning towards an earlier comment that we need to revert checkstyle updates and just make the changes needed for JDK 11, and revisit checkstyle issues later.

Ideally the whole of MP uses the same checkstyle rules.

Emily-Jiang:
@kenfinnigan I've reverted the checkstyle upgrade. I've raised [this](https://github.com/eclipse/microprofile/issues/159) issue and discuss further on the MP project level.

Issue #492, opened 2020-01-16
expose conversion mechanism in API


emattheis:
There are certain cases where a user of microprofile-config might want to convert a supplied string into different type using the same mechanism as a resolved property value. For example, when a fallback value for a particular usage point is available from a string supplier, but not appropriate to contribute as a formal config source.

I suggest adding the following method to `org.eclipse.microprofile.config.Config`:

```java
<T> T convert(String value, Class<T> type);
```

dmlloyd:
In SmallRye Config, we expose that functionality like this:

```java
    public <T> T convert(String value, Class<T> asType) {
        return value != null ? getConverter(asType).convert(value) : null;
    }

    public <T> Converter<T> getConverter(Class<T> asType) {
        // ...
    }
```

Which can be useful if one wants to cache the converter or perform repeated conversions, or create delegating converters, etc.

We also (internally) support `getOptionalConverter` which might be of use in the general API (yielding a `Converter<Optional<T>>`.  But I'm not sure every implementation can support `Optional` using the converter infrastructure like we do.

Emily-Jiang:
I know converter is a useful functionality. I talked with someone else to potentially create a converter spec. Since it is a smallish update, I'll attempt to propose a PR soon.

Emily-Jiang:
Actually, I thought about this more yesterday and I'm not sure about adding the convert on the config api. It should not belong to Config object, which just a collections of configuration.

My proposal is to is directly expose Converter. I think we can do:
```
@Inject Converter converter;
MyType t = converter.convert("thisString");
#Programmtic way to look up converter
Converter converter = ConfigProviderResolver.getConverter();
converter.convert("thisString");
```
A PR will follow soon.

dmlloyd:
> Actually, I thought about this more yesterday and I'm not sure about adding the convert on the config api. It should not belong to Config object, which just a collections of configuration.

This statement is incorrect.  The `Config` object clearly associates both converters and configuration sources on a per-configuration basis as can be trivially extrapolated by the methods on `ConfigBuilder`.  The original method as proposed is more correct, especially given that custom converters can be associated with a configuration.  Your proposed mechanism loses this completely, and also doesn't actually work (there's no way to specify the target type in your programmatic approach).

emattheis:
[...]

PR #495, opened 2020-01-17
add getConverter method to Config

emattheis:
as proposed in #492

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -37,6 +37,7 @@ package org.eclipse.microprofile.config;
 import java.util.Optional;
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.Converter;
 
 /**
  * <p>
@@ -161,4 +162,15 @@ public interface Config {
      * @return the configuration sources
      */
     Iterable<ConfigSource> getConfigSources();
+
+    /**
+     * Return the {@link Converter} used by this instance to produce instances of the specified type from string values.
+     *
+     * @param <T>
+     *             the conversion type
+     * @param forType
+     *             the type to be produced by the converter
+     * @return an {@link Optional} containing the converter, or empty if no converter is available for the specified type
+     */
+    <T> Optional<Converter<T>> getConverter(Class<T> forType);
 }
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
@@ -91,6 +91,14 @@ public class ArrayConverterTest extends Arquillian {
     }
 
     @Test
+    public void testGetBooleanArrayConverter() {
+        Boolean[] value = config.getConverter(Boolean[].class).get().convert("true,false,true");
+        Assert.assertNotNull(value);
+        Assert.assertEquals(value.length, 3);
+        Assert.assertEquals(value, new Boolean[]{true, false, true});
+    }
+
+    @Test
     public void testOptionalBooleanLookupProgrammatically() {
         Optional<Boolean[]> optionalValue = config.getOptionalValue("tck.config.test.javaconfig.converter.booleanvalues",
             Boolean[].class);
@@ -107,7 +115,15 @@ public class ArrayConverterTest extends Arquillian {
         Assert.assertEquals(converterBean.getMyBooleans(), new Boolean[]{true, false, true});
     }
 
-    //test bool[] support
+    //test boolean[] support
+
+    @Test
+    public void testGetbooleanArrayConverter() {
+        boolean[] value = config.getConverter(boolean[].class).get().convert("true,false,true");
+        Assert.assertNotNull(value);
+        Assert.assertEquals(value.length, 3);
+        Assert.assertEquals(value, new Boolean[]{true, false, true});
+    }
 
     @Test
     public void testbooleanArrayInjection() {
@@ -133,6 +149,14 @@ public class ArrayConverterTest extends Arquillian {
     public void testStringLookupProgrammatically() {
         String[] value = config.getValue("tck.config.test.javaconfig.converter.stringvalues",
             String[].class);
+        Assert.assertNotNull(value);
+        Assert.assertEquals(value.length, 4);
+        Assert.assertEquals(value, new String[]{"microservice", "microprofile", "m,f", "microservice"});
+    }
+
+    @Test
+    public void testGetStringArrayConverter() {
+        String[] value = config.getConverter(String[].class).get().convert("microservice,microprofile,m\\,f,microservice");
         Assert.assertNotNull(value);
         Assert.assertEquals(value.length, 4);
         Assert.assertEquals(value, new String[]{"microservice", "microprofile", "m,f", "microservice"});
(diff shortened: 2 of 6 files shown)


api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -133,4 +133,18 @@
...
+     * @param <T>
+     *             the conversion type
+     * @param value
+     *             the value to convert
+     * @param type
+     *             the type into which the value should be converted
+     * @return the converted value
+     * @throws java.lang.IllegalArgumentException if the value cannot be converted to the specified type.
+     */
+    <T> T convert(String value, Class<T> type);
```
Emily-Jiang: I don't think this functionality should belong to Config. See my [comments](https://github.com/eclipse/microprofile-config/issues/492#issuecomment-575568151)

dmlloyd: See also [my disagreement](https://github.com/eclipse/microprofile-config/issues/492#issuecomment-575615818).  The change should be accepted.

emattheis:
@Emily-Jiang in light of #514 can we move forward with this? I think we're in agreement that an external converter spec has merit, and Config can adopt it in the future, but having access to the converters within the context of a Config is useful regardless of what spec they follow.

Emily-Jiang:
@emattheis please see my explaination on the issue.

genie-microprofile:

dmlloyd:
@emattheis would you mind rebasing this?  In particular I think the OSGi version thing probably can be dropped (we should update that separately for 2.0 anyway).

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -144,4 +145,33 @@
...
+     *             the conversion type
+     * @param value
+     *             the value to convert
+     * @param type
+     *             the type into which the value should be converted
+     * @return the converted value
+     * @throws java.lang.IllegalArgumentException if the value cannot be converted to the specified type.
+     */
+    default <T> T convert(String value, Class<T> type) {        
+        return value == null ? null
```
radcortez: I think we shouldn't return null directly if the value is null. How about if I want to have a converter that picks up nulls and convert them to a value of my choice?

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -144,4 +145,33 @@
...
+    /**
+     * Return the {@link Converter} used by this instance to produce instances of the specified type from string values. 
+     * 
+     * @param <T>
+     *             the conversion type
+     * @param forType
+     *             the type to be produced by the converter
+     * @return an {@link Optional} containing the converter, or empty if no converter is available for the specified type
+     */
+    <T> Optional<Converter<T>> getConverter(Class<T> forType);
```
radcortez: Shouldn't we also expose a `<T> Converter<T> getConverter`?

api/src/main/java/org/eclipse/microprofile/config/Config.java
dmlloyd: Converters should only return `null` if the string reflected an empty value.  Converters should never be passed `null` as an argument.

radcortez: In that case, shouldn't we adjust the javadoc of org.eclipse.microprofile.config.spi.Converter#convert and advise to throw an IllegalArgumentException if the value parameter is null?

api/src/main/java/org/eclipse/microprofile/config/Config.java
dmlloyd: This was discussed previously on a call I think, and the consensus was that the optional variant is more consistent with the other API elements.  It allows the caller to decide whether to throw an exception, return null, or stick with the `Optional` if there is no converter available, all with one method.

api/src/main/java/org/eclipse/microprofile/config/Config.java
dmlloyd: We could do so.

emattheis: Per our conversation several weeks back, I'm inclined to remove this method. The addition of `getConverter` can solve this use case and more elegantly handles the scenario where no suitable converter is available. It would also confine null handling to the converter.

dmlloyd: Sure go ahead and remove it; also could you please squash the commits into one?

radcortez: Agree. Remove and clarify javadoc in org.eclipse.microprofile.config.spi.Converter#convert.

dmlloyd: The PR/commit title should be updated accordingly as well.

emattheis: As far as disallowing `null` as an argument to `Converter::convert`, do we prefer `NullPointerException` or `IllegalArgumentException`? I think NPE is more conventional - e.g. `Objects::requireNonNull` - even if the latter is more precise.

emattheis: I spun up #539 so that decision doesn't get hidden in this PR. I think we should track it separately.

dmlloyd:
@emattheis not to continue to harp on you :) but would you mind reducing this down to a single commit which adds the new method?  I don't think you need to update the release notes per se (at least I don't think we've discussed that as a group, that I recall anyway).  But it shouldn't be 8 commits.

Alternatively you might want to check the "[ ] Allow edits from maintainers" box, and I can do it, if you don't have availability.

Thanks!

emattheis:
@dmlloyd I meant to ask about that in the hangout. In my day-job, we use the "[] Allow squash merging" setting so we can do that when we merge PRs into the mainline. That way we preserve individual commits in the feature branches. Not sure if that's enabled on this repo, though.

Allow edits from maintainers is enabled, so feel free to squash on this branch if you prefer.

dmlloyd:
All fixed up.  @Emily-Jiang can you please re-review or dismiss your review?

dmlloyd:
This needs a rebase after #544 is merged in order to pass CI.

radcortez:
Do we have "Require branches to be up to date before merging" on?

dmlloyd:
I don't think so.  Might be a good idea though?

CI is failing because of #544 but that's an easy fix (Erik already did it for us in fact).

radcortez:
Yes, I advice to turn it on. I can't do it.

dmlloyd:
I think it would need some discussion (but anyway I don't have permission either).  Open a process change issue?

radcortez:
Sure.

emattheis:
I updated the TCK with coverage for `getConverter` and used a local build of smallrye-config to verify. I'll squash the commits again once someone takes a look.

radcortez:
We do need https://github.com/eclipse/microprofile-config/pull/544 to be merged so we can pass the CI build.

dmlloyd:
It looks as though CI is presently testing with the branch head rather than a merge of the commit.  Looking into it.

emattheis:
want me to just rebase now that #544 is merged?

dmlloyd:
I'm going to try to fix the action first.  If that fails, then yes we'll need to rebase all of these.

dmlloyd:
Actually it seems that the settings from the branch are used for Actions, so even if I fix CI upstream the branch will still break.  So you'll still have to rebase, but let's still wait until #547 goes in, just so we can get the extra validation for it.

dmlloyd:
@emattheis OK please go ahead and rebase!  Thanks.

dmlloyd:
@Emily-Jiang  this one is also green now; can you please re-review or dismiss your old review?

Issue #496, opened 2020-01-17
add convenience methods for getting multi-valued properties


emattheis:
The existing `Config` API only permits retrieving multi-valued properties as arrays like this:

```java
for (String value : config.getValue("my.key", String[].class)) {
    // do something with value
}
```

I propose we add the following methods to permit retrieving more than one value in a more natural way:

```java
<T> Iterable<T> getValues(String propertyName, Class<T> propertyType);
    
<T> Optional<Iterable<T>> getOptionalValues(String propertyName, Class<T> propertyType);
```

This would enable the use of lamba-friendly patterns like this:

```java
config.getValues("my.key", String.class)
      .forEach(value -> {
            // do something with value
      });
```

As a bonus, the developer doesn't need to worry about asking for `String.class` or `String[].class` depending on the scenario.

kenfinnigan:
Not sure I agree with your final comment around bonus, either way, you need to know whether you're dealing with a single value or array of values.

I've heard that the use of lambdas instead of `forEach()` is not better for performance and marginal on readability, so not sure about the benefits of this.

Curious as to what others think

emattheis:
My last comment wasn't meant to imply that a developer doesn't need to know whether she is dealing with multiple values, just that she needn't concern herself with creating the actual array class. Selecting the first element of a list of one or more is a fairly common idiom, and if you want to layer something on top of microprofile-config to abstract away the difference between a single value or many values of an arbitrary type you have to deal with this: https://stackoverflow.com/questions/4901128/obtaining-the-array-class-of-a-component-type.

Debates about whether lambdas are "better" in the readability and performance departments aside, functional programming is alive and well on the Java platform, so I think we ought to at least reduce any friction there.

Even with traditional Java programming techniques, the new methods are still valuable because implementations are not restricted to constructing arrays. By using `Iterable<T>`, this is still an option:

```java
for (String value : config.getValues("my.key", String.class)) {
    // do something with value
}
```

and the implementation is free to use any iterable under the covers.

dmlloyd:
In SmallRye we take the approach of allowing the user to give a generator which could yield (for example) a List, Set, or Sorted/NavigableSet.  We could likely use that to implement a feature like this without too much trouble.  That said, it might be worth deciding about config accessors before we add more methods to `Config` itself...

emattheis:
I did see the SmallRye approach, but felt it might be overcomplicated for the basic need, but I'm not opposed to it as an overloaded option.

[...]

PR #501, opened 2020-01-23
add multi-value getters to Config

emattheis:
as proposed in #496

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -34,6 +34,9 @@
 
 package org.eclipse.microprofile.config;
 
+import java.lang.reflect.Array;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
@@ -109,6 +112,30 @@ public interface Config {
     <T> T getValue(String propertyName, Class<T> propertyType);
 
     /**
+     * Return the resolved property values with the specified type for the
+     * specified property name from the underlying {@linkplain ConfigSource configuration sources}.
+     * <p>
+     * The configuration values are not guaranteed to be cached by the implementation, and may be expensive
+     * to compute; therefore, if the returned values are intended to be frequently used, callers should consider storing
+     * rather than recomputing them.
+     *
+     * @param <T>
+     *             The property type
+     * @param propertyName
+     *             The configuration property name
+     * @param propertyType
+     *             The type into which the resolved property values should get converted
+     * @return the resolved property values as a list of instances of the requested type
+     * @throws java.lang.IllegalArgumentException if the property values cannot be converted to the specified type
+     * @throws java.util.NoSuchElementException if the property isn't present in the configuration
+     */
+    default <T> List<T> getValues(String propertyName, Class<T> propertyType) {
+        @SuppressWarnings("unchecked")
+        Class<T[]> arrayType = (Class<T[]>) Array.newInstance(propertyType, 0).getClass();
+        return Arrays.asList(getValue(propertyName, arrayType));
+    }
+
+    /**
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@linkplain ConfigSource configuration sources}.
      * <p>
@@ -129,6 +156,30 @@ public interface Config {
      * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
      */
     <T> Optional<T> getOptionalValue(String propertyName, Class<T> propertyType);
+
+    /**
+     * Return the resolved property values with the specified type for the
+     * specified property name from the underlying {@linkplain ConfigSource configuration sources}.
+     * <p>
+     * The configuration values are not guaranteed to be cached by the implementation, and may be expensive
+     * to compute; therefore, if the returned values are intended to be frequently used, callers should consider storing
+     * rather than recomputing them.
+     *
+     * @param <T>
+     *             The property type
+     * @param propertyName
+     *             The configuration property name
+     * @param propertyType
+     *             The type into which the resolved property values should be converted
+     * @return The resolved property values as an {@code Optional} wrapping a list of the requested type
+     *
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     */
+    default <T> Optional<List<T>> getOptionalValues(String propertyName, Class<T> propertyType) {
+        @SuppressWarnings("unchecked")
+        Class<T[]> arrayType = (Class<T[]>) Array.newInstance(propertyType, 0).getClass();
+        return getOptionalValue(propertyName, arrayType).map(Arrays::asList);
+    }
 
     /**
      * Returns a sequence of configuration property names. The order of the returned property names is unspecified.
--- a/spec/src/main/asciidoc/release_notes.asciidoc
+++ b/spec/src/main/asciidoc/release_notes.asciidoc
@@ -130,3 +130,18 @@ A full list of changes delivered in the 1.4 release can be found at link:https:/
 
 - Exclude EL api transitive dependency (link:https://github.com/eclipse/microprofile-config/issues/440[#440])
 - Other minor spec wording or JavaDoc updates
+
+[[release_notes_20]]
+== Release Notes for MicroProfile Config 2.0
+
+A full list of changes delivered in the 2.0 release can be found at link:https://github.com/eclipse/microprofile-config/milestone/8?closed=1[MicroProfile Config 2.0 Milestone].
+
+=== API/SPI Changes
+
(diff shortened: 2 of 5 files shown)


dmlloyd:
This needs a rebase.

emattheis:
rebased against master

emattheis:
@Emily-Jiang I think the main value here is making it clear at a glance that we support multi-valued properties. As a developer, if you only see `Config::getValue` and `Config::getOptionalValue` it is not obvious that such support exists. I have seen code where the developer is getting a string and splitting on commas simply because they did not know that config supported multi-value properties and didn't bother to dig deeper. I'm confident those same developers would have understood immediately with the proposed method additions.

Type-safe, multi-valued properties are a GREAT feature of MicroProfile Config and deserve special treatment. I'm personally okay with deferring this until other accessor issues are tackled, but I think it's low risk to add it now. I'd be happy to take a stab at the TCK as well to provide coverage.

genie-microprofile:

emattheis:
added test coverage to the TCK, but haven't checked with an impl yet

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -129,6 +156,30 @@
...
+     *             The property type
+     * @param propertyName
+     *             The configuration property name
+     * @param propertyType
+     *             The type into which the resolved property values should be converted
+     * @return The resolved property values as an {@code Optional} wrapping a list of the requested type
+     *
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     */
+    default <T> Optional<List<T>> getOptionalValues(String propertyName, Class<T> propertyType) {
```
radcortez: I believe we must clarify https://github.com/eclipse/microprofile-config/issues/446 first before exposing a return type of `Optional<List<T>>`. I'm usually not a big fan of exposing containers inside containers.

emattheis: `Optional<T[]>` is already possible today, so I don't think this should be blocked by #446.

radcortez: What if we drop these specific multi value methods and introduce a GenericType as an argument, like you have with `readEntity` in JAX-RS?

Then we could just expose a `getValue/getOptionalValue(String propertyName, GenericType type) and you can call it like this:

`.getValue("something", new GenericType<List<String>>() {});`

This will allow you to just overload the methods, instead of adding new ones with the same method parameters and support all the other collection types, instead of just List.

dmlloyd: Is that better than a SmallRye-style `getValue("something", new ListConverter<String>())`?  Anonymous class type tokens increase metaspace usage...

dmlloyd: Is that better than a SmallRye-style `getValue("something", new ListConverter<String>())`?  Anonymous class type tokens increase metaspace usage...

radcortez: No, but in that case you are wrapping your `List` in the Converter type and works fine in the method signature with `.getValue(String, Converter)`.

The proposed method is a `.getValues(String, Class)`. First, since the signature is already take, we need to rename the method and then, because we are not able to pass a Collection type, we are always limited by the return type.

I'm fine if we expose a `<T> T getValue(String, Converter<T>)`. This would cover any collection type if you provider a Converter for it. Not so sure about a `<T> List<T> getValues(String, Class<T>)` or the Optional variant.

radcortez: Or maybe we need to think about this in a more integrated way, instead of looking into each issue separately.

Maybe if we implement the ConfigAccessor feature we won't need to add this here and the ConfigAccessor would be responsible to handle all of these cases. Something like, `ConfigAccessor#getValueAsList`?

dmlloyd: Yeah no answer seems ideal.  Maybe delaying until we have accessors is better.  Or maybe we should just add a documentation line like:

```
If you want a {@code List} of the given type, use: {@code Arrays.asList(config.getValue(key, Item[].class))}.
```

Or something like that.

dmlloyd: The type token idea has some merit though, I don't mean to dismiss it.  Once nice thing about it is that it gives you parity with CDI.  If I have registered a converter which implements `Converter<Thing<Foo, Bar<Baz>>`, then I could theoretically inject a `Thing<Foo, Bar<Baz>>` in to CDI, but getting such a value using the API is more difficult.  Using the type token approach makes this possible to do with the API.

radcortez: I confess that I didn't thought about that use case. Might be interesting as well, indeed :)

The idea was mostly to make the return type more flexible, instead of being locked down in List.

emattheis: We should really have this discussion in the issue, not the PR, and we _really_ should have had it before I updated the TCK test case 😛

I can certainly sympathize with the issues the name collision may cause with the SmallRye implementation, but didn't we already basically agree that exposing `List<T>` made the most sense?

I'm all for coming up with a more flexible way to get access to multivalued variables, but I think it would be a shame to release 2.0 without an obvious method signature that exposes multivalued properties. The `T[].class` mechanism is really clunky, IMHO.

radcortez: Sorry, I guess we didn't thought well about this :(

The point here is that we need to be extra careful with that we put in there. We could break compatibility in the future if we are not happy with how it turned out, but I believe that once we code something here it will stay forever.

emattheis: At this rate, we won't have to worry about compatibility because we will never change anything.

The premise of micro profile was to move fast and not be paralyzed by "big spec" thinking. We should not be afraid to throw this out when/if accessors come along.

The fact that SmallRye has so much more functionality in SmallRyeConfig than the spec is surely a sign that we're not moving fast enough here.

As an individual contributor in my spare time, I don't want to waste effort on something we're not aligned on. At this point, I would respectfully ask that the committers decide whether to accept or reject this PR so we can move forward.

dmlloyd: I'm OK with this change as it stands; I think we can improve things later without introducing any backwards compatibility issues.  TBH I don't see a problem with having many useful methods and overloads on this interface.  I think that if you can do more with less, that's great, but it doesn't cost much to do more with more either.  If we do introduce type tokens, these default methods can be efficiently implemented in terms of that solution (as can `getOptionalValue` for that matter).  It really doesn't matter that much.

So my ✅review stands: I'm happy with this as it is.  Even in the absolute worst case where we come up with something better immediately, it doesn't cost us much to have these methods - we already have far more than these in SmallRye Config!

radcortez: Well, I'm sorry. My intention was not to block anything. My concern here was about having a solid and generic API. For that, we don't need to use that much time or be paralysed.

One thing is what SmallRye does on its own. Another thing is what the API does. And BTW, SmallRye does expose it this way:

`public <T, C extends Collection<T>> Optional<C> getOptionalValues(String name, Class<T> itemClass, IntFunction<C> collectionFactory)`

At least, you are not constrained to use only List like the proposal and you can even pass in your own constructed collection. Might as well use it this way.

emattheis: @radcortez this was discussed [in the issue](https://github.com/eclipse/microprofile-config/issues/496#issuecomment-575898716). I'm not necessarily opposed to the SmallRye approach, but I would argue that `List<T>` is the most common/obvious usage. Per @dmlloyd's comments, there's no reason we can't eventually have both. Honestly, I'd say we should add both right now, but I think we'll get pushback in the next hangout 😛

radcortez: If everyone else is happy with the change, I'm not going to push it back, so please move forward.

emattheis:
I tested the additional TCK methods using a local build of the smallrye-config master, and it passed without issue.

radcortez:
We need this one first https://github.com/eclipse/microprofile-config/pull/544 so we can pass the CI build.

dmlloyd:
This one also needs a rebase now.

dmlloyd:
OK all green now.  @Emily-Jiang it's up to you now; can you please re-review or dismiss your review?

emattheis:
Since this PR just introduces an alternative to the existing array support, I simply duplicated the existing tests for the new methods. I could certainly add coverage for single-valued arrays, but that is not something that is currently covered by the TCK for array support.

I think if we want to explicitly test for that scenario, we should do so across all the existing multivalued coverage and that will be a bit tedious. Not sure it should hold up this PR...

emattheis:
@Emily-Jiang I went ahead and added comprehensive coverage for single-value array/list/set in the TCK

dmlloyd:
This still waits for @Emily-Jiang.

Issue #371, opened 2018-06-08
Is the return value of Config#getConfigSources() specified to be immutable and unchanging?


ljnelson:
In #370 @struberg notes that "Config infra is basically immutable".  Is there a note somewhere in the specification that indicates that, for example, `Config#getConfigSources()` must return an unchanging `Set` of `ConfigSource`s when invoked?  Should there be?

Similarly, what is the contract for [`ConfigSourceProvider#getConfigSources(ClassLoader)`](https://static.javadoc.io/org.eclipse.microprofile.config/microprofile-config-api/1.2.1/org/eclipse/microprofile/config/spi/ConfigSourceProvider.html#getConfigSources-java.lang.ClassLoader-)?  Is it required or prohibited or undefined for a `ConfigSourceProvider` implementation to return a dynamic `Set`?

struberg:
The specification describes all the setup and discovery which happens at startup. And that's it. 
So from that you can assume that those parts are not intended to get altered later.

ljnelson:
Well, the specification doesn't actually say anything about how `ConfigSource`s and `Config` are linked together.  As you say, I _assume_ that once `ConfigSource`s are discovered by the service provider mechanism, then they're supposed to be used by `Config` and should? must? be returned by `Config#getConfigSources()` but I didn't see anything that says they must be.

Also, the specification [has a section on dynamic `ConfigSource` providers](https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc#custom-configsources-via-configsourceprovider).  I don't think I can make any assumptions about what they can do.  Or are you saying, yes, I can, they must be called once and must produce immutable unchanging `Set`s themselves?

struberg:
Indeed that might be good to clarify.

Emily-Jiang:
@struberg and I talked this a bit more in today's hangout. We think it is best left unspecified so that implementations can have freedom to decide whether they want to support dynamically reloaded config sources.

ljnelson:
Thanks; in that case, I suppose that it must be specified that the return value must be safe for concurrent use by multiple threads?  What must I synchronize on when iterating over it?  Itself?

Emily-Jiang:
@ljnelson it is best for you to join our MP Config weekly hangout to discuss your use case. Can you join next week? Please let me know on gitter if you need help with the meeting calendar.

ljnelson:
I'll try to make it.

[...]

PR #510, opened 2020-01-24
#371 improve javadoc on getConfigSources(), also fixed the checkstyle…

Emily-Jiang:
… error introduced be a previous PR

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -139,7 +139,7 @@ public interface Config {
 
     /**
      * Return all of the currently registered {@linkplain ConfigSource configuration sources} for this configuration.
-     * The returned sources will be sorted by descending ordinal value and name.
+     * The returned sources will be sorted by descending ordinal value and name, which can be iterated in a thread-safe manner.
      *
      * @return the configuration sources
      */
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
@@ -53,7 +53,8 @@ import java.util.Set;
  * <ol>
  * <li>{@linkplain System#getProperties() System properties}, with an {@linkplain #getOrdinal() ordinal value} of {@code 400}</li>
  * <li>{@linkplain System#getenv() Environment properties}, with an ordinal value of {@code 300}</li>
- * <li>The {@code /META-INF/microprofile-config.properties} {@linkplain ClassLoader#getResource(String) resource}, with an ordinal value of {@code 100}</li>
+ * <li>The {@code /META-INF/microprofile-config.properties} {@linkplain ClassLoader#getResource(String) resource}, 
+ * with an ordinal value of {@code 100}</li>
  * </ol>
  *
  * <h3>Environment variable name mapping rules</h3>


api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -139,7 +139,7 @@
 
     /**
      * Return all of the currently registered {@linkplain ConfigSource configuration sources} for this configuration.
-     * The returned sources will be sorted by descending ordinal value and name.
+     * The returned sources will be sorted by descending ordinal value and name, which can be iterated in a thread-safe manner.
```
ljnelson: This necessarily means that for any given implementation either (a) the returned sources must be a snapshot-in-time (i.e. not changing "underneath" you as you iterate) or (b) the returned sources may only grow, not shrink.  In either case the assumed behavior must be made specific in the specification language.  I think this (still) relates to #371: is the return value and behavior of this method intended to describe a set of objects whose cardinality is unchanging, or is it intended to describe a set of objects whose cardinality can only grow, not shrink?

kenfinnigan: I think I'd lean towards unchanging/immutable.

Not sure what @Emily-Jiang, @radcortez, or @dmlloyd think?

It should be easy enough to get a minor wording change in before 1.4

ljnelson: Unchanging or immutable can have two meanings here and while I know I sound pedantic I really, really, really, really think this is important:

1. Unchanging with respect to cardinality, but not with respect to content.  (At time 0, the returned set has three elements in it.  They are A, B, and C.  At time 1, the returned set continues to have three elements in it (its cardinality is unchanging).  However now it has A, D and Q.)
2. Unchanging with respect to cardinality _and_ content.  (At times 0 and 1 and _n_, the returned set has three elements in it which are A, B and C.)

(There's even more nuance here about whether A, B and C can have changing properties of their own or not; I'm tabling that for now, but that's _also_ important.)

kenfinnigan: Completely fair and understandable.

I'd lean towards option 2, as I think we want to promote the notion that once a `Config` is created it's not possible to add additional `ConfigSource` instances to change the config content. For that, you need to recreate `Config`, but I could be reading too much into the spec there.

ljnelson: Naïvely I'd assume option 2 was intended as well, but [there is a section on dynamic `ConfigSource`s](https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc#custom-configsources-via-configsourceprovider) that _also_ doesn't say anything one way or another about the mutability of cardinality or content of `ConfigSource`s provided in this manner.  I'm still feeling like there are a lot of assumptions and implications in a set of documents that should be much more prescriptive.  I understand this is an iterative process but I want to ensure that these assumptions and implications remain in the spotlight as the effort moves forward.

kenfinnigan: Ah yeah, I think we tried to cleanup the dynamic config source issue but didn't succeed.

And I appreciate you continuing to raise them as they need to be resolved

radcortez: I believe the initial thought was to alway have an unchanging set of sources both for content and cardinality. At least, this is what I would expect, since you are not able to add additional ConfigSources after a Config is created.

I think that the ability to dynamically reload or change ConfigSources should not be left unspecified for the implementor to choose. This is to big of a decision and would only cause confusion how this works between users.

ljnelson: Respectfully, you believe and think this, and Ken assumes it, and Emily and Mark indicated at one point that it shouldn't be specified, which means that it actually would be possible…and I'm pointing out with this comment thread that for reasons enumerated above this behavior _must_ be _explicitly_ specified, not believed, assumed, indicated or thought.  So I personally agree with you but my point is none of this is spelled out, and it must be.

radcortez: Ah, but I completely agree with you. It should be specified one way or another.

Issue #475, opened 2019-11-19
CDI @ConfigProperty default value consistency should be enforced


dmlloyd:
If the same configuration property is given more than once via the `@ConfigProperty` annotation, they should be enforced to have the same default value string, or else an error should result.

This is a necessary prerequisite to #405; if multiple default values are allowed for a given property, then it becomes impossible to define which value would be returned for an expression string which references that property.

The specification text for `@ConfigProperty` should be updated to this effect.

dmlloyd:
An alternative is to explicitly specify that default values given in `@ConfigProperty` are only scoped to that access and will not appear in variable expansions if/when #405 is ratified.

dmlloyd:
One challenge here is that the only way that equivalence can be enforced would be to cause the container to fail to deploy the application.  But we don't really have any way to test that this happens in the TCK.  Should we go ahead with the spec change, with wording like "The container *may* cause the application to fail" or something like that?

Azquelt:
I don't think we should do this, primarily because I don't think that an annotation on one injection point should affect a value injected at another injection point.

```java
@Inject @ConfigProperty(value="foo", defaultValue="bar")
private String baz;
```
should be seen as the CDI equivalent of
```java
String baz = Config.getConfig().getOptionalValue("foo", String.class).orElse("bar");
```

If you need to set the default for a config value across your application, this can be done in the `microprofile-config.properties` file within the application.

radcortez:
Yes, that works and aligns with the proposal of default values being scoped to the injection point, but then these should be excluded from property expansion. Consider:

```java
@Inject
@ConfigProperty(value="foo", defaultValue="bar")
private String bar
@Inject
@ConfigProperty(value="foo", defaultValue="baz")
private String baz
@Inject
@ConfigProperty(value="${foo}", defaultValue="baz")
private String expand

Config.getConfig().getOptionalValue("expand", String.class);
```

Which one gets expanded into `expand`?

Azquelt:
Assuming "foo" is not defined, neither.

Azquelt:
If you want to assign a default value to `foo` so that it gets expanded, doing this makes more sense:

```
@Inject
@ConfigProperty(value="foo")
private String bar

@Inject
@ConfigProperty(value="foo")
private String baz

foo=bar
expand=${foo}
Config.getConfig().getOptionalValue("expand", String.class);
```

Azquelt:
[...]

Issue #365, opened 2018-05-30
Null must be injectable


rmannibucau:
@Inject
    @ConfigProperty(name="...")
    String value;

It must be allowed to inject that and get null, a workaround is to use an optional but it is not recommanded (even the opposite) to use optionals as fields so this is not a solution. Injecting null is fine from all point of views and compatible with mainstream config solution so it should probably make its path in mp-config asap.

Emily-Jiang:
Do you mean you defined 
`myConfig=`

in config source, you will get empty string instead of null. Do you want to get null instead? If not, can you elaborate your use case?

struberg:
or `optional=true`

rmannibucau:
sounds weird with Optional class no: "optional=true will set null and you can inject Optional<String> which will not be null" :s? But agree a better phrasing can be found.

Emily-Jiang:
-1 on optional=true, as it contradicts with the contract of "the presence of defaultValue" means optional. 

How about to add a new constant on ConfigProperty?
```
String NULL_VALUE="org.eclipse.microprofile.config.configproperty.nullvalue";

@Inject
@ConfigProperty(name="..." defaultValue=ConfigProperty.NULL_VALUE)
String value;
```

rmannibucau:
I had that in mind as well so +1

dmlloyd:
The currently proposed solution does not seem correct.  One alternative discussed in #512 was to add a separate `nullable` annotation property indicating that the injected field should be treated as if it were optional with an `orElse(null)` qualification.  This property would not be allowed to be `true` if the injection target is not nullable (that is, it is primitive).

radcortez:
+1

rmannibucau:
Hmm, why isnt it correct? Having a flag or a flag is the same. One is likely more elegant than the other but both are strictly equivalent.
Now, for the primitive case, you must consider the magic string NULL_VALUE means "the configured value is null but passthrough" so the output takes its default.
Overall point is that duplicating this notion just adds noise to the spec.
Note, however, this is already outside the spec - like geronimo proxy support - since in practise using a default of NULL_VALUE is only valid in cdi injections so not primitives so we are safe anyway.

dmlloyd:
> Hmm, why isnt it correct? Having a flag or a flag is the same. One is likely more elegant than the other but both are strictly equivalent.

[...]

PR #512, opened 2020-01-27
Revert #388 pending further discussion

dmlloyd:
Revert #388.  This methodology of injecting `null` introduces inconsistencies with the base API:

* It becomes impossible to implement injectable default values with a config source
* It becomes impossible to support a default value with a nullable property
* It introduces another problem case with respect to empty values
* It introduces a problem where injection produces a different result that cannot be replicated on the base API

We need to solve this problem another way. `Optional` is the preferred mechanism, but if we must support nullability on injection, then we should have a separate annotation or annotation property which indicates that the value is optional and there should be an implicit `.orElse(null)`.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java
@@ -103,16 +103,6 @@ import javax.inject.Qualifier;
 public @interface ConfigProperty {
     String UNCONFIGURED_VALUE="org.eclipse.microprofile.config.configproperty.unconfigureddvalue";
     /**
-     * Provide a way to specify {@code null} value for a property.
-     * e.g. The following example is to set the default value of {@code my.port} to null if the property is not specified in any config sources.
-     * <pre>
-     * &#064;Inject
-     * &#064;ConfigProperty(name="my.port" defaultValue=ConfigProperty.NULL_VALUE)
-     * String value;
-     * </pre>
-     */
-    String NULL_VALUE="org.eclipse.microprofile.config.configproperty.nullvalue";
-    /**
      * The key of the config property used to look up the configuration value.
      * If it is not specified, it will be derived automatically as {@code <class_name>.<injection_point_name>},
      * where {@code injection_point_name} is the field name or parameter name,
--- a/api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
@@ -44,6 +44,6 @@
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  * 
  */
-@org.osgi.annotation.versioning.Version("1.1.0")
+@org.osgi.annotation.versioning.Version("1.0")
 package org.eclipse.microprofile.config.inject;
 
--- a/spec/src/main/asciidoc/configexamples.asciidoc
+++ b/spec/src/main/asciidoc/configexamples.asciidoc
@@ -96,11 +96,6 @@ public class InjectedConfigUsageSample {
     @Inject
     @ConfigProperty(name="myprj.some.dynamic.timeout", defaultValue="100")
     private javax.inject.Provider<Long> timeout;
-    //Injects the value of the property myprj.name if specified in any of the configures, otherwise null will be injected. 
-    @Inject
-    @ConfigProperty(name="myprj.name" defaultValue=ConfigProperty.NULL_VALUE)
-    String name;
-    
     //The following code injects an Array, List or Set for the `myPets` property,
     //where its value is a comma separated value ( myPets=dog,cat,dog\\,cat)
     @Inject @ConfigProperty(name="myPets") private String[] myArrayPets;
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPlainInjectionTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPlainInjectionTest.java
@@ -46,8 +46,6 @@ import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.testng.annotations.BeforeTest;
 import org.testng.annotations.Test;
 
-import junit.framework.Assert;
-
 /**
  * Test cases for CDI-based API that test retrieving values from the configuration.
  * The tests depend only on CDI 1.2.
@@ -97,8 +95,6 @@ public class CDIPlainInjectionTest extends Arquillian {
         assertThat(bean.characterProperty, is(equalTo(Character.valueOf('c'))));
 
         assertThat(bean.doublePropertyWithDefaultValue, is(closeTo(3.1415, 0.1)));
-        Assert.assertNull("The property my.not.configured.nullable.property should be null",
-                 ConfigProvider.getConfig().getOptionalValue("my.not.configured.nullable.property", String.class).orElse(null));
     }
 
     /*
@@ -130,8 +126,6 @@ public class CDIPlainInjectionTest extends Arquillian {
         assertThat(bean.doublePropertyWithDefaultValue, is(closeTo(
                 ConfigProvider.getConfig().getOptionalValue("my.not.configured.double.property", Double.class)
                         .orElse(3.1415), 0.1)));
-        Assert.assertNull("The injected field nullableConfigValue is not null", bean.nullableConfigValue);
-
     }
 
     @Test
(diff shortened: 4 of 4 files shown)


Emily-Jiang:
I don't understand what you meant. Please provide an example. By the way, Open Liberty implemented this without any issues.

dmlloyd:
I've explained clearly and concisely.  Let me pose a question: if someone wanted to implement default values using a configuration source, what value would they use for `null`?

dmlloyd:
> By the way, Open Liberty implemented this without any issues.

This is a nonsense argument.  If there was a problem implementing it, I would have said so.  The problem has nothing to do with implementation.  The problem is as I explained: it cuts off solutions to things we already know need to be solved.

If we are going to implement property expansion, for example, then every default value has to be consistent (#475) and representable within a configuration source.  Adding special sentinel values which only apply to CDI causes a layering problem (CDI layers above the base API).  What happens when you get such a default value, represented within a config source, from the base API?  You get something weird or unspecified.

This change was not well-thought-out.  Please *carefully* re-read my comments.  I believe I've explained clearly the issues.

dmlloyd:
> I don't understand what you meant.

Please read again more carefully!  I believe my phrasing is fairly exact.

struberg:
-1 on the revert as well

struberg:
> I've explained clearly and concisely. Let me pose a question: if someone wanted to implement default values using a configuration source, what value would they use for `null`?

ConfigSources are not meant to provide default values. As simple as that. That's like defining the database should provide an own logic to serve values if a column entry is null.

dmlloyd:
> > I've explained clearly and concisely. Let me pose a question: if someone wanted to implement default values using a configuration source, what value would they use for `null`?
> 
> ConfigSources are not meant to provide default values. As simple as that. That's like defining the database should provide an own logic to serve values if a column entry is null.

I do not think this comparison makes any sense; if you're going to draw an analogy then please explain it in detail.  Let's say you implement a config source using a database (bringing this analogy as close as possible to the actual case).  You would not use the database for the default values; they would be a distinct config source with a lower priority.

We in fact do use config sources today to provide default values in this way; in fact, it is *necessary* to robustly support property expansion (#405) because otherwise expressions cannot reference properties with default values with any expectation that the expansion will work correctly (and we learned almost immediately that this is something they will expect).  It works well in practice also.  I believe we discussed this on a call a while back, and you indicated that you understand this and I believe you agreed that this is in fact the best way to solve the problem of property expansion with default values.

To me it looks like the implementation choice of using a sentinel default value is being used to justify the design decision.  The requirement (by my understanding) is that the user requested a way for a configuration property to be optional without using `Optional`.  There is nothing about that requirement that indicates that using `defaultValue` for this purpose is necessary.

Given that the chosen solution is not necessary to solve the initial requirement, and that the chosen solution has clear problems, it should be reverted and solved another way.

radcortez:
Original issue here: https://github.com/eclipse/microprofile-config/issues/365

I didn't want to raise any issue at first, since I thought this has already been discussed and accepted.

I do find the use of a sentinel value in default property a little weird, especially because it is a String value and it forces you to look into the API to know about it, or else you will never know it is there. Apart for any other considerations, if we are revisiting this, I would prefer to add an additional field to `ConfigProperty` like `nullable` to represent this. I believe this will be clear in terms of API and also address @dmlloyd concerns.

dmlloyd:
A weakness of the original approach can be shown quite simply: how, with the original approach, does the user specify that a property may be nullable, but by default it is not `null`?  Answer: they cannot.

Emily-Jiang:
> This is a nonsense argument.
@dmlloyd I found the above comment is hard to take. Is it ok to discuss the issue politely?

emattheis:
Personally, I would like to avoid the whole issue of indicating nullability at the injection point by simply making that the default behavior. The current behavior assumes that the configuration a CDI bean relies on will be stable at CDI initialization time and thus it is appropriate to fail initialization by default when a property is missing. In reality, bean instances will be constructed depending on their scopes and the presence of a relied upon property cannot be guaranteed by a config check when the container is initialized.

When a property is missing, we should simply inject a null value unless a default is provided at the injection point.

Further, if we already assume that an empty string is the same as a null value, we can remove the existing NOT_DEFINED sentinel string and just use an empty string as the default for defaultValue. Guarding against the presence of such sentinel strings in an actual config source is not someone anyone should have to deal with.

dmlloyd:
> > This is a nonsense argument.
> 
> @dmlloyd I found the above comment is hard to take. Is it ok to discuss the issue politely?

I'm sorry, it wasn't meant in a personal way.  I should have chosen a better way to communicate this idea.

My point is that the argument is not logically cohesive.  It is necessary for a specification change to be implementable, but it is self-evident that the implementability of a specification change is not sufficient to demonstrate its correctness or validity.  I find such arguments to be disingenuous and also detrimental to our task.

dmlloyd:
> Personally, I would like to avoid the whole issue of indicating nullability at the injection point by simply making that the default behavior. The current behavior assumes that the configuration a CDI bean relies on will be stable at CDI initialization time and thus it is appropriate to fail initialization by default when a property is missing. In reality, bean instances will be constructed depending on their scopes and the presence of a relied upon property cannot be guaranteed by a config check when the container is initialized.

This is true only when viewed through the lens of config mutability - a controversial topic indeed.

> When a property is missing, we should simply inject a null value unless a default is provided at the injection point.

This would cause some difficulties for those who wish to indicate that a given property is required to have a value, would it not?

> Further, if we already assume that an empty string is the same as a null value, we can remove the existing NOT_DEFINED sentinel string and just use an empty string as the default for defaultValue. Guarding against the presence of such sentinel strings in an actual config source is not someone anyone should have to deal with.

I agree with this point.  But I do think that it is necessary to be able to indicate whether an injected property is expected/required to have a non-empty value.

emattheis:
>This is true only when viewed through the lens of config mutability - a controversial topic indeed.

Given that we support system properties, which are inherently mutable, this seems like a moot point.

Regardless, it also assumes that the config used to retrieve the injected value is itself created and available at CDI instantiation time. This precludes the use of enlisting CDI beans themselves as config sources or converters as proposed in #125.

>This would cause some difficulties for those who wish to indicate that a given property is required to have a value, would it not?

>But I do think that it is necessary to be able to indicate whether an injected property is expected/required to have a non-empty value.

Yeah, that's fair. In my use cases I always have a default value or allow null, but for those who want a non-null guarantee I like @radcortez 's proposal to use a new field - e.g. `nullable = false`.

smillidge:
Given this is currently specified behaviour you wish to remove why don't you raise an issue first where you properly explain the issue before you implement a solution?

kenfinnigan:
Though it's currently specified in master, I don't believe the change has been part of a release to date, other than 1.4 RCs.

From a quick look, could this be a change that should've been removed when the `ConfigAccessor` pieces were removed, as it appears to have been added as part of the same large change?

emattheis:
How about we re-open #365?

smillidge:
It's a long time since a config release. I thought it must be released given the 2018 on the original PR.
I'm for reopening #365.

dmlloyd:
+1 on reopening #365.

dmlloyd:
I've reopened #365, and at least one feasible alternative solution has been proposed.  Can this be merged?

radcortez:
+1

GedMarc:
+1 Nullity in configuration sources is mandatory, they are very different from db, or business logic requirements

dmlloyd:
https://github.com/eclipse/microprofile-config/issues/365#issuecomment-579954337 - we must merge this because the original change was merged without consensus.  Let's get this merged, and then we can continue the discussion about whether we need two different ways to specify optionality for CDI config properties within #365.

radcortez:
+1

Emily-Jiang:
> we must merge this because the original change was merged without consensus. 

@dmlloyd I disagree with the above statement. The PR was merged ages ago and was approved by @OndroMih and Romain!

struberg:
> A weakness of the original approach can be shown quite simply: how, with the original approach, does the user specify that a property may be nullable, but by default it is not `null`? Answer: they cannot.

It is actually not a question of the config subsystem at all whether a value is optional or mandatory. This is solely up to the application. And it might even be mandatory for one app but not for the other.

Very early in the design discussion we also talked about metadata for config and dismissed it as too complex.

dmlloyd:
> > A weakness of the original approach can be shown quite simply: how, with the original approach, does the user specify that a property may be nullable, but by default it is not `null`? Answer: they cannot.
> 
> It is actually not a question of the config subsystem at all whether a value is optional or mandatory. This is solely up to the application. And it might even be mandatory for one app but not for the other.
> 
> Very early in the design discussion we also talked about metadata for config and dismissed it as too complex.

The existing implementation already provides two mechanisms for the application to specify whether a property should be required or optional.  The first mechanism is on `Config` itself by way of `getValue` and `getOptionalValue`.  The second mechanism is indirect by way of the CDI integration, where an application developer can type a property as `Optional`, thus requesting that the configuration system make the property optional.

In neither case is the user using or providing metadata about the configuration property; rather, they're providing metadata (if you want to call it that) about the actual usage of that property in a specific context.

I'm not arguing for the config system to start doing anything that it was not already doing before; quite the opposite in fact.

svenhaag:
Combining CDI and Optional doesn't make sense to me, as, at least in our case, in most cases, we are dealing with passivation capable beans (Serializable). I now went to route of `javax.enterprise.inject.Specializes` the ConfigProducer in combination with a custom annotation like `ConfigPropertyNullable`.

Issue #526, opened 2020-03-02
Config Lifecycle Events


radcortez:
As a user, I'm interested in being able to:

- Discover which Source was used to load my Config value
- Expand variables in my Config values
- Apply validations to my Config values

radcortez:
Related issues:
- 524
- 405
- 312
- 365
- 475

radcortez:
While all of these features could be implemented in one way or another, looking closing a lot of these operations match Config Lifecycle Events:

- Before Lookup in ConfigSource
- After Lookup in ConfigSource
- Before Value Conversion
- After Value Conversion

Maybe we can implement this API Lifecycle support, which will make the API more flexible to implement these and maybe other use cases.

dmlloyd:
Since this issue is really about a possible way to solve the use cases, I've tagged it with `Implementation`.

radcortez:
A very rough idea just to start the discussion :)

```java
public interface ConfigLifecycle {
    void beforeLookup(ConfigSource configSource, String propertyName);

    void afterLookup(ConfigSource configSource, String propertyName, String value);

    <T> void beforeConversion(Class<T> type, Converter<T> converter, String value);

    <T> void afterConversion(Class<T> type, T value);
}
```

dmlloyd:
For property expansion, you'd need some means to replace the value.  An interceptor-style approach works well in this case.

I'm not really sold that conversion and config source should be bundled up in this way (particularly if conversion gets hoisted off to its own API), given that they're different mechanisms.  The only reason to have this grouping is because there is a general correlation to the "shape" of the API, but it's not really clear that there would be any actual practical benefit to putting them together.

radcortez:
In that particular case, the Converter lifecycle events would be extracted to their own API (if that moves forward somehow).

Or, we could keep things more simple, and just have the beforeLookup and afterLookup. The problem with the Converters is that for property replacement, you do need to perform it before Conversion, no?

Probably we can have afterLookup to return a String and then you can do the property replacement over there.

dmlloyd:
Right, property replacement is really best done at the config source level.  In Quarkus we wrap the config source so it is essentially "interceptor style"; this works well.

Validation is best done at the converter level.  In SmallRye Config, we wrap converters and this is a very effective solution.

[...]

PR #527, opened 2020-03-03
Interceptor prototype to expose Config lifecycle like discussed in #526.

radcortez:
For discussion and review. Not the final implementation.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceInterceptor.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceInterceptor.java
@@ -1,0 +1,5 @@
+package org.eclipse.microprofile.config.spi;
+
+public interface ConfigSourceInterceptor {
+    String getValue(ConfigSourceInterceptorContext context) throws Exception;
+}
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceInterceptorContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceInterceptorContext.java
@@ -1,0 +1,9 @@
+package org.eclipse.microprofile.config.spi;
+
+public interface ConfigSourceInterceptorContext {
+    String getKey();
+
+    Class getConfigSource();
+
+    String proceed(String key) throws Exception;
+}
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
@@ -1,0 +1,29 @@
+package org.eclipse.microprofile.config.spi;
+
+public class ConfigValue {
+    private final String key;
+    private final String value;
+    private final Class configSource;
+
+    ConfigValue(final String key, final String value, final Class configSource) {
+        this.key = key;
+        this.value = value;
+        this.configSource = configSource;
+    }
+
+    public String getKey() {
+        return key;
+    }
+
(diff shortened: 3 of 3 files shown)


genie-microprofile:

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceInterceptor.java
```diff
@@ -0,0 +1,5 @@
+package org.eclipse.microprofile.config.spi;
+
+public interface ConfigSourceInterceptor {
+    String getValue(ConfigSourceInterceptorContext context) throws Exception;
```
dmlloyd: The return of this should probably be a `ConfigValue` I would think, and it should also accept a configuration key as a parameter (see my comment on `ConfigSourceInterceptorContext.getKey()`).  I'd drop `throws Exception` (see my comment on `ConfigSourceInterceptorContext.proceed()`.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceInterceptorContext.java
```diff
@@ -0,0 +1,9 @@
+package org.eclipse.microprofile.config.spi;
+
+public interface ConfigSourceInterceptorContext {
+    String getKey();
```
dmlloyd: We don't want to lock the key in here, otherwise we can't use an interceptor for property evaluation (which has to query potentially several property names).

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceInterceptorContext.java
```diff
@@ -0,0 +1,9 @@
...
+
+    Class getConfigSource();
```
dmlloyd: Apart from the raw type usage - we want something more here, something that allows us to give the user some information about the configuration property.  But it should probably only be on `ConfigValue` because the context should be independent of any single configuration source.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceInterceptorContext.java
```diff
@@ -0,0 +1,9 @@
...
+
+    String proceed(String key) throws Exception;
```
dmlloyd: Since `ConfigSourceInterceptor` would return a `ConfigValue`, so should this method.  Also I'd drop the `throws Exception` since we only want interceptors to throw the spec-defined exception types (`IllegalArgument` for validation problems, `NoSuchElement` for missing value cases).

kenfinnigan:
+1, what is proposed to MP should be something that's proven to be working

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceInterceptor.java
radcortez: Ops, I did create the class but forgot to use it in the return for this method.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceInterceptorContext.java
radcortez: Ops, I did create the class but forgot to use it in the return for this method.

radcortez:
Very well, lets put this on hold, and lets move to SmallRye Config and figure out something there. Thanks!

radcortez:
Actually just going to close this until we have a final solution implemented in SmallRye.

PR #528, opened 2020-03-04
Introduce issue templates

dmlloyd:
**Description**

This change introduces the usage of GitHub issue templates.  This helps to ensure that each opened issue has exactly one type, and that implementation changes relate back to use cases as discussed previously.

***Footnote***

Please indicate your support for this issue by choosing the :+1: reaction, rather than adding a "+1" or similar comment.

### DIFF
--- a/.github/ISSUE_TEMPLATE/30-impl-proposal.md
+++ b/.github/ISSUE_TEMPLATE/30-impl-proposal.md
@@ -1,0 +1,19 @@
+---
+name: Implementation proposal
+about: Request or propose an implementation that solves one or more use cases.
+title: ''
+labels: 'impl. proposal :gear:'
+assignees: ''
+
+---
+
+**Description**
+
+<!-- Describe the implementation proposal's functionality here -->
+
+
+**Use cases**
+
+<!-- Link at least one use case issue ID which would be solved by this implementation proposal -->
+
+* Fixes #xxxx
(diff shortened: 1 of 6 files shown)


radcortez:
I would like to see this discussed as a general process for all MP projects.

I know that each MP project has different people and each team might work differently, but for the general user, the look into MP as a platform, so they may get confused if they way they have to report stuff changes from project to project.

dmlloyd:
That's a great idea.  There could be a shared repository on `eclipse` which could contain the issue templates, that could be periodically merged into the downstream projects as desired.

Would you be willing to write a proposal to the main microprofile ML?

dmlloyd:
It looks as though the `.github` files have to be excluded from the license header check.  Does anyone know how to do that?

radcortez:
Sure. I'll propose something in the next couple of days.

radcortez:
> It looks as though the `.github` files have to be excluded from the license header check. Does anyone know how to do that?

Yes, try here:
https://github.com/eclipse/microprofile-config/blob/master/pom.xml#L291-L313

dmlloyd:
Perfect, thanks!

.github/ISSUE_TEMPLATE/30-impl-proposal.md
```diff
@@ -0,0 +1,19 @@
+---
+name: Implementation proposal
```
Emily-Jiang: I thought a bit on this. Since implementation proposal needs to solve at least one use case, why don't we just add implemenation ideas on that use case and link to the related use case?

PR #537, opened 2020-03-10
Fixes #529. Added Github Action Workflow for Build.

### LINKED ISSUE #529: Move to Github Actions
**Description**

We could move to Github Actions to handle Build and Release process for MicroProfile Config. Some of the main motivations to move are:

- Run the build in forks. You don't need to do the PR to check the status of the build
- Change the Build / Release configuration by just submitting a PR
- Committers have control over the CI
- Flexible to execute actions on Github Events

I've detailed some of the work done over SmallRye to use Github Actions here:
https://github.com/smallrye/smallrye-parent/wiki/SmallRye-Release-Process

The idea is not to pull the plug on Jenkins yet. We can run both in parallel. We can start with just the build part, which is easy to implement and go from there.

***Footnote***

Please indicate your support for this issue by choosing the :+1: reaction, rather than adding a "+1" or similar comment.

radcortez:


genie-microprofile:

.github/workflows/build.yml
```diff
@@ -0,0 +1,47 @@
...
+      - 'README*'
+      - 'site.yaml'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [8, 11]
+    name: build with jdk ${{matrix.java}}
```
Emily-Jiang: @radcortez at the moment, we are building all MP API Jars with Java 8. With the matrix including 11, which Java version will the Github action use by default? I am new to the action and this might be a naive question.

radcortez: In this particular case, there is no default, it will just run 2 separate builds, one with JDK 8 and another with 11. I think this might be useful, so we can be up to date to any things that might break on either version. It generate 2 separate status check, but we can setup to just pass the JDK 8 build and we keep the information about 11. Both are run on push or pull request events.

This does not handle the release of the project. For that, we need to add a separate workflow. In that case, we won't use a matrix strategy and just set the JDK to 8.

dmlloyd:
Can this be merged?

radcortez:
I believe so.

PR #538, opened 2020-03-11
Clarify rules and guarantees around property name sets

### LINKED ISSUE #431: What must the set of property names returned by Config#getPropertyNames() consist of?

dmlloyd:
Fixes #431

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -130,8 +130,25 @@ public interface Config {
     <T> Optional<T> getOptionalValue(String propertyName, Class<T> propertyType);
 
     /**
-     * Return all property names returned by any of the underlying {@linkplain ConfigSource configuration sources}.
-     * The order of the returned property names is unspecified.
+     * Returns a sequence of configuration property names. The order of the returned property names is unspecified.
+     * <p>
+     * The returned property names are unique; that is, if a name is returned once by a given iteration, it will not be
+     * returned again during that same iteration.
+     * <p>
+     * There is no guarantee about the completeness or currency of the names returned, nor is there any guarantee that a
+     * name that is returned by the iterator will resolve to a non-empty value or be found in any configuration source
+     * associated with the configuration; for example, it is allowed for this method to return an empty set always.
+     * However, the implementation <em>should</em> return a set of names that is useful to a
+     * user that wishes to browse the configuration.
+     * <p>
+     * It is implementation-defined whether the returned names reflect a point-in-time "snapshot" of names, or an
+     * aggregation of multiple point-in-time "snapshots", or a more dynamic view of the available property names.
+     * Implementations are not required to return the same sequence of names on each iteration; however, the produced
+     * {@link java.util.Iterator Iterator} must adhere to the contract of that class, and must not return any more
+     * elements once its {@link java.util.Iterator#hasNext() hasNext()} method returns {@code false}.
+     * <p>
+     * The returned instance is thread safe and may be iterated concurrently.  The individual iterators are not
+     * thread-safe.
      *
      * @return the names of all configured keys of the underlying configuration.
      */
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
@@ -54,7 +54,7 @@ import java.util.Set;
  * <ol>
  * <li>{@linkplain System#getProperties() System properties}, with an {@linkplain #getOrdinal() ordinal value} of {@code 400}</li>
  * <li>{@linkplain System#getenv() Environment properties}, with an ordinal value of {@code 300}</li>
- * <li>The {@code /META-INF/microprofile-config.properties} {@linkplain ClassLoader#getResource(String) resource}, 
+ * <li>The {@code /META-INF/microprofile-config.properties} {@linkplain ClassLoader#getResource(String) resource},
  * with an ordinal value of {@code 100}</li>
  * </ol>
  *
@@ -124,6 +124,13 @@ public interface ConfigSource {
      * Implementations should consider replacing this with a more performant implementation.
      * <p>
      * The returned property names may be a subset of the names of the total set of retrievable properties in this config source.
+     * <p>
+     * The returned set is not required to allow concurrent or multi-threaded iteration; however, if the same set is
+     * returned by multiple calls to this method, then the implementation must support concurrent and multi-threaded
+     * iteration of that set.
+     * <p>
+     * The set of keys returned <em>may</em> be a point-in-time snapshot, or <em>may</em> change over time (even
+     * during active iteration) to reflect dynamic changes to the available set of keys.
      *
      * @return a set of property names that are known to this configuration source
      */


dmlloyd:
Note that the language I use is based on what we do in SmallRye.  If your implementation needs more strict rules, please speak up!

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -132,6 +132,17 @@
     /**
      * Return all property names returned by any of the underlying {@linkplain ConfigSource configuration sources}.
      * The order of the returned property names is unspecified.
+     * <p>
+     * The returned property names are unique; that is, if a name is returned once, it will not be returned again.
+     * The names returned may reflect a subset of the total set of available names.
```
ljnelson: "names returned may reflect"?  Or "set of names must be"?

dmlloyd: The returned value is not a set, so we can't require it to be one.

ljnelson: (But the notional set of "all property names" is a `Set`, yes?  So mustn't the `Iterable` here have `Set` semantics?)

My larger point was the use of "may": by this specification, I could return an `Iterable` with `nonexisting.property.name1, nonexisting.property.name2` in it when none of my `ConfigSource`s have such property names and I'd be compliant.

I also just noticed that I don't understand what "all property names" at the top means, given line 137.  Can I return a subset of the `ConfigSource` property names, and only a subset of them?  Or must I return all of them?  May I return values in the `Iterable` that do not have corresponding values in the property names originating from the `ConfigSource` instances?  Must I not?  You see what I'm getting at.

dmlloyd: "Return all property names returned by any of the underlying..." is unambiguous: if the underlying config source returns the property name, it must be reported by the iterable.  That said, it might be a good idea to loosen this a bit.

The word "may" means that the implementation might show a reduced set of names compared to what might naively be considered available.  This follows on to the normative language in `ConfigSource` which says "The returned property names may be a subset of the names of the total set of retrievable properties in this config source".  But I can tighten up that language.

> My larger point was the use of "may": by this specification, I could return an `Iterable` with `nonexisting.property.name1, nonexisting.property.name2` in it when none of my `ConfigSource`s have such property names and I'd be compliant.

That is not true; the returned names would not be a subset of the total available names in this case.  That said though, there's probably no reason to make this quite so strict: really the config object could be allowed to return any old set of names.

Maybe this would be better language:

```java
/**
 * Returns a sequence of configuration property names.
 * The order of the returned property names is unspecified.
 * <p>
 * The returned property names are unique; that is, if a name is returned once by a given iteration, it will not be returned again during that same iteration.
 * <p>
 * There is no guarantee about the completeness or currency of the names returned, nor is there any guarantee that a name that is returned by the iterator will resolve to a non-empty value or be found in any configuration source associated with the configuration.
* <p>
 * It is implementation-defined whether the returned names reflect a point-in-time "snapshot" of names, or
 * an aggregation of multiple point-in-time "snapshots", or a more dynamic view of the available property names.
 * Implementations are not required to return the same sequence of names on each iteration; however, the produced
 * {@link java.util.Iterator Iterator} must adhere to the contract of that class, and must not return any more
 * elements once its {@link java.util.Iterator#hasNext() hasNext()} method returns {@code false}.
 * <p>
 * The returned instance is thread safe and may be iterated concurrently.  The individual iterators are not thread-safe.
 */
```

ljnelson: Sources `A`, `B` and `C` report `a`, `b`, and `c`.
The spec says: `[The returned Iterable must] return all property names returned by any of the underlying {@linkplain ConfigSource configuration sources}.`
That says I MUST return `a`, `b` and `c` if they are reported at the moment of this method's invocation.  OK.  They were (in this example).  Cool.  I'll (necessarily) add `a`, `b` and `c` to my `Iterable` as I'm building it.
Then the spec says: `The names returned may [or may not, my emphasis] reflect a subset of the total set of available names.`  Cool.  It doesn't say MUST.  I'll add `d` to my iterable for arcane implementation reasons.
With the new language (`There is no guarantee about the completeness or currency of the names returned, nor is there any guarantee that a name that is returned by the iterator will resolve to a non-empty value or be found in any configuration source associated with the configuration.`), `{ a, b, c, d }` is fine, as is `{ q, x, r }`.  At any point, it might also change during iteration to, say, `{ j, k, q }`.
Do I have this right so far?  Given how loose the method now can be, is it still worthwhile?

dmlloyd: > Sources `A`, `B` and `C` report `a`, `b`, and `c`.
> The spec says: `[The returned Iterable must] return all property names returned by any of the underlying {@linkplain ConfigSource configuration sources}.`
> That says I MUST return `a`, `b` and `c` if they are reported at the moment of this method's invocation. OK. They were (in this example). Cool. I'll (necessarily) add `a`, `b` and `c` to my `Iterable` as I'm building it.

I removed this language in the above proposal.

> Then the spec says: `The names returned may [or may not, my emphasis] reflect a subset of the total set of available names.` Cool. It doesn't say MUST. I'll add `d` to my iterable for arcane implementation reasons.

I removed this language as well.

> With the new language (`There is no guarantee about the completeness or currency of the names returned, nor is there any guarantee that a name that is returned by the iterator will resolve to a non-empty value or be found in any configuration source associated with the configuration.`), `{ a, b, c, d }` is fine, as is `{ q, x, r }`. At any point, it might also change during iteration to, say, `{ j, k, q }`.

Correct.  In Quarkus we inject names in the configuration so that configuration groups are read.  We remove names from the configuration if they correspond to sensitive configuration properties.  The proposed configuration interceptors (forthcoming) certainly allows all this and more.

> Do I have this right so far? Given how loose the method now can be, is it still worthwhile?

I think it is.  If you want, I could add a recommendation like "The implementation SHOULD return a set of names that is useful to a user that wishes to browse the configuration."  It would give a general intent without restricting application-specific requirements.

Sometimes it's OK for a specification to give latitude to the implementation.

ljnelson: Goodness, yes, it is _frequently_ OK for a specification to give latitude to the implementation.  The delicate dance is: what's specified?  What's not?  In this case, the spec is now becoming "return whatever you want, so long as someone can iterate over it concurrently".  I just want to make sure that's the intent of this method.  If it _is_ the intent of this method, then yes, the "should" statement—giving a sense of the use cases involved—would be a good thing IMHO, since otherwise a method that lets me just return `frob` and `quux` for a hard-to-discern purpose is probably surface area we don't want to support. 😄

dmlloyd: Pushed.

dmlloyd:
@ljnelson any further comments/feedback?

ljnelson:
This is much, much better than what we have now.  It might be nice to add a sentence that points out explicitly that it is legal for an implementation to return `Collections.emptySet()` or the like.  If you think about it from an implementor's standpoint coming to the one source of truth (the javadocs are, here, the specification, full stop) then it's useful to anticipate her questions.  It seems to me  that "can I skip this altogether initially?" might very well be one of them.

dmlloyd:
OK I've pushed up some additional language to that effect.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -130,8 +130,25 @@
     <T> Optional<T> getOptionalValue(String propertyName, Class<T> propertyType);
 
     /**
-     * Return all property names returned by any of the underlying {@linkplain ConfigSource configuration sources}.
-     * The order of the returned property names is unspecified.
+     * Returns a sequence of configuration property names. The order of the returned property names is unspecified.
```
ljnelson: Sorry, missed this earlier.  Up to you whether you want to deal with it.  A sequence (definitionally) has order.  Perhaps you mean just "collection".

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -130,8 +130,25 @@
...
+     * <p>
+     * There is no guarantee about the completeness or currency of the names returned, nor is there any guarantee that a
+     * name that is returned by the iterator will resolve to a non-empty value or be found in any configuration source
+     * associated with the configuration; for example, it is allowed for this method to return an empty set always.
+     * However, the implementation <em>should</em> return a set of names that is useful to a
+     * user that wishes to browse the configuration.
+     * <p>
+     * It is implementation-defined whether the returned names reflect a point-in-time "snapshot" of names, or an
+     * aggregation of multiple point-in-time "snapshots", or a more dynamic view of the available property names.
+     * Implementations are not required to return the same sequence of names on each iteration; however, the produced
```
ljnelson: Same as above

api/src/main/java/org/eclipse/microprofile/config/Config.java
dmlloyd: Right but iterations are by definition a sequence, and thus they do have *an* order.  We just don't specify what the order is.

PR #530, opened 2020-03-05
Fixes #513. Add TCK and Spec note about supporting OptionalInt, OptionalLong and OptionalDouble.

radcortez:

Issue #513, opened 2020-01-28
Complete Optional family support


rmannibucau:
Follow up ticket of a side note of https://github.com/eclipse/microprofile-config/issues/365

Wonder if there is any rational to not support OptionalInt and friends as valid injection types out of the box.

emattheis:
I would also suggest specifying support for `Optional<List<T>>` and `Optional<Set<T>>`.

Emily-Jiang:
@rmannibucau we do support Optional Injection. See the spec and [javadoc](https://download.eclipse.org/microprofile/microprofile-config-1.4-RC2/apidocs/) of `ConfigProperty` .

Emily-Jiang:
Do you see any other gaps in the Optional injection? If not, please close this issue.

rmannibucau:
This issue is about all optional flavors. Only Optional is supported, not OptionalInt and friends.

Emily-Jiang:
Since you cannot do `Optional<int>`, I think the boxed version `Optional<Integer>` makes sense.

rmannibucau:
@Inject @ConfigProperty OptionlInt optionalInt;

Same for doubles, booleans... ;)

Emily-Jiang:
ah. sorry. I now got it. Thanks! +1 on support it!

radcortez:
These are the specialised Optional types:

`java.lang.OptionalDouble`
`java.lang.OptionalInt`
`java.lang.OptionalLong`

Emily-Jiang:
Let's get it done in June's release.

Emily-Jiang:
@radcortez

radcortez:
Done.

PR #540, opened 2020-03-12
Add language to converters spec about additional optional types

dmlloyd:
Follows up #513, follows up #530


dmlloyd:
@radcortez please have a look.

PR #541, opened 2020-03-12
Establish consistent behavior for empty value handling

### LINKED ISSUE #397: Need a way to specify empty list/array in microprofile-config.properties

### LINKED ISSUE #446: Consistent empty/missing behavior

### LINKED ISSUE #531: Enable explicit clearing of values
**Description**

As a:

- [X] Application user/user of the configuration itself
- [ ] API user (application developer)
- [ ] SPI user (container or runtime developer)

...I need to be able to:

have a way to specify that a property is explicitly cleared.

...which enables me to:

* clear properties on the command line or system environment for testing and debugging purposes e.g. `-Dsome.property=`
* reuse configurations by overlay, even if the underlying configuration specifies a configuration value I do not want to have set

### LINKED ISSUE #532: Consistent empty/missing value handling (use case)
**Description**

As a:

- [X] API user (application developer)
- [ ] SPI user (container or runtime developer)

...I need to be able to:

* Have a consistent experience handling values that are empty or missing that does not need to vary depending on the conversion type
* Detect that a property is empty regardless of its type
* Decode an empty value into the target type in a way that is appropriate for the use site
* Not be concerned with whether a collection with a given element type can or cannot contain an empty value

...which enables me to:

Simplify my experience and apply patterns for empty values to any value type, and enable property clearing (#531).

dmlloyd:
Fixes #446, Fixes #531, Fixes #532, Fixes #397.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -34,6 +34,7 @@
 
 package org.eclipse.microprofile.config;
 
+import java.util.NoSuchElementException;
 import java.util.Optional;
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
@@ -94,6 +95,10 @@ public interface Config {
      * The configuration value is not guaranteed to be cached by the implementation, and may be expensive
      * to compute; therefore, if the returned value is intended to be frequently used, callers should consider storing
      * rather than recomputing it.
+     * <p>
+     * The result of this method is identical to the result of calling {@code getOptionalValue(propertyName, propertyType).get()}.
+     * In particular, if the value element of the given property name does not exist, a {@link NoSuchElementException} is thrown.
+     * This method never returns {@code null}.
      *
      * @param <T>
      *             The property type
@@ -101,11 +106,11 @@ public interface Config {
      *             The configuration property name
      * @param propertyType
      *             The type into which the resolved property value should get converted
-     * @return the resolved property value as an instance of the requested type
-     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
-     * @throws java.util.NoSuchElementException if the property isn't present in the configuration
+     * @return the resolved property value as an instance of the requested type (not {@code null})
+     * @throws IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws NoSuchElementException if the property does not exist
      */
-    <T> T getValue(String propertyName, Class<T> propertyType);
+    <T> T getValue(String propertyName, Class<T> propertyType) throws IllegalArgumentException, NoSuchElementException;
 
     /**
      * Return the resolved property value with the specified type for the
@@ -125,9 +130,9 @@ public interface Config {
      *             The type into which the resolved property value should be converted
      * @return The resolved property value as an {@code Optional} wrapping the requested type
      *
-     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws IllegalArgumentException if the property cannot be converted to the specified type
      */
-    <T> Optional<T> getOptionalValue(String propertyName, Class<T> propertyType);
+    <T> Optional<T> getOptionalValue(String propertyName, Class<T> propertyType) throws IllegalArgumentException;
 
     /**
      * Returns a sequence of configuration property names. The order of the returned property names is unspecified.
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
@@ -102,12 +102,28 @@ import java.io.Serializable;
  * <em>Implicit</em> converters are only created when no other converter was found; therefore, they do not have a
  * priority.
  *
+ * <h3 id="empty">Empty values</h3>
+ *
+ * For all converters, the empty string {@code ""} <em>must</em> be considered an empty value. Some converters <em>may</em>
+ * consider other values to be empty as well.
+ * <p>
+ * When a conversion results in an empty value, most converters should normally return {@code null}.  All of the built-in
+ * converters <em>must</em> return {@code null} for an empty value, <em>except</em> for the built-in converters for
+ * {@code OptionalInt}, {@code OptionalLong}, and {@code OptionalDouble}, which <em>must</em> return their corresponding
+ * {@code empty()} value for an empty input.
+ * <p>
+ * Implementations <em>may</em> (but are not required to) implement {@code Config.getOptionalValue()} using a
+ * {@code Converter}.  If so, this converter <em>must</em> return {@code Optional.empty()} for an empty input.
+ *
  * <h3>Array conversion</h3>
  *
  * A conforming implementation must support the automatic creation of an <em>implicit</em> converter for array types.
  * This converter uses a comma ({@code U+002C ','}) as a delimiter.  To allow a comma to be embedded within individual
  * array element values, it may be escaped using a backslash ({@code U+005C '\'}) character.  Any escaped comma character
  * will be included as a plain comma within the single element (the backslash is discarded by the converter).
+ * <p>
+ * Empty elements <em>must</em> not be included in the final array.  An array which would consist of only empty values
+ * <em>must</em> be considered empty; the array converter <em>must</em> return {@code null} in this case.
  *
  * @author <a href="mailto:rsmeral@apache.org">Ron Smeral</a>
  * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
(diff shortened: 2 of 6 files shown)


dmlloyd:
This first draft only addresses documentation.  TCK testing will come in a follow-up commit in this PR before it is taken out of "draft" state.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -94,18 +95,22 @@
      * The configuration value is not guaranteed to be cached by the implementation, and may be expensive
      * to compute; therefore, if the returned value is intended to be frequently used, callers should consider storing
      * rather than recomputing it.
+     * <p>
+     * The result of this method is identical to the result of calling {@code getOptionalValue(propertyName, propertyType).get()}.
+     * In particular, if the value element of the given property name does not exist, a {@link NoSuchElementException} is thrown.
```
Emily-Jiang: How about:
If the given property name or the value element of the give property name does not exist, a {@link NoSuchElementException} is thrown.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -94,18 +95,22 @@
...
      * @param propertyName
      *             The configuration property name
      * @param propertyType
      *             The type into which the resolved property value should get converted
-     * @return the resolved property value as an instance of the requested type
-     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
-     * @throws java.util.NoSuchElementException if the property isn't present in the configuration
+     * @return the resolved property value as an instance of the requested type (not {@code null})
+     * @throws IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws NoSuchElementException if the property does not exist
```
Emily-Jiang: NoSuchElementException if the property name or the value element of the given property name does not exist.

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -33,7 +33,7 @@ A higher `ordinal` means that the values taken from this `ConfigSource` will ove
 This allows a configuration to be customized from outside a binary, assuming that external `ConfigSource` s have higher `ordinal` values than the ones whose values originate within the release binaries.
 
 It can also be used to implement a drop-in configuration approach.
-Simply create a jar containing a `ConfigSource` with a higher ordinal and override configuration values in it.
+Simply create a jar containing a `ConfigSource` with a higher ordinal and override or clear configuration values in it.
```
radcortez: Should we also add here that the empty string is considered an empty value for effects of clearing a configuration?

jbee:
Let me see if I can summarise the new semantics: If a value is given or "overridden" with the empty string the outcome of a lookup is the same as the value not being present. Does that sound like the intention here?

api/src/main/java/org/eclipse/microprofile/config/Config.java
struberg, line 111: what does 'does not exist' mean? of course if the key is not contained in the config. But what happens with an empty string?

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -102,12 +102,28 @@
  * <em>Implicit</em> converters are only created when no other converter was found; therefore, they do not have a
  * priority.
  *
+ * <h3 id="empty">Empty values</h3>
+ *
+ * For all converters, the empty string {@code ""} <em>must</em> be considered an empty value. Some converters <em>may</em>
```
struberg, line 107: I have not fully reviewed whether this changes any logic. But please be aware, that changing this logic will also have an effect on default values within mp-config but also business apps. It might be a hidden API change.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -30,7 +30,7 @@ All this must be possible without the need to re-package the whole application b
 
 MicroProfile Config provides a way to achieve this goal by aggregating configuration from many different <<configsource,ConfigSources>> and presents a single merged view to the user.
 This allows the application to bundle default configuration within the application.
-It also allows to override the defaults from outside, e.g. via an environment variable a Java system property or via a container like Docker.
+It also allows to override or clear the defaults from outside, e.g. via an environment variable a Java system property or via a container like Docker.
```
struberg, line 33: this change is redundant. To 'override' already includes clearing.

api/src/main/java/org/eclipse/microprofile/config/Config.java
svenhaag: It has the same semantical meaning, meanwhile.

spec/src/main/asciidoc/architecture.asciidoc
dmlloyd: This is deliberate, in order to make the specification clearer.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
dmlloyd: This does change logic.  But it's not a hidden change; on the contrary it's spelled out clearly for the first time.

api/src/main/java/org/eclipse/microprofile/config/Config.java
dmlloyd: See Converter.java#107.

dmlloyd:
> Let me see if I can summarise the new semantics: If a value is given or "overridden" with the empty string the outcome of a lookup is the same as the value not being present. Does that sound like the intention here?

That is exactly the intent.  In addition to enabling this ability, this change also fixes several inconsistencies and contradictions.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -102,12 +102,28 @@
...
+ * consider other values to be empty as well.
+ * <p>
+ * When a conversion results in an empty value, most converters should normally return {@code null}.  All of the built-in
+ * converters <em>must</em> return {@code null} for an empty value, <em>except</em> for the built-in converters for
```
Emily-Jiang, line 111: what about the primitive types, e.g. int?

dmlloyd: There are no primitive type converters, only boxed type converters.  And boxed type converters must also return `null` for empty values.  If a CDI-style property has a "primitive" type, it will unbox for you, but it cannot be empty by nature (because primitive values have no empty representation).

Don't confuse conversion result with the result of the property retrieval operation.  If a conversion returns `null` and `getValue` or an equivalent (e.g. CDI injection of a non-optional type) is used, then `NoSuchElementException` would be thrown (as these operations can never yield a `null` value), because the value is empty and the type does not allow empty values.

Types such as `OptionalInt` allow an emptiable `int` value to be represented.

Emily-Jiang: I understand from the implementation point of view. I directly comment on the above text. In the converter spec, we list `int` under Built-in converters. I think people will be confused when they saw the above statement. 

> 

The following Converters are provided by MicroProfile Config by default:

    boolean and java.lang.Boolean , values for true (case insensitive) "true", "1", "YES", "Y" "ON". Any other value will be interpreted as false

    byte and java.lang.Byte

    short and java.lang.Short

    int, java.lang.Integer, and java.util.OptionalInt

    long, java.lang.Long, and java.util.OptionalLong

    float and java.lang.Float; a dot '.' is used to separate the fractional digits

    double, java.lang.Double, and java.util.OptionalDouble; a dot '.' is used to separate the fractional digits

    char and java.lang.Character

    java.lang.Class based on the result of Class.forName



Maybe in the built-in converters section, we can add the primitive convertion is done via the boxed type.

radcortez:
I think we should move this for the next version (2.0). We don't have much time left.

radcortez:
Moved to #589

PR #542, opened 2020-03-12
Clarify that nulls cannot be passed in to Converters

### LINKED ISSUE #539: clarify behavior when null is passed to Converter::convert
Per [a discussion](https://github.com/eclipse/microprofile-config/pull/495#discussion_r391108071) in #495, `null` should never be passed to a converter. We should clarify what exception should be thrown if this occurs.

Obvious choices are `NullPointerException` or `IllegalArgumentException`.

dmlloyd:
Fixes #539

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
@@ -121,12 +121,14 @@ import java.io.Serializable;
  */
 public interface Converter<T> extends Serializable {
     /**
-     * Convert the given string value to a specified type.
+     * Convert the given string value to a specified type. Callers <em>must not</em> pass in {@code null}
+     * for {@code value}; doing so may result in a {@code NullPointerException} being thrown.
      *
-     * @param value the string representation of a property value
+     * @param value the string representation of a property value (must not be {@code null})
      * @return the converted value, or {@code null} if the value is empty
      *
      * @throws IllegalArgumentException if the value cannot be converted to the specified type
+     * @throws NullPointerException if the given value was {@code null}
      */
-    T convert(String value);
+    T convert(String value) throws IllegalArgumentException, NullPointerException;
 }
--- a/spec/src/main/asciidoc/converters.asciidoc
+++ b/spec/src/main/asciidoc/converters.asciidoc
@@ -44,6 +44,7 @@ The following ``Converter``s are provided by MicroProfile Config by default:
 
 All built-in ``Converter``s have the `@Priority` of `1`.
 
+The converters for these types must throw an NPE if given a null value to convert.
 
 === Adding custom Converters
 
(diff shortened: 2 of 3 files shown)


dmlloyd:
There are two possible things that we can test:

* That the implementation never calls `convert` with a `null` value - but this cannot be proven with a test;
* That all converters throw exactly NPE with `null` - but we can't possibly test all converters, and the language specifically marks this as undefined (but allowed) behavior.

So what should be tested in the TCK?  If you agree with the documentation changes, then you must also conclude that adding a TCK test for this behavior is impossible.

Emily-Jiang:
If there is no tck, how can we ensure implemenations did this? I suggest passing a null, the test expects to get NPE.

dmlloyd:
Please carefully read the specification.  The point is to make the implementation as simple as possible for implementors, by guaranteeing that a conforming specification implementation never passes `null`.  However this is impossible to prove because we can't analyze every point where an implementation might call this method.

We're not trying to prove that implementations *do* something, we're trying to restrict them so they *do not* do something.  But you cannot prove such a thing by a test; it is impossible.

At the same time, we should not *require* all implementations to throw an NPE.  And the TCK does not test all possible implementations anyway - at most it can only test the implementations provided by the MP Config implementation being tested, but *requiring* an NPE here does nothing to help the user or the implementor.  In particular, *requiring* an NPE might make some lambda implementations of `Converter` impossible!

See also https://tools.ietf.org/html/rfc2119 for the most common way that specification language logically implies requirements.

Emily-Jiang:
Since we have exposed the converter via config, when an enduser calls config.convert(null), with this clarification, I expect a NPE.

dmlloyd:
I believe we did not end up exposing `convert` via `Config`.

radcortez:
Indirectly only. We expose the `Converter`, but not the `converter` method. Calling `converter` its on the user on the retrieved `Converter`.

Emily-Jiang:
Am I right to say the only remaining task is to add a TCK for this? @dmlloyd @radcortez

radcortez:
I believe so.

dmlloyd:
See my comment above. No TCK is possible for this particular change.

dmlloyd:
It doesn't matter; `Converter` can be implemented by the user so we can't TCK-test it.  It would only be possible to TCK-test if we could somehow get every registered converter from the implementation.  Such an operation would have no use outside of the TCK.  This can't be done!

radcortez:
Would it make sense to test the built in converters?

dmlloyd:
**As a separate test/spec requirement** we *could* say "The converters for these types: (list them) must throw an NPE if given a `null` value".  This is *not the same* as saying "The implementation will never pass in a `null` value to converters" and would represent an **additional** requirement (and should be a separate issue IMO because the value of adding this requirement is highly debatable).

radcortez:
Just added a TCK for the Built In Converters and added the requirement in the spec. If this is not desired, we can just drop the last commit.

Emily-Jiang:
@radcortez can you add this to the release notes? Sorry, I forgot to mention this earlier.

PR #544, opened 2020-03-13
Update version to allow API changes

dmlloyd:
Signed-off-by: Erik Mattheis <erikmattheis@me.com>


dmlloyd:
I'm merging this one so we can more forward with other changes.  Ping me ASAP if this is a problem.

PR #547, opened 2020-03-16
Update build.yml to make git use merge commit

dmlloyd:


dmlloyd:
@radcortez does this make sense to you?

dmlloyd:
The "push" ones are extras because I created the PR using the GUI thus creating a branch on upstream.  It would be nice if GitHub let you create the PR with a branch on your own fork, but it doesn't.

radcortez:
You shouldn't need to do that.

The checkout action uses the $GITHUB_SHA to checkout the code. On a pull_request event this is setup to be:

- GITHUB_SHA Last merge commit on the GITHUB_REF branch
- GITHUB_REF PR merge branch refs/pull/:prNumber/merge

dmlloyd:
That's what I thought, but when I looked in the logs, it did not appear to be using the merge SHA - it was checking out the actual branch instead.  So....?????

dmlloyd:
I should have captured the logs and put it in the PR.

radcortez:
Weird. I've just took that out of their documentation:
https://help.github.com/en/actions/reference/events-that-trigger-workflows

Until now, it worked fine in SmallRye Config. So, not sure what is the issue.

dmlloyd:
Yeah it's weird.  I checked another older PR and it used the merge commit, but more recent ones did not.  I'll see if I can find another example.

radcortez:
Ah, maybe because the build is only triggered when you push something to the PR. So build was done before your change and you were looking to some old build?

The build doesn't auto run for active PR's if you push stuff to master.

dmlloyd:
It wasn't that it was an outdated commit - it was actually checking out the branch, and naming the checked out branch for the PR's source branch, completely disregarding the merge commit.  I wonder if something on the GitHub side was wrong somehow?

carlosdlr:
Sorry guys this is to allow for every commit in a PR run the CI or just when this is merge it to master?

radcortez:
Well, Github as been having some weird issues sometimes.

dmlloyd:
@carlosdlr the idea is to run CI against a merge commit of the PR and master (instead of just running CI on the PR branch itself).  The weird thing is that (at least sometimes) CI already does this... but other times it does not?  I guess we'll have to keep an eye on it.  At some point we can try removing this to see what happens.

carlosdlr:
@dmlloyd got it yes better to keep an eye maybe is coronavirus "bad taste joke" or maybe the are running some maintenance window, but didn't receive some email about it

Issue #43, opened 2017-02-02
Access to config value metadata using a ConfigValue wrapper


starksm64:
As far as I can see from all of the current proposals, the Config.get(...) is returning a resolved value that has thrown away any notion of the source and conversion information. Wouldn't it be better for the base return value from Config.get(...) to be a ConfigValue that maintains the information along the lines of:

public interface ConfigValue {

    /**
     * Return the resolved property value with the specified type for the
     * specified property name
     * 
     * @param <T>
     *             the property type
     * @param propertyType
     *             The type into which the resolve property value should get converted
     * @return the resolved property value as an Optional of the requested type.
     */
    <T> Optional<T> getValue(Class<T> propertyType);

    /**
     * Get configuration value propertyName.
     *
     *             
     * @return The configuration propertyName.
     *
     */
    String getName();

    /**
     * @return all currently registered {@link ConfigSource}s for this value
     */
    Iterable<ConfigSource> getConfigSources();
...

Emily-Jiang:
Can you describe your user case? You can obtain a list of config sources from config in an ordered fashion. If you want to find out the serving config source for a particular config, if you search search the config source and the first config source found will be the config owner. I don't see how frequent you need to retrieve its source. Therefore, I don't think it is necessary to return a wrapper again.

starksm64:
The use case is that in a management interface that is allowing runtime editing/viewing of the existing configuration, one can see an annotation for the source of the current value. Although this does not have to happen from the value level itself, it is simpler, and it lends itself as the point for  extensions we have discussed down the road such as change listeners, etc.

struberg:
Hi Scott!
ConfigSources are not intended to be modified by a user. Each ConfigSource can of course provide it's very own way to change configured values. 
And for just looking up which ConfigSource contains a value you can also easily iterate over the ConfigSources and ask each of them on it's own.
Any change-listener approach would need a much deeper integrated functionality imo.

The reason for the ConfigValue is to allow functionality which needs to keep some state. E.g. the last load time, previously loaded values, etc. 

There are 4 features I like to get covered in our Config system. We do not need to implement all of them now, but they are all useful:

- variable replacement, e.g. myprj.url=http://${myprj.archive.server}/docs
- dynamic reloading / caching: val.cacheFor(5, TimeUnit.MINUTES);
[...]

Issue #312, opened 2018-01-26
Enable users to determine 'winning' source for a value - where is it coming from?


hutchig:
When working with the issue described in #310. Resolving the cause of the problem took longer as it was not initially obvious that a environment variable was being overridden by a system property of the same name. 

Should a 'source' of a value be identified by something in the API or by implementation specific debug/trace? (Or the latter making use of the former.)

Emily-Jiang:
It is useful to find out which source serves the config. I think it can be down but not tie with a dedicated method. config.wrap() as discussed #84 might be a good fit for this. Any other useful details can be logged there. We can provide a noop impl for this. The implementor can choose what do you want to provide in this method. One good suggestion is to provide the info noted in this issue.

Emily-Jiang:
@gunnarmorling what do you think? Is this a good use case?

jmesnil:
We had a similar discussion in the ConfigJSR https://github.com/eclipse/ConfigJSR/issues/10.

I think that the mp-config implementation should have tools to help resolve/analyze the configuration that is provided to the application.
However such tools can be built on top of the current API by separate tool or by the application itself

Emily-Jiang:
#43 should be able to figure out which configsource a particular config is coming from. By the way, I don't think the immutability has something to do with this one as the config_ordinal is static and newly added configure source will not be discovered after app starts.

dmlloyd:
> #43 should be able to figure out which configsource a particular config is coming from. By the way, I don't think the immutability has something to do with this one as the config_ordinal is static and newly added configure source will not be discovered after app starts.

It does.  What I said above applies to the implementation option of a method which provides the `ConfigSource` of a given configuration key independently of accessing its value.  If a config source can change, then a configuration key can disappear in the time between when a value is read and when the originating configuration source of that value is queried, and this has nothing to do with config ordinals.

radcortez:
This is now possible with #551.

PR #551, opened 2020-04-03
ConfigValue API

radcortez:
Here is the initial proposal to support the ConfigValue API in MP Config.

Most likely, we need more detail in the spec, but I didn't want to be really extensive, since all of this is subject to change during revision.

Fixes issues:
https://github.com/eclipse/microprofile-config/issues/43
https://github.com/eclipse/microprofile-config/issues/312

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -90,7 +90,6 @@ import org.eclipse.microprofile.config.spi.Converter;
  */
 @org.osgi.annotation.versioning.ProviderType
 public interface Config {
-
     /**
      * Return the resolved property value with the specified type for the
      * specified property name from the underlying {@linkplain ConfigSource configuration sources}.
@@ -110,6 +109,27 @@ public interface Config {
      * @throws java.util.NoSuchElementException if the property isn't present in the configuration
      */
     <T> T getValue(String propertyName, Class<T> propertyType);
+
+    /**
+     * Return the {@link ConfigValue} for the specified property name from the underlying
+     * {@linkplain ConfigSource configuration source}. The lookup of the configuration is performed immediatily,
+     * meaning that calls to {@link ConfigValue} will always yeld the same results.
+     * <p>
+     *
+     * The configuration value is not guaranteed to be cached by the implementation, and may be expensive
+     * to compute; therefore, if the returned value is intended to be frequently used, callers should consider storing
+     * rather than recomputing it.
+     * <p>
+     *
+     * A {@link ConfigValue} is always returned even if a property name cannot be found. In this case, every method in
+     * {@link ConfigValue} returns {@code null} except for {@link ConfigValue#getName()}, which includes the original
+     * property name being looked up.
+     *
+     * @param propertyName
+     *              The configuration property name
+     * @return the resolved property value as a {@link ConfigValue}
+     */
+    ConfigValue getConfigValue(String propertyName);
 
     /**
      * Return the resolved property values with the specified type for the
--- a/api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package org.eclipse.microprofile.config;
+
+/**
+ * The ConfigValue holds additional information after the lookup of a configuration property and is itself immutable.
+ * <p>
+ *
+ * Holds information about the configuration property name, configuration value, the
+ * {@link org.eclipse.microprofile.config.spi.ConfigSource} name from where the configuration property was loaded and
+ * the ordinal of the {@link org.eclipse.microprofile.config.spi.ConfigSource}.
+ * <p>
+ *
+ * This is used together with {@link Config} to expose the configuration property lookup metadata.
+ *
+ * @author <a href="mailto:radcortez@yahoo.com">Roberto Cortez</a>
+ */
+public interface ConfigValue {
+    /**
+     * The name of the property.
+     *
+     * @return the name of the property.
+     */
+    String getName();
+
+    /**
+     * The value of the property lookup.
+     *
+     * @return the raw value of the property lookup or {@code null} if the property could not be found.
+     */
+    String getValue();
+
+    /**
+     * The {@link org.eclipse.microprofile.config.spi.ConfigSource} name that loaded the property lookup.
+     *
+     * @return the ConfigSource name that loaded the property lookup or {@code null} if the property could not be found.
+     */
+    String getSourceName();
+
+    /**
+     * The {@link org.eclipse.microprofile.config.spi.ConfigSource} ordinal that loaded the property lookup.
+     *
+     * @return the ConfigSource ordinal that loaded the property lookup or {@code 0} if the property could not be
+     * found.
+     */
+    int getSourceOrdinal();
+}
--- a/spec/src/main/asciidoc/configexamples.asciidoc
+++ b/spec/src/main/asciidoc/configexamples.asciidoc
@@ -38,8 +38,11 @@ public class ConfigUsageSample {
         Config config = ConfigProvider.getConfig();
 
         String serverUrl = config.getValue("acme.myprj.some.url", String.class);
-
         callToServer(serverUrl);
+
+        // or
+        ConfigValue configServerUrl = config.getConfigValue("acme.myprj.some.url");
+        callToServer(configServerUrl.getValue());
     }
 }
 ----
@@ -80,6 +83,13 @@ public class InjectedConfigUsageSample {
     @Inject
     @ConfigProperty(name="myprj.some.url")
     private String someUrl;
+
+    // You can also inject a configuration using the ConfigValue metadata object. The
+    // configured value will not lead to a DeploymentException if the value is missing.
+    // A default value can also be specified like any other configuration.
+    @Inject
+    @ConfigProperty(name="myprj.another.url")
+    private ConfigValue anotherUrl;
 
     //The following code injects an Optional value of myprj.some.port property.
     //Contrary to natively injecting the configured value, this will not lead to a
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigValueTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigValueTest.java
@@ -1,0 +1,175 @@
+/*
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.config.tck;
+
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.eclipse.microprofile.config.ConfigValue;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.testng.Arquillian;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.testng.annotations.Test;
+
+import javax.enterprise.context.Dependent;
+import javax.inject.Inject;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertNull;
+
+public class ConfigValueTest extends Arquillian {
+    @Deployment
+    public static Archive deployment() {
+        JavaArchive testJar = ShrinkWrap
+            .create(JavaArchive.class, "ConfigValueTest.jar")
+            .addClasses(ConfigValueBean.class)
+            .addAsServiceProvider(ConfigSource.class, ConfigValueConfigSource.class)
+            .addAsServiceProvider(ConfigSource.class, ConfigValueLowerConfigSource.class)
+            .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+            .as(JavaArchive.class);
+
+        return ShrinkWrap
+            .create(WebArchive.class, "ConfigValueTest.war")
+            .addAsLibrary(testJar);
+    }
+
+    @Test
+    void configValue() {
+        ConfigValue configValue = ConfigProvider.getConfig().getConfigValue("my.prop");
+        assertNotNull(configValue);
+        assertEquals("my.prop", configValue.getName());
+        assertEquals("1234", configValue.getValue());
+        assertEquals("ConfigValueConfigSource", configValue.getSourceName());
+        assertEquals(1000, configValue.getSourceOrdinal());
+    }
+
+    @Test
+    public void configValueEmpty() {
+        ConfigValue configValue = ConfigProvider.getConfig().getConfigValue("not.found");
+        assertNotNull(configValue);
+        assertEquals("not.found", configValue.getName());
+        assertNull(configValue.getValue());
+        assertNull(configValue.getSourceName());
+        assertEquals(0, configValue.getSourceOrdinal());
+    }
+
+    @Inject
+    private ConfigValueBean configValueBean;
+
+    @Test
+    public void configValueInjection() {
+        final ConfigValue configValue = configValueBean.getConfigValue();
+        assertNotNull(configValue);
+        assertEquals("my.prop", configValue.getName());
+        assertEquals("1234", configValue.getValue());
+        assertEquals(ConfigValueConfigSource.class.getSimpleName(), configValue.getSourceName());
+        assertEquals(1000, configValue.getSourceOrdinal());
+
+        final ConfigValue configValueDefault = configValueBean.getConfigValueDefault();
+        assertNotNull(configValueDefault);
+        assertEquals("my.prop", configValue.getName());
+        assertEquals("default", configValueDefault.getValue());
+        assertNull(configValueDefault.getSourceName());
+    }
+
+    public static class ConfigValueConfigSource implements ConfigSource {
+        private Map<String, String> properties;
+
+        public ConfigValueConfigSource() {
+            properties = new HashMap<>();
+            properties.put("my.prop", "1234");
+        }
+
+        @Override
+        public Set<String> getPropertyNames() {
+            return properties.keySet();
+        }
+
+        @Override
+        public String getValue(final String propertyName) {
+            return properties.get(propertyName);
+        }
+
+        @Override
+        public String getName() {
+            return this.getClass().getSimpleName();
+        }
+
+        @Override
+        public int getOrdinal() {
+            return 1000;
+        }
+    }
+
+    public static class ConfigValueLowerConfigSource implements ConfigSource {
+        private Map<String, String> properties;
+
+        public ConfigValueLowerConfigSource() {
+            properties = new HashMap<>();
+            properties.put("my.prop", "5678");
+        }
+
+        @Override
+        public Set<String> getPropertyNames() {
+            return properties.keySet();
+        }
+
+        @Override
+        public String getValue(final String propertyName) {
+            return properties.get(propertyName);
+        }
+
+        @Override
+        public String getName() {
+            return this.getClass().getSimpleName();
+        }
+
+        @Override
+        public int getOrdinal() {
+            return 900;
+        }
+    }
+
+    @Dependent
+    public static class ConfigValueBean {
+        @Inject
+        @ConfigProperty(name = "my.prop")
+        private ConfigValue configValue;
+
+        @Inject
+        @ConfigProperty(name = "my.prop.default", defaultValue = "default")
+        private ConfigValue configValueDefault;
+
+        ConfigValue getConfigValue() {
+            return configValue;
+        }
+
+        ConfigValue getConfigValueDefault() {
+            return configValueDefault;
+        }
+    }
+}


genie-microprofile:

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValueConfigSource.java
```diff
@@ -0,0 +1,203 @@
...
+ * <p>
+ *
+ * This is to work around the limitation from the original {@link ConfigSource}. It exposes everything as plain Strings
+ * and it is not possible to retrieve additional information associated with the Configuration. The
+ * ConfigValueConfigSource tries to make this possible.
+ * <p>
+ *
+ * @author <a href="mailto:radcortez@yahoo.com">Roberto Cortez</a>
+ */
+public interface ConfigValueConfigSource extends ConfigSource {
```
radcortez, line 43: I'm not sure if we should leave this separate or add them as default methods in `ConfigSource`.

dmlloyd: I was thinking we should separate it completely, introducing a `ConfigValueSource` and `ConfigValueSourceProvider` API (separating also allows us to shorten the name a little bit).  We could then deprecate `ConfigSource`; for interim compatibility, we could specify that the implementation must automatically wrap old style `ConfigSource` with `ConfigValueSource`.  Then we can eventually drop those off after an acceptable period of time, and not have to deal with lots of sticky overloads.

WDYT?

radcortez:
You mean the `ConfigSourceInterceptor`? What is your proposal then?

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
```diff
@@ -0,0 +1,162 @@
...
+        return configSourceName;
+    }
+
+    public int getConfigSourceOrdinal() {
+        return configSourceOrdinal;
+    }
+
+    public int getLineNumber() {
+        return lineNumber;
+    }
```
emattheis, line 72: I think we should consider an abstraction for source info. Not all sources will be line-oriented.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
```diff
@@ -0,0 +1,162 @@
...
+    private final int configSourceOrdinal;
+
+    private final int lineNumber;
+
+    private ConfigValue(final ConfigValueBuilder builder) {
+        this.name = builder.name;
+        this.value = builder.value;
+        this.configSourceName = builder.configSourceName;
+        this.configSourceOrdinal = builder.configSourceOrdinal;
+        this.lineNumber = builder.lineNumber;
```
Emily-Jiang, line 51: I would remove this as the config could be in a database and the linenumber is not applicable. If it is in a file, string search is good enough.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValueConfigSource.java
Emily-Jiang: For simplicity, I prefer to add them as default methods in ConfigSource. I don't see any conflicts. Besides, I am not a fan to depricate for naming reasons. ConfigSource is very easy to be understood.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
dmlloyd: I would not remove this.  If the backend has a line number, this can return it.  If the backend does not, it can return a sentinel (such as -1) indicating that is the case.  Having the line number allows useful diagnostics to be printed, without requiring the larger overhead of having special location classes.

This is similar to what is done in libraries such as JAXP.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
dmlloyd: The problem with this approach is that every abstraction requires another instance to allocate, and that can add up quickly.  See also #552.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValueConfigSource.java
dmlloyd: For simplicity, I prefer the opposite.  Keep it separate.  Adding default methods makes the compatibility story far more complex.  A separate interface keeps things simple and in cases like this it is a much better way to change behavior without compromising compatibility with old implementations, and it also avoids complexity with new implementations.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
emattheis: If we're going to prioritize performance, then we should be certain there's a clear need for this sort of traceability. Do we have a use case for this one?

dmlloyd: Yes.  In Quarkus we want to be able to report the source and the line number when a configuration validation error occurs.

radcortez: Well at least we wanted to report what is the source that load a certain Config, so having the ConfigSource name is useful.

Priority could help you make decisions when you are intercepting stuff.

LineNumber works for file bases sources, but can work for databases as well. Maybe that can be renamed to a more neutral name.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
radcortez: We implemented this in SmallRye Config and I find it very useful. Its kinda like having the line number in the stacktrace :)

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
emattheis: I can see the motivation, I'm just curious if this is something that users are asking for. I definitely like the idea of identifying the config source that provided the value, but line number is obviously only useful in strictly line-oriented files - so essentially just `.properties` files in our ecosystem - and once you know the file and the property name, the  line number is not terribly helpful for locating the value.

I'm not opposed to including the data point if it's useful for someone, but it seems like, if that is the case, we ought to make it useful in other cases too. For instance, in a YAML or JSON file you might want the column number in addition to the line number. Also, I suppose including the line number presumes that the file name is somehow included in the config source name, but that's not really clear.

As to the object allocation concern, it seems like sacrificing flexibility for performance here is a departure from other design decisions in the API. Can we come up with an optional abstraction and leave implementations free to include their own optimized data points under the covers?

Maybe instead of `int getLineNumber()` we could have `String getSourceLocation()`. Then a source could return `/foo/bar/file.properties (line: 42)` or `/foo/bar/file.json (line: 42, col: 13)` or `jdbc:mysql://foo:3306/bar (table: properties)`, and we can default to `n/a`.

dmlloyd: It doesn't integrate well when you have a diagnostics infrastructure or tooling integration of some sort and you need the line number as a piece of data.  Another problem is that you either have to create a new string on every config property lookup (even if the value isn't used) or else you have to keep another string for every config key.  Keeping an int costs almost nothing.

emattheis: Sure, but if we entertain tooling or diagnostic use cases that need specific information, that's all the more reason to provide a better abstract that can support a richer set of config sources.

If the actual use case is "report the source and the line number when a configuration validation error occurs", then a string suffices.

My main concern here is that if we include `getLineNumber()` in the API, we open the door for users to expect similar context for other config sources without an abstraction to accommodate it.

Maybe we change ConfigValue to an interface (without `getLineNumber()`) and let implementations extend as they see fit.

We could even specify mix-in interfaces with common functionality that implementations could standardize on.

e.g.
```java
public interface LineOriented {
    int getLineNumber();
}
```
```java
public interface FileBased {
    Path getPath();
}
```

dmlloyd: The trouble with mixins is that every config source and interceptor which must construct an instance has to agree to construct the same type; interceptors in particular would have a hard time figuring this out.  A location class is at least better in this regard.

emattheis: Oh, right, I overlooked the the part where interceptor implementors need to construct the value containers. Forcing interceptors to propagate hidden interfaces is certainly not ideal :P Of course, if we're relying on every participant to contribute appropriate tracing information, that's not a very reliable solution anyway.

I'll rest my case here since this thread is getting lengthy for a PR comment. Including lineNumber is not a showstopper for me. Sounds like we're aware of the risks and benefits.

dmlloyd: Well, a location type abstraction is not a showstopper for me either, I guess.  In the worst case it seems like interceptors would be propagating them verbatim from the config source, and hopefully the config source would not have to instantiate them ad-hoc (or if it does, it would be able to cache them) in cases where the config source would otherwise be static (i.e. efficient).

radcortez:
Per our discussions in the meeting on 16/04, I've done the following changes:

- Removed the Interceptor API to focus only on the ConfigValue API (Interceptor API will go into a separate PR)
- Deprecated `ConfigSource` in favour of `ConfigValueSource`

If we are happy with this, I'll proceed to complete the Spec and enhance the TCK.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValueSource.java
```diff
@@ -0,0 +1,185 @@
...
+     * The default implementation of this method looks for a configuration property named
+     * "{@link #CONFIG_ORDINAL config_ordinal}" to determine the ordinal value for this configuration source. If the
+     * property is not found, then the {@linkplain #DEFAULT_ORDINAL default ordinal value} is used.
+     * <p>
+     *
+     * This method may be overridden by configuration source implementations to provide a different behavior.
+     *
+     * @return the ordinal value
+     */
+    default int getOrdinal() {
```
dmlloyd, line 173: Referencing #553 here.

radcortez: Yes, this is just as is copy from the original `ConfigSource`. I'm in favour of #553 for both.

radcortez:
Depends on #553.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -31,29 +31,31 @@
...
+ * {@link ConfigValueSource} with the highest ordinal will be used.
  * <p>
- * Resolves the property value by searching through all configured
- * {@link ConfigSource ConfigSources}. If the same property is specified in multiple
- * {@link ConfigSource ConfigSources}, the value in the {@link ConfigSource} with the highest
- * ordinal will be used.
- * <p>If multiple {@link ConfigSource ConfigSources} are specified with
- * the same ordinal, the {@link ConfigSource#getName()} will be used for sorting.
+ *
+ * If multiple {@link ConfigValueSource ConfigSources} are specified with the same ordinal, the
```
Emily-Jiang, line 51: Since we have ConfigSource and ConfigValueSource, I don't think you should remove the mentioning of ConfigSource. What will happen if the ConfigValueSource and ConfigSource have the same ordinal, which one wins? This needs to be speced.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -90,11 +92,11 @@
  */
 @org.osgi.annotation.versioning.ProviderType
 public interface Config {
-
     /**
-     * Return the resolved property value with the specified type for the
-     * specified property name from the underlying {@linkplain ConfigSource configuration sources}.
+     * Return the resolved property value with the specified type for the specified property name from the underlying
+     * {@linkplain ConfigValueSource configuration sources}.
```
Emily-Jiang, line 97: I think you need to mention ConfigSource as there might be no ConfigValueSource provided.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -112,9 +114,25 @@
...
      * <p>
+     *
+     * The configuration value is not guaranteed to be cached by the implementation, and may be expensive
+     * to compute; therefore, if the returned value is intended to be frequently used, callers should consider storing
+     * rather than recomputing it.
+     *
+     * @param propertyName
+     *              The configuration property name
+     * @return the ConfigValue, or {@code null} if the property is not present
```
Emily-Jiang, line 127: Should null be returned if there is no ConfigValueSource supplied?

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -207,12 +231,13 @@
...
-     * Return all of the currently registered {@linkplain ConfigSource configuration sources} for this configuration.
-     * The returned sources will be sorted by descending ordinal value and name, which can be iterated in a thread-safe manner.
+     * Return all of the currently registered {@linkplain ConfigValueSource configuration sources} for this
+     * configuration. The returned sources will be sorted by descending ordinal value and name, which can be iterated
+     * in a thread-safe manner.
      *
      * @return the configuration sources
      */
-    Iterable<ConfigSource> getConfigSources();
+    Iterable<ConfigValueSource> getConfigSources();
```
Emily-Jiang, line 240: You should not delete the previous one. I think you should add the new method `getConfigValueSources()`

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
```diff
@@ -65,6 +65,14 @@
      */
     ConfigBuilder addDiscoveredConverters();
 
+    /**
+     * Add all interceptors which can be <a href="ConfigSourceInterceptor.html#discovery">discovered</a> from
+     * this configuration builder's {@linkplain #forClassLoader(ClassLoader) class loader}.
+     *
+     * @return this configuration builder instance
+     */
+    ConfigBuilder addDiscoveredInterceptors();
```
Emily-Jiang, line 74: This method should be deleted as the inteceptor bit has been lifted out.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValueSource.java
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * Configuration sources may also be added by defining {@link org.eclipse.microprofile.config.spi.ConfigSourceProvider}
+ * classes which are discoverable in this manner.
+ *
+ * <h3>Closing configuration sources</h3>
+ *
+ * If a configuration source implements the {@link AutoCloseable} interface, then its {@linkplain AutoCloseable#close()
+ * close method} will be called when the underlying configuration is released.
+ *
+ * @author <a href="mailto:radcortez@yahoo.com">Roberto Cortez</a>
```
Emily-Jiang, line 90: For copyright and IP issue, I think you need to lift all contributors from ConfigSource file here.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
jbee, line 74: Can't see the ConfigSourceInterceptor type in master nor in this PR

api/src/main/java/org/eclipse/microprofile/config/Config.java
dmlloyd: It would be exactly the same as having two old style config sources with the same ordinal. It's not a special case.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValueSource.java
dmlloyd: Did you consult with a lawyer?

api/src/main/java/org/eclipse/microprofile/config/Config.java
dmlloyd: Presumably all ConfigSources would be wrapped, so no.

api/src/main/java/org/eclipse/microprofile/config/Config.java
dmlloyd: It really should not matter since they'll be wrapped.

api/src/main/java/org/eclipse/microprofile/config/Config.java
radcortez: Ok!

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
radcortez: Yes, I've missed that one. Removing.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValueSource.java
radcortez: I had no idea what to do there. I'm fine either way.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
radcortez: Yes, this is from an old proposal and forgot to remove this. I'll remove it.

jbee:
Emily made most of the points I was about to make on the detail level. 

On the strategic evolution of the API I generally see the need for a more capable and convenient API. 

Use cases I encounter as "unsolved":
* null-safe and exception-less resolution of properties with fluent API and default values (convenience) 
* access to conversion capabilities that are otherwise not accessible (convenience and added capability)
* conversion to lists and sets of elements (convenience and more clearly standardised)

Use cases I see solved by the PR:
* provide more information on the source of a property values for logging, debugging or presentation purposes

Given the above I can see how this PR can be the basis of an evolution towards the use cases I encounter as "unsolved" but it does not solve any of them yet. Thinking this forward I find the proposal made in this PR somewhat problematic. If sources already operate on the level of property value objects this moves conversion into sources or forces conversion to be bolted on through some form of wrapping or callback which I would find quite unsatisfactory as it either exposes "internals" in the public API or requires to extend the source implementation beyond the public API to implement the collaboration between the `Config` implementation and its sources. A smaller issue I see is the lack of a type parameter for the `ConfigValue` capturing the result type the user wants to have. While this could be added later on it should be in the same release in that case. 

My suggestion is to keep `ConfigSource` as is returning `String` and to extend the interface with an additional method to retrieve  additional source information (as far as I can see in this PR this would only affect the line number). Given such an additional method elevating the `String` returned by a source to a `ConfigValue` can be done by the `Config`. This keeps the conversion and generally control outside of sources and solely within the `Config` implementation. The `ConfigValue` should only be defined in terms of its capabilities (interface), not its implementation limiting ways to implement features like conversion and default fallbacks.

dmlloyd:
> Emily made most of the points I was about to make on the detail level.
> 
> On the strategic evolution of the API I generally see the need for a more capable and convenient API.
> 
> Use cases I encounter as "unsolved":
> 
> * null-safe and exception-less resolution of properties with fluent API and default values (convenience)
> * access to conversion capabilities that are otherwise not accessible (convenience and added capability)
> * conversion to lists and sets of elements (convenience and more clearly standardised)

None of these are use cases though the last directly implies a use case which is addressed by a different issue.

Null-safe resolution already should be a reality.  What do you see as missing here?  What do you mean by "exception-less"?

What conversion capabilities are you alluding to?  What is the use case (i.e. what are you trying to do, as opposed to how do you imagine it being done)?

> Use cases I see solved by the PR:
> 
> * provide more information on the source of a property values for logging, debugging or presentation purposes
> 
> Given the above I can see how this PR can be the basis of an evolution towards the use cases I encounter as "unsolved" but it does not solve any of them yet. Thinking this forward I find the proposal made in this PR somewhat problematic. If sources already operate on the level of property value objects this moves conversion into sources or forces conversion to be bolted on through some form of wrapping or callback which I would find quite unsatisfactory as it either exposes "internals" in the public API or requires to extend the source implementation beyond the public API to implement the collaboration between the `Config` implementation and its sources. A smaller issue I see is the lack of a type parameter for the `ConfigValue` capturing the result type the user wants to have. While this could be added later on it should be in the same release in that case.

Don't worry, we're not in a rush to release.  We have consciously tried to make this PR be incremental for the purpose of segmenting debate and preventing a situation where conversation is confused.

> My suggestion is to keep `ConfigSource` as is returning `String` and to extend the interface with an additional method to retrieve additional source information (as far as I can see in this PR this would only affect the line number). Given such an additional method elevating the `String` returned by a source to a `ConfigValue` can be done by the `Config`. This keeps the conversion and generally control outside of sources and solely within the `Config` implementation. The `ConfigValue` should only be defined in terms of its capabilities (interface), not its implementation limiting ways to implement features like conversion and default fallbacks.

We discussed this already.  The main reason `ConfigValue` is being added is to support multiple other enhancements to the API to meet a complex, interleaving set of use cases.  It exists for SPIs just as mach as for APIs, if not more so.  The SPIs are still being introduced.  It's a chicken and egg problem: you can't bring in the more complex changes without their predecessors, and the predecessors cannot support the more complex changes if you design and integrate them in a short-term way.

Note that conversion is not (yet) part of the `ConfigValue` story.

api/src/main/java/org/eclipse/microprofile/config/Config.java
Emily-Jiang: It needs to be speced and ConfigSource mention should not be removed.

api/src/main/java/org/eclipse/microprofile/config/Config.java
Emily-Jiang: I started thinking just reuse ConfigSource and add methods instead, which is much cleaner.

Emily-Jiang:
> My suggestion is to keep ConfigSource as is returning String and to extend the interface with an additional method to retrieve additional source information (as far as I can see in this PR this would only affect the line number). Given such an additional method elevating the String returned by a source to a ConfigValue can be done by the Config. This keeps the conversion and generally control outside of sources and solely within the Config implementation. The ConfigValue should only be defined in terms of its capabilities (interface), not its implementation limiting ways to implement features like conversion and default fallbacks.

I was in the last week's discussion. After I reviewed the updated PR and had a further thought on this, I changed my mind and agreed with @jbee before I even read this above comment. My main concern is that it creates a lot of confusions with two classes live next to each other. Besides, it is ok to add methods to ConfigSource and update as we wish. We can just increate the major version for this package.

Emily-Jiang:
@jbee when @radcortez adds another method `withDefault()` on  `ConfigValue`, your use case of setting default will be addressed.

dmlloyd:
I think it is probably better to break compatibility and make ConfigSource use ConfigValue in that case.

jbee:
> adds another method withDefault() on ConfigValue, your use case of setting default will be addressed.

@Emily-Jiang only partly. Default in combination with conversion is the use case. The one arising from `@ConfigProperty` requires to be able to set the default as `String` but then resolving the value as converted value. This can be e.g. `Integer` (use `Converter<Integer>`) or `Integer[]` (split string first and use use `Converter<Integer>` on each element). So this is much closer to the https://github.com/struberg/microprofile-config/blob/ConfigAccessor/api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java @struberg sketched. With converters exposed this could be done by the user of the API but it is very inconvenient to let the user go through this each time.

jbee:
> My main concern is that it creates a lot of confusions with two classes live next to each other. Besides, it is ok to add methods to ConfigSource and update as we wish

Agree, adding methods to `ConfigSource` would be a much smaller change with better long term backwards compatibility and an API that is easier to understand as responsibilities are more clear.

With conversion added to capabilities of `ConfigValue` it would further indicate a mismatches with making the creation of `ConfigValue` a responsibility of the sources. Conversion isn't their responsibility. If it remains in `Config`'s responsibility but sources create the `ConfigValue`s there are two places which have to be involved which I find a path to a confusing solution that has to jump through some hoops to work.

The issue I have with  `ConfigValue` being a class, especially when originating from sources is that all data has to be filled all the time needed or not. Making it an interface originating from `Config` implementation the implementations has some options on how to implement it. It can defer lookup of line number data to the point the user requests this from the `ConfigValue`. It also narrows the `ConfigValue` API surface as no builder pattern is needed. The API would only offer methods to work with the value and receive information on the property all of which can be done eager or lazy. Making this a problem of the implementation also means the implementation can chose the solution that works best with other aspects used by the implementation like caching of values.

api/src/main/java/org/eclipse/microprofile/config/Config.java
radcortez: Idea was to wrap the old ConfigSource in the new ConfigValueSource, so we wouldn't need to refer to both of them all the time.

api/src/main/java/org/eclipse/microprofile/config/Config.java
radcortez: Idea was to wrap the old ConfigSource in the new ConfigValueSource, so we wouldn't need to refer to both of them all the time.

radcortez:
> Agree, adding methods to `ConfigSource` would be a much smaller change with better long term backwards compatibility and an API that is easier to understand as responsibilities are more clear.

I understand. On the other hand, we want implementors to implement the `ConfigValue` style methods, so if we add them as default methods, no one is going to implement them (or very few). If we added them as methods to be implemented we are breaking compatibility anyway, so that is why we added a separate interface. 

> 
> With conversion added to capabilities of `ConfigValue` it would further indicate a mismatches with making the creation of `ConfigValue` a responsibility of the sources. Conversion isn't their responsibility. If it remains in `Config`'s responsibility but sources create the `ConfigValue`s there are two places which have to be involved which I find a path to a confusing solution that has to jump through some hoops to work.

I do think that `ConfigSource` and `ConfigValue` shouldn't have to deal with any conversion details. I look into `ConfigSource` has a raw data source and `ConfigValue` as metadata to my raw data. I would prefer that conversions is looked as a separate step, especially considering that we discussed a few times to separate Conversion and make a separate spec out of it, since other projects also require it.

> 
> The issue I have with `ConfigValue` being a class, especially when originating from sources is that all data has to be filled all the time needed or not. Making it an interface originating from `Config` implementation the implementations has some options on how to implement it. It can defer lookup of line number data to the point the user requests this from the `ConfigValue`. It also narrows the `ConfigValue` API surface as no builder pattern is needed. The API would only offer methods to work with the value and receive information on the property all of which can be done eager or lazy. Making this a problem of the implementation also means the implementation can chose the solution that works best with other aspects used by the implementation like caching of values.

I'm ok to change `ConfigValue` to an interface, but we do need to agree on the API.

jbee:
> I do think that ConfigSource and ConfigValue shouldn't have to deal with any conversion details.

Generally I support the thought of this being concepts that are dealing with another domain, the domain of configuration, and thus should not be involved with conversion. Without conversion though I'd also think the API is unreasonable inconvenient to use. While discussing conversion is a topic on its own I find it relevant here to the extend that it has to fit in with where values are coming from and what that implies for what you can do with them and what you can't or at least what gets tricky or dirty to implement. In that sense I feel the topic of conversion is relevant here. 

> ... make a separate spec out of it, since other projects also require it.

In general this seems a reasonable thing to do. But given the very limited nature of the conversion MP config offers I'd think it makes for a rather poor general conversion standard. In context of configuration I can see how its limitations are a reasonable way but to lift this to a conversion standard would mean it has to become more capable at which point I'd wonder why such a standard would start from the conversion done in MP config as the hard parts are everything beyond the basic capabilities MP config has.

radcortez:
> In general this seems a reasonable thing to do. But given the very limited nature of the conversion MP config offers I'd think it makes for a rather poor general conversion standard. In context of configuration I can see how its limitations are a reasonable way but to lift this to a conversion standard would mean it has to become more capable at which point I'd wonder why such a standard would start from the conversion done in MP config as the hard parts are everything beyond the basic capabilities MP config has.

Yes, a Conversion spec must offer way more and the one in Config is very limited. I don't say that it needs to start with Config, but we definitely need something. Other projects in MP like JWT also require conversions, so it is a waste that this problem is being tackled individually in each spec instead of globally. Then you end up with Conversion API's for JPA, JAXB, JsonB, Config, JWT, etc. Probably being too optimistic thinking that we can solve this issue :)

Anyway, ideally, we should design our API's in a way that if we move forward with it, we are able to replace current Conversion mechanism with another one. 

Or we can just forget about it and do what we think is better for Config in particular.

jbee:
> ideally, we should design our API's in a way that if we move forward with it, we are able to replace current Conversion mechanism with another one.

Good point. That is why I think conversion should be transparent e.g. behind the `ConfigValue` abstraction and not delegated to the user (e.g. let the user access `Converter`s from the `Config` would "bind" the user code to a specific conversion API) .

radcortez:
Maybe we need a new abstraction, `ConvertedValue`? :D

dmlloyd:
`ConfigValue` could support a `<T> T getValue(Converter<T> conv)` method today without any problem.  Supporting `<T> T getValue(Class<T> clazz)` is more problematic: the `ConfigValue` construction path would have to be such that it retains a reference to `Config` or its corresponding internals to allow the converter for the given type to be queried.  Not impossible, but yet another change.

radcortez:
How about some sort of an hybrid API? We can add specific methods for the most common types like `getAsInt`, `getAsBoolean`, etc. and a `getValue(Converter)` for specific cases?

smillidge:
At the most fundamental level I can't review this PR as I can't see any link that describes what it is trying to solve.

radcortez:
@smillidge This has several iterations and actually started has something different. A lot was discussed in the zoom meetings and we probably failed to provide enough context for people that didn't attend them.

I'll link some issues in the PR description to give more detail.

jbee:
@dmlloyd , @radcortez good to see you giving thoughts to the aspect of how users can access converted values. The critical (because current unsupported) combination is in connection with a not yet converted default. This is not just required to implement the CDI but also something users can benefit from as the code needed to extract values handling all cases can be reduced immensely with an API that is aware of defaults and conversion.

The suggested way of passing the `Converter` to the `ConfigValue` is moving most of the "work" to the user making this clunky and binding the user code to the converter API which isn't a good move when we already anticipate a common converter API for MP.

Having not been present at the meetings and discussion around the original proposal including the interceptors I am also having a hard to understand what user level problem it was solving. From what I can understand you already have the interceptors in your vendor implementation as a solution to vendor specific features or characteristics. What is it that makes this important to lift from a vendor solution to a standard capability?

radcortez:
> @dmlloyd , @radcortez good to see you giving thoughts to the aspect of how users can access converted values. The critical (because current unsupported) combination is in connection with a not yet converted default. This is not just required to implement the CDI but also something users can benefit from as the code needed to extract values handling all cases can be reduced immensely with an API that is aware of defaults and conversion.
> 

Hey, we are here to come up with solutions that help everyone :)

> The suggested way of passing the `Converter` to the `ConfigValue` is moving most of the "work" to the user making this clunky and binding the user code to the converter API which isn't a good move when we already anticipate a common converter API for MP.
> 

I think it wouldn't harm to have a `.getValue(Converter)` while still supporting all the standard types and registered converters. Consider.

```java
ConfigValue value = ...
// ops I don't have a Converter for address, now I need to go and register one, create another class and add a service loader file.
Address address = value.getValue(value -> ... )
```

So I think it is useful for those cases where you have a single conversion custom point and you are lazy enough to go and implement a separate class :)

> Having not been present at the meetings and discussion around the original proposal including the interceptors I am also having a hard to understand what user level problem it was solving. From what I can understand you already have the interceptors in your vendor implementation as a solution to vendor specific features or characteristics. What is it that makes this important to lift from a vendor solution to a standard capability?

I've already retracted the interceptors part, so I wouldn't worry too much about it. It is a separate discussion. They are not required, make things way easier imo. We are now discussing to add features like profiles, variable expansion, etc. I think this can be easily implemented with interceptors. Yes, they can be implemented in other ways. 

Fundamentally, I think that MP Config is lacking in extension points where you can hookup code to make life easier for implementations and even users :)

radcortez:
BTW, linked some issues in the PR description to give more context. Also, I would like to remember everyone that this is only a Draft PR to start the discussion and reach a solution that we are comfortable with. 

Nothing is set into stone and all ideas and feedback are welcomed :)

api/src/main/java/org/eclipse/microprofile/config/Config.java
Emily-Jiang: I think it is much simpler and easier to add methods to ConfigSource.

api/src/main/java/org/eclipse/microprofile/config/Config.java
Emily-Jiang: no. I suggest not to introduce ConfigValueSource.

Emily-Jiang:
> I understand. On the other hand, we want implementors to implement the ConfigValue style methods, so if we add them as default methods, no one is going to implement them (or very few). If we added them as methods to be implemented we are breaking compatibility anyway, so that is why we added a separate interface.

no. Don't add default impl. We are doing config 2.0 release and this is the ideal time to introduce this.

api/src/main/java/org/eclipse/microprofile/config/Config.java
radcortez: As default methods or just abstract methods?

Emily-Jiang: abstract methods

radcortez: Seems more reasonable, since we want to force implementations to do it, but also it means we are going to break everyone :D

Emily-Jiang: This is why we release Config 2.0 not Config 1.5. We have to go one way or the other. If we lean more for customers to implements the new method, we need to make it abstract unless we can sufficiently implement it.

radcortez: I'm OK with this. Is just that we were trying to prevent it and in the end we just came back to the same place :)

radcortez:
Updated this with the latest proposal:
- Add new abstract methods that return `ConfigValue` to `ConfigSource` directly. This means that we will break compatibility.
- `getValue` in `ConfigSource` is now default based on `getConfigValue`
- Deprecated `getProperties` in `ConfigSource`
- Added `ConfigValue getConfigValue(String)` in `Config`. This may not be the required, depending on the result of the discussion about supporting Conversion in the API directly.
- We do need to agree on `ConfigValue` structure as well. No modification for now, since I think we should agree on `ConfigSource` changes and then we can look into `ConfigValue` with more detail.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
```diff
@@ -0,0 +1,161 @@
...
+                value.equals(that.value) &&
+                configSourceName.equals(that.configSourceName);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, value, configSourceName);
+    }
+
+    ConfigValueBuilder from() {
```
Emily-Jiang, line 112: How useful is this?

Emily-Jiang:
Thank you @radcortez and it looks much more simplier and easy to be understood!

> Updated this with the latest proposal:

    Add new abstract methods that return ConfigValue to ConfigSource directly. This means that we will break compatibility.

It is ok as we are permitted to release MP Config 2.0, containing backward incompatible change.

    getValue in ConfigSource is now default based on getConfigValue
    Deprecated getProperties in ConfigSource
    Added ConfigValue getConfigValue(String) in Config. This may not be the required, depending on the result of the discussion about supporting Conversion in the API directly.

ok
    We do need to agree on ConfigValue structure as well. No modification for now, since I think we should agree on ConfigSource changes and then we can look into ConfigValue with more detail.

I have one concern on the implementation of lineNumber. How can you easily supply this? I feel it is pretty difficult for all properties. If no one populates, it is a waste and we should not include it for now. The other thought I have for debugging purpose of configValue, maybe the location instead of configsource name is a better indicator. Thoughts?

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
radcortez: Yes, maybe we need to recheck the visibly modifiers. I'll have a look

radcortez:
Maybe we can add a Location type abstraction to hide these details and allow for more extension points. The idea was already discussed in previous comments and seems that there might be some agreement on that.

Emily-Jiang:
@radcortez since the ConfigValue is a purely metadata. What do you think of renaming it to be PropertyMetaData/ConfigMetaData or something along the similar line?

dmlloyd:
The specification doesn't define the term "metadata", but regardless, I can't imagine any reasonable definition of that term that could be fulfilled in any way by `ConfigValue`.  The class is aptly named AFAICT.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -109,14 +109,13 @@
      *
      * @return a map containing properties of this configuration source
      */
+    @Deprecated
```
struberg, line 112: What would replace this? A @Deprecated annotation without any explanation is always bad. Pretty much prefer the JavaDoc @deprecated

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -179,7 +178,18 @@ default int getOrdinal() {
      * @param propertyName the property name
      * @return the property value, or {@code null} if the property is not present
      */
-    String getValue(String propertyName);
+    default String getValue(String propertyName) {
```
struberg, line 181: we did not do this due to performance implication.
This also kind of turns the whole design upside down.
The ConfigValue is supposed to be ON TOP of the 'native raw config' which `getProperties` and `getValue` represent.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
```diff
@@ -0,0 +1,161 @@
...
+ * Right now, it is able to hold information like the configuration name, value, the Config Source from where
+ * the configuration was loaded, the ordinal of the Config Source and a line number from where the configuration was
+ * read if exists.
+ * <p>
+ *
+ * This is used together with {@link ConfigSource} to expose the Configuration lookup metadata.
+ *
+ * @author <a href="mailto:radcortez@yahoo.com">Roberto Cortez</a>
+ */
+public class ConfigValue {
```
struberg, line 37: You have to be careful when choosing names which already have a pretty established meaning in the mp-config and ConfigJSR area. Like ConfigValue which later got renamed to ConfigAccessor. People will be terribly confused I fear.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
```diff
@@ -0,0 +1,161 @@
...
+
+import java.util.Objects;
+
+/**
+ * The ConfigValue is a metadata object that holds additional information after the lookup of a configuration.
+ * <p>
+ *
+ * Right now, it is able to hold information like the configuration name, value, the Config Source from where
+ * the configuration was loaded, the ordinal of the Config Source and a line number from where the configuration was
+ * read if exists.
```
struberg, line 30: What do you need these kind of metadata for?
The only use case we had for this information so far is a summary page which kind of enlists all the ConfigSources and values therein for diagnostic reasons. But this can be perfectly done even with the mp-config-1.0 API already. So what is the additional value of completely changing the whole design (in likely backward incompatible ways)

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
struberg: why would the line number matter at all for the end user use case? Any developer would just look at the file. Also note that in yaml, json or xml you might not even be able to define a line number for the resulting key because the value is a result of information from many lines together.

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/SampleYamlConfigSource.java
```diff
@@ -46,6 +47,16 @@ public String getValue(String key) {
         return config.get(key);
     }
 
+    @Override
+    public ConfigValue getConfigValue(final String propertyName) {
+        return ConfigValue.builder()
```
struberg, line 52: That just doesn't look very well performing to be honest!

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
Emily-Jiang: This will not be deprecated any more after the further conversation. @radcortez please correct this.

OndroMih:
> On the other hand, we want implementors to implement the ConfigValue style methods

What's the rational behind this, @radcortez , @Emily-Jiang ? I vehemently disagree with this. Why we should replace a method that simply returns string with a method that constructs an object and potentially introduces a few other method calls to construct it, and then force users to implement it?

It doesn't make sense to me. In most cases, users are interested only in the config value and not in anything else. So it doesn't make sense to force all this on them.

Another thing is that this proposal is **unnecessarily complicated**. For example, `ConfigValue` could simply be an interface, leaving all implementation details out of the spec. There's no need for a ConfigValueBuilder in the API and also no need for getters and especially setters. Another example is adding `lineNumber` support. This is completely orthogonal to ConfigValue API and should be discussed separately in another proposal.

I like the idea of ConfigValue API but I really want to make it simple and without implementation details.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
radcortez: Other libraries do report the line number like JAXP, which is useful for debug diagnosis. Yes, you can look into the file, but again I think it is a nice addition. Anyway, this has caused a lot of discussion back and forward, so I'll most likely remove it.

Also, I believe the reason why we are adding a new method to `ConfigSource` has nothing to do with the `lineNumber`. In my opinion, the `ConfigSource` should have a way to return an object view with metadata information about the lookup. Right now all the information can be acquired by existing methods, but that doesn't mean that additional information may not be included in the future.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
radcortez: The reason why we discussed this to be deprecated is because, potentially computing the getValue for all the properties in the ConfigSource might be too expensive.

This has been discussed before:
https://github.com/eclipse/microprofile-config/issues/431

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
radcortez: It's fine, we can remove the default implementation.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
radcortez: I'm terrible for naming :D

I didn't come up with the name myself. It was reference like that through our meetings and there are multiple references across other issues.

Do we want to name this ConfigAccessor then?

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
radcortez: Well, we have to start somewhere. Yes, the initial use case would be to provide debug information about the key lookup. Idea was to expand this in the future to add more things.

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/SampleYamlConfigSource.java
radcortez: This is just a TCK and it was an easy fix to compile the project.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
```diff
@@ -0,0 +1,161 @@
...
+                .withConfigSourceName(configSourceName)
+                .withConfigSourceOrdinal(configSourceOrdinal)
+                .withLineNumber(lineNumber);
+    }
+
+    public static ConfigValueBuilder builder() {
+        return new ConfigValueBuilder();
+    }
+
+    public static class ConfigValueBuilder {
```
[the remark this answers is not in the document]
radcortez: If we expose a `ConfigValue` method in `ConfigSource`, don't we need a way for users to create their own object?

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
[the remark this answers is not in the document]
radcortez: We can do as many changes as we like.

The main reason why `ConfigValue` is a class is so users can do what they like with it and implement it in `ConfigSource`. Similar to `Response` in JAX-RS.

There is some discussion about adding this or not to `ConfigSource` so we probably need to resolve that first, because it will help clear out the outcome of this.

radcortez:
> Why we should replace a method that simply returns string with a method that constructs an object and potentially introduces a few other method calls to construct it, and then force users to implement it?

The idea was not to replace, but to allow users to push whatever metadata makes sense into ConfigValue for the particular ConfigSource they are implementing.

Right now, it does seems a waste and I agree, because all of the information can be retrieved from already existent methods, but we need to start from somewhere. I think it makes sense to add what we already have and then build on top of it to add new stuff.

@OndroMih one thing that is not clear to me and please clarify me, is where do you think that a `ConfigValue` API should hook in Config? From what I understood, you disagree to add a new method to return the `ConfigValue` in `ConfigSource`. Does it only live in `Config`?

OndroMih:
Yes, I believe that `ConfigValue` should initially live only in `Config` and shouldn't be provided by config sources. At least for an initial proposal. We can have another discussion (another PR later) about adding it also to ConfigSources. Besides line numbers, which many people suggest to remove from this proposal, it's not needed to modify the `ConfigSource` interface (and introduce a breaking change) to support `ConfigValue` in the `Config` interface.

dmlloyd:
I believe that two different things are being proposed here.  "Moving" `ConfigValue` to `Config` in this proposal essentially eliminates the original use cases for this proposal and addresses different use cases instead.  Once again, a useful enhancement is defeated by confusion and misunderstanding.

Before suggesting any more changes to the code, I think it would be a good idea to review each of the issues that are being solved by the proposal (listed up top).  For any given change proposal, if the corresponding use cases are still solved by that change, then go ahead and suggest it.  If the use case is invalidated, then I'd say the proposal is not appropriate.  In that case the next step would be to address any questions or problems with the use case on the use case issue, and to propose alternative implementation options as new issues or PRs.

radcortez:
> Yes, @radcortez , I believe that `ConfigValue` should initially live only in `Config` and shouldn't be provided by config sources. At least for an initial proposal. We can have another discussion (another PR later) about adding it also to ConfigSources. Besides line numbers, which many people suggest to remove from this proposal, it's not needed to modify the `ConfigSource` interface (and introduce a breaking change) to support `ConfigValue` in the `Config` interface.

I understand. 

My view on this is that `ConfigSource` without `ConfigValue` may seem incomplete. Remember, you can call `org.eclipse.microprofile.config.Config#getConfigSources` to iterate and call ConfigSources directly. If we make `ConfigValue` an interface then there is no way for the end user to get a `ConfigValue`, unless he creates the object with some vendor based API. 

Do we think it makes sense if I retrieve a config from `Config` I can get a `ConfigValue`, but if I use a `ConfigSource` directly, I cannot do it? I'm not saying one is right or the other is wrong, is just something for us to consider and discuss.

OndroMih:
Re: @dmlloyd 
I've review all 5 issues referred in the initial description of this PR and I'm sure all can be addressed by having `ConfigValue` just in `Config`. Providing `ConfigValue` at the `ConfigSource` level doesn't add any additional value related to those issues so I don't see how the enhancement would be defeated by this change.

Re: @radcortez 
Yes, I believe that `ConfigValue` at the `Config` level makes perfect sense, while having it at `ConfigSource` level makes very little sense. At `Config` level, it would provide finer information about underlying config sources (e.g. the list of sources that provide the value). On `ConfigSource` level, there's not much to provide on top of existing methods in `ConfigSource`. Except the need for adding debugging information like line number, which is solely not enough to justify adding so much complexity and I believe we can find a better solution for it later. I would rather like to agree on something simple and justified and later have a more detailed discussion about what can a `ConfigValue` provide at the `ConfigSource` level to justify adding it there. 

Ad. interface vs. class: In API, interfaces are always preferred over classes to allow more flexibility for implementations. There's no problem with instantiation as any interface A can easily be instantiated using `new A() { ... }`. We can always provide a helper abstract class or a builder if needed. But if `ConfigValue` is an interface and only available at `Config` level, users have no need to instantiate it, only consume it.

radcortez:
Ok, I'm proposing a simplified version that hopefully we can agree on so we can move forward and later on, build stuff on top.

Emily-Jiang:
I took a look at the simple APIs and it looks good. I believe it should be able to address #43 #312 . I believe the next step is to spec the behaviour via another commit. Thanks @radcortez !

radcortez:
Let's wait for additional feedback. If we have more consensus I'm happy to add the missing pieces.

Emily-Jiang:
Hi @radcortez, 
@OndroMih and I met and discussed the current status. @OndroMih is happy with the current changes.
please go ahead to add the missing piece.
Thanks

radcortez:
Ok @Emily-Jiang . Thank you.

I believe we only need a few TCK, right? Is there anything out of this that we want to add into the spec?

Emily-Jiang:
> I believe we only need a few TCK, right? Is there anything out of this that we want to add into the spec?

Both spec and TCKs need to be added.

radcortez:
Ok, but what do we want to add in the spec? This doesn't really change anything spec related. just a couple of new API's.

Emily-Jiang:
@radcortez [This section](https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configexamples.asciidoc) needs to be updated to demonstrate the ConfigValue usage.
By the way, I think we should consider the injection of ConfigValue as well.
`@Inject @ConfigProperty(name="client.name") ConfigValue client;

Thoughts?

radcortez:
Never thought about injection of `ConfigValue`. I don't see any reasons to not do it, but lets wait for some more feedback.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
struberg: #431 is about the return type of the property name method, not getValue. Could you please elaborate what it has to do with this very line?

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
struberg: Why not look at work already done. This sadly has a long history with a ton of existing IP.
The original work of accessing config values was actually done in DeltaSpike many years ago (2011). It was also part of the original mp-config work I brought over in 1.0. But back then it was deemed to be 'probably too much in one step'. So we postponed it and reviewed it for ConfigJSR as ConfigValue:
https://github.com/eclipse/ConfigJSR/commit/a5757192d1d11da7e8602e56a337931d6a169541#diff-84c09a90e39a639354bfc30e2100d487

it later got renamed to ConfigAccessor it would be a better fit to what it actually does
https://github.com/eclipse/ConfigJSR/commit/2021eab512c1f5999272d0ac866619868e9d89da#diff-2334b9d3b42caf9078203da0f6f7d82f

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
radcortez: This was already removed from the PR, so it doesn't make sense to discuss it anymore.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigValue.java
radcortez: This was already reviewed and simplified in the current PR. Please check if you agree with the current proposal, regarding the API (only 4 methods for now, to retrieve the property name, value, source from where it was loaded and ordinal). 

Regarding the name, it was always discussed like this (at least in the conversions I've participated). I have no issues in naming it `ConfigAccessor` if everyone else is ok with it.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,64 @@
...
+     * The name of the lookup property.
+     *
+     * @return the name of the lookup property.
+     */
+    String getName();
+
+    /**
+     * The value of the config lookup.
+     *
+     * @return the raw value of the config lookup or {@code null} if the property could not be found.
```
jbee, line 46: This should clarify if this is the value at the point in time the `ConfigValue` object was looked up or if it is the value when `getValue()` is called.

radcortez: Yes, it needs to be when the `ConfigValue` was looked up, because of the other methods. Let me add a clarification for that.

Not so sure about adding a `getTime()` method.

Emily-Jiang: I don't think we should add `getTime()` method as this `getValue()` is no difference from the `config.getValue()`. The return result of the `getValue()` should always be the updated value not cached value if a runtime supports dynamic config. The ConfigValue should be treated as a proxy.

radcortez: Do you mean that when you call `getValue` on `ConfigValue` is when the lookup is performed? What about the other methods? What would happen if you call `getSourceName` before calling `getValue?

Emily-Jiang: As mentioned in our call earlier, I am fine with making configValue as immutable as getConfigVaue will retrieve the updated value.

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,64 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package org.eclipse.microprofile.config;
+
+/**
+ * The ConfigValue holds additional information after the lookup of a configuration and is itself immutable.
```
Emily-Jiang, line 23: a configuration => a configuration property

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,64 @@
...
+ * <p>
+ *
+ * Holds information about the configuration name, configuration value, the
```
Emily-Jiang, line 26: configuration => configuration property

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,64 @@
...
+ * {@link org.eclipse.microprofile.config.spi.ConfigSource} name from where
+ * the configuration was loaded and the ordinal of the {@link org.eclipse.microprofile.config.spi.ConfigSource}.
```
Emily-Jiang, line 28: configuration => configuration property

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,64 @@
...
+ * <p>
+ *
+ * This is used together with {@link Config} to expose the configuration lookup metadata.
```
Emily-Jiang, line 31: configuration => configuration property

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,64 @@
...
+public interface ConfigValue {
+    /**
+     * The name of the lookup property.
+     *
+     * @return the name of the lookup property.
+     */
+    String getName();
+
+    /**
+     * The value of the config lookup.
```
Emily-Jiang, line 44: config => property

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @return the raw value of the config lookup or {@code null} if the property could not be found.
```
Emily-Jiang, line 46: config lookup => property lookup 
so that it is consistent.

spec/src/main/asciidoc/configexamples.asciidoc
```diff
@@ -38,8 +38,11 @@ public class ConfigUsageSample {
         Config config = ConfigProvider.getConfig();
 
         String serverUrl = config.getValue("acme.myprj.some.url", String.class);
-
         callToServer(serverUrl);
+
+        // or
+        ConfigValue configServerUrl = config.getConfigValue("acme.myprj.some.url");
+        callToServer(serverUrl.getValue());
```
Emily-Jiang, line 45: serverUrl -> configServerUrl

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigValueTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertNull;
+
+public class ConfigValueTest extends Arquillian {
+    @Deployment
+    public static Archive deployment() {
+        return ShrinkWrap.create(WebArchive.class)
+                         .addAsServiceProvider(ConfigSource.class, ConfigValueConfigSource.class, ConfigValueBean.class)
```
Emily-Jiang, line 47: In all of other tests, we use the following:

```
@Deployment
    public static WebArchive deploy() {
        JavaArchive testJar = ShrinkWrap
                .create(JavaArchive.class, "arrayConverterTest.jar")
                .addPackage(PizzaConverter.class.getPackage())
                .addClasses(ArrayConverterTest.class, ArrayConverterBean.class)
                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
                .addAsServiceProvider(Converter.class, PizzaConverter.class)
                .as(JavaArchive.class);
            addFile(testJar, "META-INF/microprofile-config.properties");
            WebArchive war = ShrinkWrap
                .create(WebArchive.class, "arrayConverterTest.war")
                .addAsLibrary(testJar);
            return war;
    }
```
I am not sure wheter this test can run in other runtime. Does the archive contain all of the classes needed? I suggest to use the same style as other tests.

spec/src/main/asciidoc/configexamples.asciidoc
```diff
@@ -81,6 +84,12 @@ public class InjectedConfigUsageSample {
     @ConfigProperty(name="myprj.some.url")
     private String someUrl;
 
+    // You can also inject a configuration using the ConfigValue metadata object. The
+    // configured value will not lead to a DeploymentException if the value is missing.
```
Emily-Jiang, line 88: also add a line to say defaultValue can also be specified.

Emily-Jiang:
@radcortez as promised, I have another review and logged some minor comments. Overall, it looks good.

radcortez:
Thank you. I think I've addressed all of your comments.

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigValueTest.java
```diff
@@ -0,0 +1,145 @@
+/*
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
```
jbee, line 2: 2020?

radcortez: Ops. Copy paste :)

radcortez:
Just rebased and squashed to make it more clean.

Emily-Jiang:
@OndroMih @struberg please re-review asap and I'd like this PR merged this week if possible.

radcortez:
They are tested here:
https://github.com/radcortez/microprofile-config/blob/535aca922394b69599726db0db8e85fc21347cea/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigValueTest.java#L86-L89

I can another source anyway.

Emily-Jiang:
@radcortez feel free to merge this PR!

Emily-Jiang:
I believe @radcortez has address the feedback from @OndroMih and @struberg ! This PR has been open for a while without further feedback. I'm merging this PR because of @radcortez has not got permission to merge the PR as yet.

OndroMih:
Hi, @emily-jiang, @radcortez, I didn’t have time to review before you merged this but for the record I approve now. I had a look and it really reflects all my comments, thank you!

Issue #418, opened 2019-03-12
Support configuration profiles


gastaldi:
It would be nice if the spec allowed different profiles to be used when the application starts (eg. development, production, testing). The idea is that a different set of configuration properties would be available based on a flag (eg. `-Dprofile=production`).

In an hypothetical example, you could have a `META-INF/microprofile-config.properties` file with the default parameters and a `META-INF/microprofile-config-production.properties` with values used when the `production` value is specified which would override the default values.

struberg:
Hi @gastaldi ! This is similar to what we had in Apache DeltaSpike ProjectStage, right?
http://deltaspike.apache.org/documentation/projectstage.html
It is rather easy to implement yourself with the freshly proposed `ConfigAccessor#addLookupSuffix`.
Ofc a bit more typing as with DeltaSpike. But as flexible.

gastaldi:
@struberg yeah, that seems like a good solution. It would be nice if that were part of the spec instead of being an implementation detail.

pilhuhn:
If implemented, please support some prefix for the property name like `-Dconfig.profile` and not only `-Dprofile`.

Emily-Jiang:
I thougth this further. Maybe set the stage as part of property name is a better approach.
-Dstage=production
//stage=production
%production.name=hello
%live.name=world

Emily-Jiang:
I started working on this and gave it more thought based on what @pilhuhn said.

- In order to avoid the name collison, I suggest use `mp.config.profile` or `mp.config.stage`. 

- Should we also support microprofile-config.properties and microprofile-config-testing.properties. I think many Spring users got used to this style and it is kind of easy fix. We only support microprofile-config.property but not other config sources. 

Thoughts?

gastaldi:
> I started working on this and gave it more thought based on what @pilhuhn said.
> 
> 1. In order to avoid the name collison, I suggest use `mp.config.profile` or `mp.config.stage`.

+1, I like `mp.config.profile`

> 2. Should we also support microprofile-config.properties and microprofile-config-testing.properties. I think many Spring users got used to this style and it is kind of easy fix. We only support microprofile-config.property but not other config sources.

I don't have a preference here, I like different files as long as the additional properties file is used to override properties in the original config (a higher priority over the `microprofile-config.properties` ConfigSource). OTOH having part of the property name may be easier to maintain (that's what we have been using in Quarkus BTW).

Emily-Jiang:
[...]

Issue #553, opened 2020-04-07
"Ordinal" versus "Priority" is confusing


dmlloyd:
**Describe the bug**

For `ConfigSource` we talk about "ordinal", which is a word that means "relating to position in a series", for example (first, second, third, etc.).  For other things we use "priority" (and the `@Priority` annotation).  This is very confusing because when we talk about ordinals in language, the thing that is first is first, but we're now treating `ConfigSource` ordinality the reverse way: higher (or later) ordinals take effect over lower (or earlier) ordinals.

We should change all specification language referring to "ordinal" to say "priority" instead.  We should also look into using `@Priority` for `ConfigSource` by default (possibly as a follow-up issue).

radcortez:
The pain point here will be the `ConfigSource#getOrdinal` method.

OndroMih:
We could add a new default method `ConfigSource#getPriority` which will call `ConfigSource#getOrdinal`. `getOrdinal` is already implemented as `default` and thus optional.

Old config sources would work and new config sources can just define `getPriority`.

tomas-langer:
I am not sure I follow - `Priority` works (at least in most meanings) as "the lower number wins" - hence the term "Number one priority" (see https://javaee.github.io/javaee-spec/javadocs/javax/interceptor/Interceptor.Priority.html as an example).
So how would replacing the `ordinal` with `priority` fix this issue? 
The specification is inverse to usual understanding of priority and ordinal. So unless this is changed (which would be a big backward incompatible change), it does not really matter if you call it `ordinal` or `priority`.

radcortez:
One of the issues is that we use ordinal for `ConfigSource` but priority for `Converter`, which is confusing in my opinion. The proposal was also to make the spec consistent.

ljnelson:
But see also: "highest priority".  Either interpretation works in English, which is of course the main problem: which interpretation is more primal?  Sadly, either, depending on usage.

"Ordinal", in its mathematical sense, is unambiguous: smallest number wins ("first" trumps "second").  Sadly, we've apparently redefined it in this specification.  Regardless of that, the real question is: is the _audience_ of MicroProfile Config users and MicroProfile Config-compatible configuration document authors aware of its unambiguous meaning?  I'm guessing no, i.e. no matter its definition it is sufficiently technical that "ordinary people" would have to look it up.  After all, MicroProfile Config itself has redefined it!  So maybe it's a poor choice.

[...]

PR #554, opened 2020-04-16
#418 config profile support

Emily-Jiang:
A proposal to support standard way to provide profile-specific properties that would be used only if a profile is specified. This for example enables an application to include configuration for different environments and development stages while only one of them is active (e.g. dev, test, prod).

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
@@ -116,6 +116,17 @@ public interface ConfigBuilder {
     <T> ConfigBuilder withConverter(Class<T> type, int priority, Converter<T> converter);
 
     /**
+     * Set the configuration profile.
+     * Once the configuration profile is set, the configuration property {@code %profile.config.name} overrides the property {@code config.name} 
+     * in the same configuration source. 
+     * The configuration value for the property {@code config.name} is retrieved from a configuration source with the highest ordinal. 
+     * The value is set by the property {@code %profile.config.name} if exists. Otherwise, the value is taken from the property {@code config.name}.
+     * @param profile the configuration profile name (must not be {@code null})
+     * @return this builder
+     */
+    ConfigBuilder withProfile(String profile);
+    
+    /**
      * Build a new {@link Config} instance based on this builder instance.
      *
      * @return the new configuration object
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
@@ -114,17 +114,6 @@ public interface ConfigBuilder {
      * @return this configuration builder instance
      */
     <T> ConfigBuilder withConverter(Class<T> type, int priority, Converter<T> converter);
-
-    /**
-     * Set the configuration profile.
-     * Once the configuration profile is set, the configuration property {@code %profile.config.name} overrides the property {@code config.name} 
-     * in the same configuration source. 
-     * The configuration value for the property {@code config.name} is retrieved from a configuration source with the highest ordinal. 
-     * The value is set by the property {@code %profile.config.name} if exists. Otherwise, the value is taken from the property {@code config.name}.
-     * @param profile the configuration profile name (must not be {@code null})
-     * @return this builder
-     */
-    ConfigBuilder withProfile(String profile);
     
     /**
      * Build a new {@link Config} instance based on this builder instance.
--- a/spec/src/main/asciidoc/configprofile.asciidoc
+++ b/spec/src/main/asciidoc/configprofile.asciidoc
@@ -69,10 +69,9 @@ The properties `%live.vehicle.name` and `%testing.vehicle.name` are inactive con
 
 If `mp.config.profile` is set to `live`, the property `%live.vehicle.name` is the active property. The `vehicleName` will be `train`. Similarly, `bike` will be the value of `vehicleName`, if the profile is `testing`.
 
-When retrieving the config values, this Config spec implementations must first search for the value of `mp.config.profile` (e.g. <profile name>) and then retrieve the value of the config property `%<profile name>.vehicle.name`,
-with the consideration of config source hierarchy specified via ordinal. It means if more than one config sources contain the same property `%<profile name>.vehicle.name`, 
-the value from the config source with the highest ordinal will be used. In the same config source, the property `%<profile name>.vehicle.name` overrides the config property `vehicle.name`. 
-If the property `%<profile name>.vehicle.name` does not exist, the value of the property `vehicle.name` will be retrieved.
+Conforming implementations are required to search for a configuration source with the highest ordinal (priority) that provides either the property name or the "profile-specific" property name. 
+If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned. If it doesn't contain the "profile-specific" name, the value of the plain property will be returned. 
+A "profile-specific" property name is defined as a property whose name consists of the following sequence: `% <profile name>.<original property name>`.
 
 The above algorithm applies to the following programmatic lookup as well.
 
--- a/spec/src/main/asciidoc/configprofile.asciidoc
+++ b/spec/src/main/asciidoc/configprofile.asciidoc
@@ -34,7 +34,7 @@ This is because the name of the profile is directly stored in the name of the co
 java -jar myapp.jar -Dmp.config.profile=testing
 ----
 
-The value of the property `mp.config.profile` shouldn't be updated after the application is started. It's only read once and will not be update once the `Config` object is constructed. If the property value of `mp.config.profile` is modified afterwards, the behavior is undefined and any changes to its value made later can be ignored by the implementation. 
+The value of the property `mp.config.profile` shouldn't be updated after the application is started. It's only read once and will not be updated once the `Config` object is constructed. If the property value of `mp.config.profile` is modified afterwards, the behavior is undefined and any changes to its value made later can be ignored by the implementation. 
 
 The value of the property `mp.config.profile` specifies a single active profile. Only one profile can be active at a time. Commas in the value have no special meaning. For example, if the value of `mp.config.profile` is `testing,live`,  a single profile named `testing,live` is active instead of two active profiles. 
 If the property `mp.config.profile` is specified in multiple config sources, the value of the property is determined following the same rules as other configuration properties, which means the value in the config source with the highest ordinal wins.
@@ -47,7 +47,7 @@ Conforming implementations are required to search for a configuration source wit
 If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned. If it doesn't contain the "profile-specific" name, the value of the plain property will be returned. 
 
 
-For an instance, a config source can be specified as follows.
+For instance, a config source can be specified as follows.
 
 [source, text]
 ----
--- a/spec/src/main/asciidoc/configprofile.asciidoc
+++ b/spec/src/main/asciidoc/configprofile.asciidoc
@@ -27,20 +27,25 @@ Config Profile indicates the project phase, such as dev, testing, live, etc.
 === Specify Config Profile
 
 The config profile can be specified via the property `mp.config.profile`, which can be set in any of the configuration sources. The value of the property can contain only characters that are valid for config property names. 
-This is because the name of the profile is directly stored in the name of the config property. It can be set when starting your microservice. e.g.
+This is because the name of the profile is directly stored in the name of the config property. It can be set when starting your application. e.g.
 
 [source, text]
 ----
 java -jar myapp.jar -Dmp.config.profile=testing
 ----
 
-The property `mp.config.profile` is only read once on microservice starting. Any changes after that is ignored. Furthermore, the value specifies a single profile. Only one profile can be active at a time.
-Commas in the value have no special meaning. For example, if the value of `mp.config.profile` is `testing,live`,  a single profile named `testing,live` is enabled instead of enabling two profiles. If the property `mp.config.profile` is specified in multiple config sources, 
-the value of the property is determined following the same rules as other configuration properties, which means the value in the config source with the highest ordinal wins.
+The value of the property `mp.config.profile` shouldn't be updated after the JVM is started. It's only read once during the JVM start up. If the property value is modified afterwards, the behavior is undefined and any changes to its value made later can be ignored by the implementation. 
+
+The value of the property `mp.config.profile` specifies a single active profile. Only one profile can be active at a time. Commas in the value have no special meaning. For example, if the value of `mp.config.profile` is `testing,live`,  a single profile named `testing,live` is active instead of two active profiles. 
+If the property `mp.config.profile` is specified in multiple config sources, the value of the property is determined following the same rules as other configuration properties, which means the value in the config source with the highest ordinal wins.
 
 === How Config Profile works
 
-In supporting the Config Profile, the configuration property that utlises the Config Profile needs to have its name starting with the naming convention of `%configProfile.`, (e.g. `%live.vehicle.name`).
+The configuration property that utilizes the Config Profile is called a "profile-specific" property. A "profile-specific" property name consists of the following sequence: `% <profile name>.<original property name>`.
+
+Conforming implementations are required to search for a configuration source with the highest ordinal (priority) that provides either the property name or the "profile-specific" property name. 
+If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned. If it doesn't contain the "profile-specific" name, the value of the plain property will be returned. 
+
 
 For an instance, a config source can be specified as follows.
 
@@ -69,11 +74,4 @@ In more details, if `mp.config.profile` is set to `dev`, the property `%dev.vehi
 The properties `%live.vehicle.name` and `%testing.vehicle.name` are inactive config properties and don't override the property `vehicle.name`.
 
 If `mp.config.profile` is set to `live`, the property `%live.vehicle.name` is the active property. The `vehicleName` will be `train`. Similarly, `bike` will be the value of `vehicleName`, if the profile is `testing`.
-
-Conforming implementations are required to search for a configuration source with the highest ordinal (priority) that provides either the property name or the "profile-specific" property name. 
-If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned. If it doesn't contain the "profile-specific" name, the value of the plain property will be returned. 
-A "profile-specific" property name is defined as a property whose name consists of the following sequence: `% <profile name>.<original property name>`.
-
-The above algorithm applies to the following programmatic lookup as well.
-
 
--- a/spec/src/main/asciidoc/configprofile.asciidoc
+++ b/spec/src/main/asciidoc/configprofile.asciidoc
@@ -1,0 +1,79 @@
+//
+// Copyright (c) 2020 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// Contributors:
+// Emily Jiang
+
+
+[[configprofile]]
+== Config Profile
+
+Config Profile indicates the project phase, such as dev, testing, live, etc. 
+
+=== Specify Config Profile
+
+The config profile can be specified via a system property or an environment viable `mp.config.profile`. It can be set when starting your microservice. The value of the property can contain only characters that are valid for config property names. 
+This is because the name of the profile is directly stored in the name of the config property. e.g.
+
+[source, text]
+----
+java -jar myapp.jar -Dmp.config.profile=testing
+----
+
+The property `mp.config.profile` is only read once on microservice starting. Any changes after that is ignored. Furthermore, the value specifies a single profile. Only one profile can be active at a time.
+Commas in the value have no special meaning. For example, if the value of `mp.config.profile` is `testing,live`,  a single profile named `testing,live` is enabled instead of enabling two profiles. 
+
+=== How Config Profile works
+
+In supporting the Config Profile, the config property names will need to start with the naming convention of `%configProfile.`, (e.g. `%live.vehicle.name`).
+
+For an instance, a config source can be specified as follows.
+
+[source, text]
+----
+%dev.vehicle.name=car
+%live.vehicle.name=train
+%testing.vehicle.name=bike
+vehicle.name=lorry
+----
+
+A config property associated with the Config Profile can be looked up as shown below.
+
+[source, text]
+----
+@Inject @ConfigProperty(name="vehicle.name") String vehicleName;
+----
+
+[source, text]
+----
+String vehicleName = ConfigProvider.getConfig().getValue("vehicle.name", String.class);
+----
+
+If the system property or environment variable `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` is the Active Property. An active property overrides the properties in the same config source. 
+In more details, if `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` overrides the property `vehicle.name`. The `vehicleName` will be set to `car`.
+The properties `%live.vehicle.name` and `%testing.vehicle.name` are inactive config properties and don't override the property `vehicle.name`.
+
+If `mp.config.profile` is set to `live`, the property `%live.vehicle.name` is the active property. The `vehicleName` will be `train`. Similarly, `bike` will be the value of `vehicleName`, if the profile is `testing`.
+
+When retrieving the config values, this Config spec implementations must first search for the value of `mp.config.profile` (e.g. <profile name>) and then retrieve the value of the config property `%<profile name>.vehicle.name`,
+with the consideration of config source hierarchy specified via ordinal. It means if more than one config sources contain the same property `%<profile name>.vehicle.name`, 
+the value from the config source with the highest ordinal will be used. In the same config source, the property `%<profile name>.vehicle.name` overrides the config property `vehicle.name`. 
+If the property `%<profile name>.vehicle.name` does not exist, the value of the property `vehicle.name` will be retrieved.
+
+The above algorithm applies to the following programmatic lookup as well.
+
+
--- a/spec/src/main/asciidoc/configprofile.asciidoc
+++ b/spec/src/main/asciidoc/configprofile.asciidoc
@@ -26,8 +26,8 @@ Config Profile indicates the project phase, such as dev, testing, live, etc.
 
 === Specify Config Profile
 
-The config profile can be specified via a system property or an environment viable `mp.config.profile`. It can be set when starting your microservice. The value of the property can contain only characters that are valid for config property names. 
-This is because the name of the profile is directly stored in the name of the config property. e.g.
+The config profile can be specified via the property `mp.config.profile`, which can be set in any of the configuration sources. The value of the property can contain only characters that are valid for config property names. 
+This is because the name of the profile is directly stored in the name of the config property. It can be set when starting your microservice. e.g.
 
 [source, text]
 ----
@@ -35,11 +35,12 @@ java -jar myapp.jar -Dmp.config.profile=testing
 ----
 
 The property `mp.config.profile` is only read once on microservice starting. Any changes after that is ignored. Furthermore, the value specifies a single profile. Only one profile can be active at a time.
-Commas in the value have no special meaning. For example, if the value of `mp.config.profile` is `testing,live`,  a single profile named `testing,live` is enabled instead of enabling two profiles. 
+Commas in the value have no special meaning. For example, if the value of `mp.config.profile` is `testing,live`,  a single profile named `testing,live` is enabled instead of enabling two profiles. If the property `mp.config.profile` is specified in multiple config sources, 
+the value of the property is determined following the same rules as other configuration properties, which means the value in the config source with the highest ordinal wins.
 
 === How Config Profile works
 
-In supporting the Config Profile, the config property names will need to start with the naming convention of `%configProfile.`, (e.g. `%live.vehicle.name`).
+In supporting the Config Profile, the configuration property that utlises the Config Profile needs to have its name starting with the naming convention of `%configProfile.`, (e.g. `%live.vehicle.name`).
 
 For an instance, a config source can be specified as follows.
 
@@ -63,7 +64,7 @@ A config property associated with the Config Profile can be looked up as shown b
 String vehicleName = ConfigProvider.getConfig().getValue("vehicle.name", String.class);
 ----
 
-If the system property or environment variable `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` is the Active Property. An active property overrides the properties in the same config source. 
+If the property `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` is the Active Property. An active property overrides the properties in the same config source. 
 In more details, if `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` overrides the property `vehicle.name`. The `vehicleName` will be set to `car`.
 The properties `%live.vehicle.name` and `%testing.vehicle.name` are inactive config properties and don't override the property `vehicle.name`.
 
--- a/spec/src/main/asciidoc/configprofile.asciidoc
+++ b/spec/src/main/asciidoc/configprofile.asciidoc
@@ -34,7 +34,7 @@ This is because the name of the profile is directly stored in the name of the co
 java -jar myapp.jar -Dmp.config.profile=testing
 ----
 
-The value of the property `mp.config.profile` shouldn't be updated after the application is started. It's only read once during the application start up. If the property value is modified afterwards, the behavior is undefined and any changes to its value made later can be ignored by the implementation. 
+The value of the property `mp.config.profile` shouldn't be updated after the application is started. It's only read once and will not be update once the `Config` object is constructed. If the property value of `mp.config.profile` is modified afterwards, the behavior is undefined and any changes to its value made later can be ignored by the implementation. 
 
 The value of the property `mp.config.profile` specifies a single active profile. Only one profile can be active at a time. Commas in the value have no special meaning. For example, if the value of `mp.config.profile` is `testing,live`,  a single profile named `testing,live` is active instead of two active profiles. 
 If the property `mp.config.profile` is specified in multiple config sources, the value of the property is determined following the same rules as other configuration properties, which means the value in the config source with the highest ordinal wins.
--- a/spec/src/main/asciidoc/configprofile.asciidoc
+++ b/spec/src/main/asciidoc/configprofile.asciidoc
@@ -34,7 +34,7 @@ This is because the name of the profile is directly stored in the name of the co
 java -jar myapp.jar -Dmp.config.profile=testing
 ----
 
-The value of the property `mp.config.profile` shouldn't be updated after the JVM is started. It's only read once during the JVM start up. If the property value is modified afterwards, the behavior is undefined and any changes to its value made later can be ignored by the implementation. 
+The value of the property `mp.config.profile` shouldn't be updated after the application is started. It's only read once during the application start up. If the property value is modified afterwards, the behavior is undefined and any changes to its value made later can be ignored by the implementation. 
 
 The value of the property `mp.config.profile` specifies a single active profile. Only one profile can be active at a time. Commas in the value have no special meaning. For example, if the value of `mp.config.profile` is `testing,live`,  a single profile named `testing,live` is active instead of two active profiles. 
 If the property `mp.config.profile` is specified in multiple config sources, the value of the property is determined following the same rules as other configuration properties, which means the value in the config source with the highest ordinal wins.
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/CustomConfigProfileConfigSource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/CustomConfigProfileConfigSource.java
@@ -44,6 +44,7 @@ public class CustomConfigProfileConfigSource implements ConfigSource {
     public int getOrdinal() {
         return 500;
     }
+
     @Override
     public String getValue(String key) {
         return configValues.get(key);
@@ -54,12 +55,8 @@ public class CustomConfigProfileConfigSource implements ConfigSource {
         return "customConfigProfileSource";
     }
 
-
     @Override
-    public Set<String> getPropertyNames() {
-        
+    public Set<String> getPropertyNames() {    
         return configValues.keySet();
     }
-
-
 }
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/CustomConfigProfileConfigSource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/CustomConfigProfileConfigSource.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package org.eclipse.microprofile.config.tck.configsources;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+/**
+ * @author Emily Jiang
+ */
+public class CustomConfigProfileConfigSource implements ConfigSource {
+
+    private Map<String, String> configValues = new HashMap<>();
+
+    public CustomConfigProfileConfigSource() {
+        configValues.put("mp.config.profile", "test");
+        configValues.put("%dev.vehicle.name", "bike");
+        configValues.put("%prod.vehicle.name", "bus");
+        configValues.put("%test.vehicle.name", "van");
+        configValues.put("vehicle.name", "car");
+    }
+
+    @Override
+    public int getOrdinal() {
+        return 500;
+    }
+    @Override
+    public String getValue(String key) {
+        return configValues.get(key);
+    }
+
+    @Override
+    public String getName() {
+        return "customConfigProfileSource";
+    }
+
+
+    @Override
+    public Set<String> getPropertyNames() {
+        
+        return configValues.keySet();
+    }
+
+
+}
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.eclipse.microprofile.config.tck.profile;
+
+
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+
+import javax.enterprise.context.Dependent;
+import javax.enterprise.inject.spi.CDI;
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.config.tck.base.AbstractTest;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.testng.Arquillian;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.testng.annotations.BeforeTest;
+import org.testng.annotations.Test;
+
+/**
+ * Test cases for Config profile
+ * 
+ * @author Emily Jiang
+ */
+public class DevConfigProfileTest extends Arquillian {
+
+    
+    @Deployment
+    public static Archive deployment() {
+        JavaArchive testJar = ShrinkWrap
+                .create(JavaArchive.class, "DevConfigProfileTest.jar")
+                .addClasses(DevConfigProfileTest.class, ProfilePropertyBean.class)
+                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+                .as(JavaArchive.class);
+
+        AbstractTest.addFile(testJar, "META-INF/microprofile-config.properties");
+
+        WebArchive war = ShrinkWrap
+                .create(WebArchive.class, "DevConfigProfileTest.war")
+                .addAsLibrary(testJar);
+        return war;
+       
+    }
+
+    @BeforeTest
+    public void setUpTest() {
+        clearAllPropertyValues();
+        ensureAllPropertyValuesAreDefined();
+    }
+
+    @Test
+    public void testConfigProfileWithDev() {
+        ProfilePropertyBean bean = CDI.current().select(ProfilePropertyBean.class).get();
+
+        assertThat(bean.getConfigProperty(), is(equalTo("bike")));
+        assertThat(ConfigProvider.getConfig().getValue("vehicle.name", String.class), is(equalTo("bike")));
+    }
+
+    private void ensureAllPropertyValuesAreDefined() {
+        System.setProperty("mp.config.profile", "dev");
+        
+    }
+
+    private void clearAllPropertyValues() {
+        System.getProperties().remove("mp.config.profile");
+        
+    }
+
+    @Dependent
+    public static class ProfilePropertyBean {
+        @Inject
+        @ConfigProperty(name="vehicle.name")
+        private String vehicleName;
+
+        public String getConfigProperty() {
+            return vehicleName;
+        }
+    }
+    
+    
+}
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
@@ -32,15 +32,14 @@ import javax.inject.Inject;
 
 import org.eclipse.microprofile.config.ConfigProvider;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
-import org.eclipse.microprofile.config.tck.base.AbstractTest;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.testng.Arquillian;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
-import org.testng.annotations.BeforeTest;
 import org.testng.annotations.Test;
 
 /**
@@ -56,10 +55,18 @@ public class DevConfigProfileTest extends Arquillian {
         JavaArchive testJar = ShrinkWrap
                 .create(JavaArchive.class, "DevConfigProfileTest.jar")
                 .addClasses(DevConfigProfileTest.class, ProfilePropertyBean.class)
+                .addAsManifestResource(
+                    new StringAsset(
+                        "mp.config.profile=dev\n" +
+                        "%dev.vehicle.name=bike\n" +
+                        "%prod.vehicle.name=bus\n" +
+                        "%test.vehicle.name=van\n" +
+                        "vehicle.name=car"),
+                        "microprofile-config.properties")
                 .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
                 .as(JavaArchive.class);
 
-        AbstractTest.addFile(testJar, "META-INF/microprofile-config.properties");
+                
 
         WebArchive war = ShrinkWrap
                 .create(WebArchive.class, "DevConfigProfileTest.war")
@@ -68,11 +75,6 @@ public class DevConfigProfileTest extends Arquillian {
        
     }
 
-    @BeforeTest
-    public void setUpTest() {
-        clearAllPropertyValues();
-        ensureAllPropertyValuesAreDefined();
-    }
 
     @Test
     public void testConfigProfileWithDev() {
@@ -82,15 +84,6 @@ public class DevConfigProfileTest extends Arquillian {
         assertThat(ConfigProvider.getConfig().getValue("vehicle.name", String.class), is(equalTo("bike")));
     }
 
-    private void ensureAllPropertyValuesAreDefined() {
-        System.setProperty("mp.config.profile", "dev");
-        
-    }
-
-    private void clearAllPropertyValues() {
-        System.getProperties().remove("mp.config.profile");
-        
-    }
 
     @Dependent
     public static class ProfilePropertyBean {
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
@@ -48,8 +48,7 @@ import org.testng.annotations.Test;
  * @author Emily Jiang
  */
 public class DevConfigProfileTest extends Arquillian {
-
-    
+   
     @Deployment
     public static Archive deployment() {
         JavaArchive testJar = ShrinkWrap
@@ -64,9 +63,7 @@ public class DevConfigProfileTest extends Arquillian {
                         "vehicle.name=car"),
                         "microprofile-config.properties")
                 .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
-                .as(JavaArchive.class);
-
-                
+                .as(JavaArchive.class);   
 
         WebArchive war = ShrinkWrap
                 .create(WebArchive.class, "DevConfigProfileTest.war")
@@ -75,26 +72,20 @@ public class DevConfigProfileTest extends Arquillian {
        
     }
 
-
     @Test
     public void testConfigProfileWithDev() {
         ProfilePropertyBean bean = CDI.current().select(ProfilePropertyBean.class).get();
-
         assertThat(bean.getConfigProperty(), is(equalTo("bike")));
         assertThat(ConfigProvider.getConfig().getValue("vehicle.name", String.class), is(equalTo("bike")));
     }
-
 
     @Dependent
     public static class ProfilePropertyBean {
         @Inject
         @ConfigProperty(name="vehicle.name")
         private String vehicleName;
-
         public String getConfigProperty() {
             return vehicleName;
         }
-    }
-    
-    
+    }    
 }
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/InvalidConfigProfileTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/InvalidConfigProfileTest.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.eclipse.microprofile.config.tck.profile;
+
+
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+
+import javax.enterprise.context.Dependent;
+import javax.enterprise.inject.spi.CDI;
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.config.tck.base.AbstractTest;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.testng.Arquillian;
+import org.jboss.shrinkwrap.api.Archive;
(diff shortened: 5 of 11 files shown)


spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+// Emily Jiang
+
+
+[[configprofile]]
+== Config Profile
+
+Config Profile indicates the project phrase, such as dev, testing, live, etc. 
+=== Specify Config Profile
+
+The config profile can be specified via a system property `mp.config.profile`. It can be set when starting your microservice, e.g.
```
radcortez, line 28: I think we can add that only one valid profile is active at a time.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+----
+
+In the above config source, if the system property `mp.config.profile` is set to `dev`. The `vechileName` in the following snippet will be resolved to be `car`.
+Similarly, it will be `train` if the `mp.config.profile` is set to `live`, `bike` if the profile is `testing`.
+
+[source, text]
+----
+@Inject @ConfigProperty(name="vehicle.name") String vechileName;
+----
+
```
dmlloyd, line 54: We need some normative text covering what conforming implementations are required to do.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+The config profile can be specified via a system property `mp.config.profile`. It can be set when starting your microservice, e.g.
+[source, text]
+----
+java -jar myapp.jar -Dmp.config.profile=testing
+----
+
+=== How Config Profile works
+
+In supporting the Config Profile, the config property names will need to start with the naming convention of `%configProfile%`, (e.g. `%live%vechile.name`).
```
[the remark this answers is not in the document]
Emily-Jiang: I thought about this. If we use %%, we can clearly demonstrate the profile. In your scenario, when you do the lookup, you will use `@ConfigProperty(name="config.key")`. Basically Quarkus use % and . to separate the profile name. I think starting and ending with % is better.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+For an instance, a config source can be specified as follows.
+
+[source, text]
+----
+%dev%vehicle.name=car
+%live%vehicle.name=train
+%testing%vehicle.name=bike
+----
+
+In the above config source, if the system property `mp.config.profile` is set to `dev`. The `vechileName` in the following snippet will be resolved to be `car`.
```
radcortez, line 47: I think we need to clarify what happens if you activate a profile and search for a key that doesn't have a profile value. Expected behaviour will be to fallback to the key without profile.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+[source, text]
+----
+java -jar myapp.jar -Dmp.config.profile=testing
+----
+
+=== How Config Profile works
+
+In supporting the Config Profile, the config property names will need to start with the naming convention of `%configProfile%`, (e.g. `%live%vechile.name`).
+
+For an instance, a config source can be specified as follows.
```
radcortez, line 38: Would it be possible to override profiles even if they are active?

spec/src/main/asciidoc/configprofile.asciidoc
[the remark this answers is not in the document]
dmlloyd: No, we don't use `%` as a separator.  We prepend it on to the profile name and use that as a config segment.  This means that we only have one separator which is `.`.  Anyway I think MP ought to standardize working approaches rather than invent new ones.  Is anything else using this double `%` strategy?

gastaldi, line 36: There is a typo in the document: `vechile` != `vehicle`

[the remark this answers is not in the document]
Emily-Jiang: This is the first occurrence of this kind of naming convention as far as I know. Personally I feel %xxx% is standing out :o.

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: Do you mean during the running you could do `System.setProperty(mp.config.profile, "testing")`? I think this profile should only be looked up on microservice starting to avoid headaches.

Emily-Jiang: If you mean another config source with higher ordinal also contains the same config property `%live%vechile.name`, the value of `%live%vechile.name` will be taken from a configsource having a higher ordinal.

Emily-Jiang:
I am working on TCKs. Please continue to comment on the behaviour.

spec/src/main/asciidoc/configprofile.asciidoc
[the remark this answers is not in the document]
dmlloyd: Remember that our goal isn't to stand out. AFAIK our goal is to standardize existing best practices, and all existing implementations of this functionality use `%profilename.` as a prefix.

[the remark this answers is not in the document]
gastaldi: I stand with @dmlloyd in this one. It's better to not deviate from working solutions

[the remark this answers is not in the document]
radcortez: Maybe I'm already biased, but %[profile].[key] seems more readable than %[profile]%key

spec/src/main/asciidoc/configprofile.asciidoc
radcortez: In the case where you have a higher ordinal with just `vehicle.name`. Will the profile always take precedence no matter the ordinal?

dmlloyd: That's a trick question! :wink:

No existing implementation works that way.  Profile keys should only take precedence over configuration keys from the same or lesser configuration source otherwise users can have some truly surprising behavior.

radcortez: Yes, that is why I'm asking, so we can clarify it in the spec.

spec/src/main/asciidoc/configprofile.asciidoc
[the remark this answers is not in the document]
Emily-Jiang: By the way, the spec has no obligation to adopt the existing implementation. It can come up with different alternatives or completely different designs to address the same problems. This is the practices we have been using. For this matter, since you three are from the same project, I think it is good idea to hear what others think, especially users.

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: > In the case where you have a higher ordinal with just vehicle.name. Will the profile always take precedence no matter the ordinal?
The profile should always take precedence no matter the ordinal because of the config name rules. If multiple config sources contain the same profile config (e.g. `%live%vehicle.name`), the config value from the highest ordinal wins.

dmlloyd: That's the opposite of how SmallRye and Quarkus work and it means that overriding config properties with system properties won't work if the original property is in a profile. Why would you do it this way? Do you know of some existing implementation that does things this way?

Emily-Jiang: Let's take an example of defining `%live%vehicle.name` in microprofile-config.properties and then in Kubernetes deployment file, Todd, a devops engineer, defines an environment variable '_live_vehicle_name` to override the value, I would expect the value from environment variable wins. By the way, as mentioned earlier, what we defined here should not be limited by the existing implementation.

OndroMih: I agree with @Emily-Jiang that it's possible to overwrite properties in a profile with other properties in the same profile, no issue there. But the question is which of the approaches is better.

It's probably easier to maintain configuration if properties in a profile only override properties in the same config source. Otherwise it can be a huge mess to make sure that profiles are clearly defined.

On the other hand, there's a perfectly good usecase for overriding properties defined in all config sources. For example, if I have a custom config source that only defines properties in profiles. E.g., the default config is in a database and profile-specifc configs are in a file (or more files, e.g. one per profile).

I prefer to keep things simple and allow that profile properties override base properties only in the same config source. Or in all config sources with the same ordinal, although that might be awfully confusing.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+// Emily Jiang
+
+
+[[configprofile]]
+== Config Profile
+
+Config Profile indicates the project phase, such as dev, testing, live, etc. 
+=== Specify Config Profile
+
+The config profile can be specified via a system property `mp.config.profile`. It can be set when starting your microservice, e.g.
```
OndroMih, line 28: We should also allow builders to set the profile for the Config they are building. Using a system property is nice on a global level but it doesn't make sense if an application uses a custom config or even more custom configs. 

A new method like this in `ConfigBuilder` would allow it:

```
ConfigBuilder withProfile(String profileName)
```

spec/src/main/asciidoc/configprofile.asciidoc
OndroMih: Overriding properties only in the same config source is also more efficient in the most cases. Here's some comparison of the number of requests to config sources for various cases (N is the number of config sources):

| Case        | across all config sources           | only in the same config source  |
| ------------- |:-------------:| -----:|
| config source with **highest** ordinal contains **profile** property      | 1 request | 1 request |
| source with **highest** ordinal contains **normal** property, only source with the **lowest** ordinal contains **profile** property       | 2xN      |   2 |
| source with the **highest** ordinal contains **normal** property, no source contains profile property | 2xN      |    2 |
| source with the **lowest** ordinal contains **profile** property, no other source contains profile or normal property | N      |    2xN |
| **no source** contains profile or normal property | 2xN      |    2xN | 
| only source with the **median** ordinal contains **profile** property, no other source contains profile or normal property  | N/2      |    N | 
| only source with the **median** ordinal contains **normal** property, no other source contains profile or normal property  | N + N/2      |    N | 

Mediam ordinal means that if you order sources from highest ordinal to lowest, the one in the middle has median ordinal.

spec/src/main/asciidoc/configprofile.asciidoc
dmlloyd: It ought to be possible to set the config profile via a configuration property IMO.

spec/src/main/asciidoc/configprofile.asciidoc
smillidge: Detailed description of resolution rules and interplay between ordinals and other config sources are needed here and these rules also tested within the TCK.

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: @OndroMih Thanks for the explaination and I understood what you meant. I think we are in agreement
profile properties override base properties only in the same config source. 
yields 
profile properties with a higher ordinal override base or other profile properties

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: +1 @smillidge , I will add more TCKs for the config source ordinal rules once we are happy with the rules. See here for more explaination.

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: I'll update the spec wording to state "profile properties override base properties only in the same config source."

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: Since when reading config sources, we need to figure out the config profile value. Using system properties is much simplier. What will happen if set different value for `mp.config.profile` in different config sources, which makes the situation very complicated.

dmlloyd: Highest config source wins, as usual.  We do support self-configuration in other areas, so I don't think this is so unusual.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+[source, text]
+----
+java -jar myapp.jar -Dmp.config.profile=testing
+----
+The property is only read once on microservice starting. Any changes after that is ignored. Furthermore, the value is a singular. If it contains a comma, such as `testing, live`,
```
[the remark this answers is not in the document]
Emily-Jiang: updated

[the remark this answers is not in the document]
Emily-Jiang: reworded.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+Config Profile indicates the project phase, such as dev, testing, live, etc. 
+=== Specify Config Profile
+
+The config profile can be specified via a system property `mp.config.profile`. It can be set when starting your microservice, e.g.
+[source, text]
+----
+java -jar myapp.jar -Dmp.config.profile=testing
+----
+The property is only read once on microservice starting. Any changes after that is ignored. Furthermore, the value is a singular. If it contains a comma, such as `testing, live`, 
+if value of `mp.config.profile` is `testing, live` as a single profile instead of two profiles.
```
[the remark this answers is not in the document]
Emily-Jiang: updated.

spec/src/main/asciidoc/configprofile.asciidoc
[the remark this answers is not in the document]
Emily-Jiang: updated

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
```diff
@@ -115,6 +115,16 @@
...
     <T> ConfigBuilder withConverter(Class<T> type, int priority, Converter<T> converter);
 
+    /**
+     * Set the config profile.
+     * Once the config profile is set, the config property {@code %profile.config.name} overrides the property {@code config.name} 
+     * in the same config source. 
+     * The config value for the property {@code config.name} is retrieved from a config source with the highest ordinal. 
+     * The value is set by the property {@code %profile.config.name} if exists. Otherwise, the value is taken from the property {@code config.name}.
+     * @param profile the config profile
+     * @return
```
dmlloyd, line 125:
```suggestion
     * Set the configuration profile.
     * Once the configuration profile is set, the configuration property {@code %profile.config.name} overrides the property {@code config.name} 
     * in the same configuration source. 
     * The configuration value for the property {@code config.name} is retrieved from a configuration source with the highest ordinal.
     * The value is set by the property {@code %profile.config.name} if exists. Otherwise, the value is taken from the property {@code config.name}.
     * @param profile the configuration profile name (must not be {@code null})
     * @return this builder
```

Also the resolution of #553 would affect the wording here.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
```diff
@@ -115,6 +115,16 @@
...
+     */
+    ConfigBuilder withProfile(String profile);
     /**
```
dmlloyd, line 128: A small detail, but there should be a blank line in between the method names for consistency with the rest of the file.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,80 @@
...
+----
+
+If the system property `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` is the Active Property. An active property overrides the properties in the same config source. 
+In more details, if the system property `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` overrides the property `vehicle.name`. 
+The properties `%live.vehicle.name` and `%testing.vehicle.name` are inactive config property and don't override the property `vehicle.name`.
+
+In more details, on setting the system property `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` is the active property. The `vehicleName` is `train`.
+On setting the system property `mp.config.profile` is set to `live`, the property `%live.vehicle.name` is the active property. The `vehicleName` is `train`. Similarly, `bike` will be the value of `vehicleName`, if the profile is `testing`.
+
+When retrieving the config values, this Config spec implementations must first search for the value of `mp.config.profile` and then retrieve the value of the config property `%${mp.config.profile}.vehicle.name`,
```
dmlloyd, line 73: I don't think that using the proposed property expansion syntax here is a good idea, since that proposal has not yet been accepted; I think it would just be confusing.  Also, I don't think this counts as "normative" since it's really an example.

I would add some wording like:

> Conforming implementations are required to return the value of the "profile-specific" property name whenever the "profile-specific" property name occurs within a configuration source whose overall priority¹ is equal to, or higher than, the value corresponding to the original property name (if any).  A "profile-specific" property name is defined as a property whose name consists of the following sequence: `%` \<profile name> `.` \<original property name>

Add the footnote for "overall priority" because configuration source ordering is more complex than just ordinal value.  That definition should be specified & linked from here.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,80 @@
...
+with the consideration of config source hierarchy specified via ordinal. It means if more than one config source contains the same property `%${mp.config.profile}.vehicle.name`, 
+the value from the config source with the highest ordinal will be used. In the same config source, the property `%${mp.config.profile}.vehicle.name` overrides the config property `vehicle.name`. 
+If no value is found, the value of the config name `vehicle.name` will be retrieved.
```
OndroMih, line 76: I think all this is too complicated. I suggest to describe the algorithm as follows:

> Conforming implementations are required to search for a configuration source with the highest ordinal (priority) that provides either the property name or the "profile-specific" property name. If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned. If it doesn't contain the "profile-specific" name, the value of the plain property will be returned. A "profile-specific" property name is defined as a property whose name consists of the following sequence: `% <profile name> . <original property name>`

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,80 @@
...
+----
+java -jar myapp.jar -Dmp.config.profile=testing
+----
+
+The property `mp.config.profile` is only read once on microservice starting. Any changes after that is ignored. Furthermore, the value specifies a single profile. Only one profile can be active at a time.
+Commas in the value have no special meaning. For example, if the value of `mp.config.profile` is `testing,live`,  a single profile named `testing,live` is enabled instead of enabling two profiles. 
+
+=== How Config Profile works
+
+In supporting the Config Profile, the config property names will need to start with the naming convention of `%configProfile.`, (e.g. `%live.vehicle.name`).
```
OndroMih, line 42: Do you realize that `%` is not allowed in some config sources, e.g. in environment variable names? For environment variables, the "profile" property would need to be specified using `_` instead of `%`, such as: `_live_vehicle_name`. I'm not sure if it would cause any problems but we should think about it.

dmlloyd: It hasn't been a problem for us; generally users have not seemed to have needed to use env vars (or system properties for that matter) to override profile-specific properties.

smillidge: However if it is to be usable in all Config Sources then perhaps we should think of a different identifier.

Emily-Jiang: If it is specified as environment variables, it will be transformed from % to _. It is same as other config properties.

spec/src/main/asciidoc/configprofile.asciidoc
smillidge: So does the search stop if the highest ordinal config source returns a non-profile specific value for the property?

OndroMih: Yes, that's the idea. We discussed this in this thread: https://github.com/eclipse/microprofile-config/pull/554/files#r409673972. And agreed that "profile properties override base properties only in the same config source.". So the search can stop with the highest ordinal config source that provides the property, even if it's a non-profile specific property.

This is more efficient to implement. It's also less confusing for users because it's easier to find a profile-specific property that overrides the default value because it has to be in the same config source.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
```diff
@@ -0,0 +1,107 @@
...
+                .addAsLibrary(testJar);
+        return war;
+       
+    }
+
+    @BeforeTest
+    public void setUpTest() {
+        clearAllPropertyValues();
+        ensureAllPropertyValuesAreDefined();
+    }
```
radcortez, line 75: We may want to figure out a different way to pass in the profile name to test.

The main issue is that `@BeforeTest` runs before any test being executed and not for this class only. It means that every `@BeforeTest` will execute and provide a non deterministic value for the config. `@BeforeClass` doesn't work either, because it runs after the Arquillian deployment is done (and the Config initialised). 

We can leave it this way, but will have to set up separate `test` with the specific classes of profiles or maybe we can just put the profile name in `microprofile-config.properties` and change it on the fly with the deployment.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
Emily-Jiang: updated.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
Emily-Jiang: ok. fixed.

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: I removed the property expansion syntax but reworded slightly. Please double check to see whether it is understandable.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,80 @@
...
+----
+
+[source, text]
+----
+String vehicleName = ConfigProvider.getConfig().getValue("vehicle.name", String.class);
+----
+
+If the system property `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` is the Active Property. An active property overrides the properties in the same config source. 
+In more details, if the system property `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` overrides the property `vehicle.name`. 
+The properties `%live.vehicle.name` and `%testing.vehicle.name` are inactive config property and don't override the property `vehicle.name`.
```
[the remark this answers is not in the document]
Emily-Jiang: updated.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,80 @@
...
+
+In more details, on setting the system property `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` is the active property. The `vehicleName` is `train`.
```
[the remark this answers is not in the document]
Emily-Jiang: updated.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,80 @@
...
+On setting the system property `mp.config.profile` is set to `live`, the property `%live.vehicle.name` is the active property. The `vehicleName` is `train`. Similarly, `bike` will be the value of `vehicleName`, if the profile is `testing`.
```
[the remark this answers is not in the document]
Emily-Jiang: updated

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
Emily-Jiang: The profile name can be set via system property or environment variable. We can't set it in microprofile-config.properties unfortunately.

dmlloyd: Why not?

Emily-Jiang: It is because in the spec we said the profile name is set via system property or environment variable. How can we directly set it in a configuration file in the tck?

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: I rephrased as @OndroMih suggested.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
radcortez: Well a profile should be like any other config property, so I think we can set it in any ConfigSource.

Anyway, as we have now these tests will not pass because they don't get their own value when the test starts (either is set before for all tests or it is too late when the Config is already created).

If we want to keep it this way, then we need to provide a suite that runs these tests independently.

Emily-Jiang: @radcortez profile needs to be read before reading any other config sources. If we allow specifying this property in any config sources, we need to gather all config sources and work out the ordinal and then figure out the `profile` value.

We have used this approach for other tests. See [here](https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPlainInjectionTest.java#L71).

spec/src/main/asciidoc/configprofile.asciidoc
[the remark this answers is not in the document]
Emily-Jiang: updated

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
radcortez: I've tested it with SmallRye and it doesn't work. All methods annotated with `@BeforeTest` execute before any actual class execute. So for instance if you are running the entire suite, you may have values set from other test classes, unless you run each class separately.

Emily-Jiang: Thanks @radcortez for bearing with me! I see what you meant now. One possible solution is to create multiple test suite for these profile test. The other suggestion is to set only one value for `mp.config.profile`, which is not ideal.

Emily-Jiang: I'll define different test suites, which should fix this issue.

Emily-Jiang: hmmm. It is not going well as it needs to launch a different jvm. Here is what I think we can do:

- enable specifying mp.config.profile in microprofile-config.properties
or 

- continue with the current set but test to see which profile is enabled and configure the tests accordingly.
e.g. testConfigProfileWithDev will have a precondition check for mp.config.profile. If it is not set to dev, the tests will be skipped.
Thoughts?

radcortez: It didn't work with different suites? Strange, I think it did work for me.

Anyway, most likely the best approach is to just allow to configure `mp.config.profile` in `microprofile-config.properties` and then use different files in the test.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,78 @@
...
+
+The config profile can be specified via a system property or an environment viable `mp.config.profile`. It can be set when starting your microservice. The value of the property can contain only characters that are valid for config property names. 
+This is because the name of the profile is directly stored in the name of the config property. e.g.
+
+[source, text]
+----
+java -jar myapp.jar -Dmp.config.profile=testing
+----
+
+The property `mp.config.profile` is only read once on microservice starting. Any changes after that is ignored. Furthermore, the value specifies a single profile. Only one profile can be active at a time.
```
OndroMih, line 37:
```suggestion
The property `mp.config.profile` shouldn't change after the JVM is started. It's only read once during JVM start up. If the property is modified afterwards the behavior is undefined and any changes to its value made later can be ignored by the implementation. Furthermore, the value specifies a single profile. Only one profile can be active at a time.
```

We shouldn't refer to "microservices" which is a vague term and not all MP implementations are used to build microservices. Rather I'd refer to a running JVM instead of a microservice.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
OndroMih: Wait, do you mean that we need to amend the spec to be able to test it? For me this is a no go. This is a significant change as the spec now says that the profile property is only read once during startup. If implementations need to support reading the profile property from a property file in the deployed application, they need to reconfigure config each time an application is deployed. Some implementations provide internal config sources and they would need to reconfigure them too. And I'm not even talking about implementations that support multiple deployments.

If we can't find any better solution, we could create different profiles for tests and require implementers to run the TCK for each profile separately. Some implementations would need to be restarted with a different system property before each TCK.

Alternatively, we can use test suites and provide a hook for implementers to reconfigure/restart their implementation before each suite is run.

spec/src/main/asciidoc/configprofile.asciidoc
OndroMih: This is resolved now that we've added this section:

> Conforming implementations are required to search for a configuration source with the highest ordinal (priority) that provides either the property name or the "profile-specific" property name. 
> If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned. If it doesn't contain the "profile-specific" name, the value of the plain property will be returned. 
> A "profile-specific" property name is defined as a property whose name consists of the following sequence: `% <profile name>.<original property name>`.

spec/src/main/asciidoc/configprofile.asciidoc
OndroMih: Since environment variables would map `%` to  `_`, can't we just use `_` as the prefix also for system properties instead of introducing a new special character `%`? Implementations can support alternative notations like using `%`, no big deal. But a standard notation `_configProfile.property` would be more consistent with the current MP Config specification and wouldn't introduce a new special character.

spec/src/main/asciidoc/configprofile.asciidoc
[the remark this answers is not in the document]
OndroMih: This discussion follows in the new comment thread: https://github.com/eclipse/microprofile-config/pull/554/files#r419090888

spec/src/main/asciidoc/configprofile.asciidoc
OndroMih: For the record, the profile property names were discussed here: https://github.com/eclipse/microprofile-config/pull/554#discussion_r409668376 and this thread continues that discussion.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
OndroMih: By the way, a requirement to read the profile from `microprofile-config.properties` again introduces recursion because this config source would need to be initialized before all the other properties can be read. Some implementations (including Payara) initialize some config sources before an application is fully deployed and before config sources in the application are initialized. So I'm really against reading the profile value from any config source.

radcortez: No, there is no limitation to test this. Having different test suits work.

On the other hand, I don't see a reason for the profile configuration to be treated differently to any other configuration property.

Do you read any config before adding the application config sources?

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigBuilder.java
Emily-Jiang: When I tried to add tests for this spi, I suggest against the idea adding this method, because it will create more confusions. Setting `mp.config.profile` is better set via the config sources to avoid unnecessary conflicts of setting profile via a method or via a property.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
OndroMih: > Do you read any config before adding the application config sources?

It really depends. The current wording doesn't specify when the profile value should be read. It only says it's read once and then changes to the profile value are ignored. If implementations read the profile value during app startup but before the application config source is initialized then it's a problem. The spec would need to mandate that the profile value is read only after the microprofile-config.properties source is initialized. If we allow it for all sources, then it needs to be read after all config sources are initialized. 

There are many potential problems for some types of implementations:

* implementations that allow deployment of an application after startup can't guarantee that the profile value is read after the application is deployed
* some implementations allow multiple applications to be deployed. If they both define a different profile, which should be used?
* some implementations themselves are configured by MP Config and read config values during startup, before the app is started and potentially before the microprofile-config.properties source in the app is initialized. So they would ignore the profile defined by the application because it would be too late

For example, Payara allows using MP config values in the server configuration to externalize internal configuration. This configuration is read very early during bootstrap and it's almost impossible to delay it after an app is deployed.

So if the test suite can be adapted to work with a profile defined by a system property or env var, I wouldn't really introduce support for profile definition in MP config sources as it introduces a lot of complexity for very little gain. Of course, implementations can experiment and support it or can allow specifying the profile by other means on top of the specification.

dmlloyd: > So if the test suite can be adapted to work with a profile defined by a system property or env var, I wouldn't really introduce support for profile definition in MP config sources as it introduces a lot of complexity for very little gain. Of course, implementations can experiment and support it or can allow specifying the profile by other means on top of the specification.

In our implementation, it is simpler to get the profile from MP config sources than it is to read it from the env and sys props separately.  MP Config is already a framework that abstracts the env and sys properties: if it can't be reused for this purpose then there's a design problem.

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: updated.

Emily-Jiang:
@OndroMih we talked about setting `mp.config.profile` in last week's hangout. The TCKs can be fixed using multiple test suites. However, the more we talked about, it makes more sense to make the property `mp.config.profile` to be specified at normal config sources. It is a static property. We define this from a use case pespective. As for implementations, it is not complicated as you can build config as normal. However when you look up a particular config property, you can use interceptor to resolve the value, which was demonstrated by @radcortez .

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
Emily-Jiang: @OndroMih the property `mp.config.profile` is no different from other except it is static, where the others might be dynamic. I think the design makes MP Config much simplier and more usable. I think you can build your config as before. The easier implementation for this is to transform the config value as the config object contains all of the raw values. I think you cannot filter the properties at the config source level.

radcortez: @OndroMih and when do you actually build the Config instance? I believe that the right thing to do might be to specify that the configuration is read when the Config instance is being created.

Now, it does look weird to me that we are working on a Config Spec and we are introducing configuration keys that don't go through our own API :)

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: I had  a further thought on this. I am going to change to application starting instead of JVM startup as it is more accurate to say application starting due to some application server might have multiple applications installed on the same server.

radcortez: How about on Config creation?

Emily-Jiang: > How about on Config creation?

This will be from implementation point of view though, which is different from user perspective. Some implementations might create the config a lot earlier.

dmlloyd: We can all agree that the profile is read one time.  But it doesn't make sense to specify exactly when it must be read; all we can really say for sure is that the implementation will (of a necessity) read the profile property before the application reads or injects configuration properties.  Since that's all the spec can reasonably require, that's exactly what should be specified.  Obviously this doesn't happen during JVM startup; it might not happen during application startup either.

OndroMih:
We talked about supporting `mp.config.profile` property in any config source with @Emily-Jiang at the hangout on 21st May (after last @Emily-Jiang's comment on this PR) and we came to an agreement documented in [the hangout minutes](https://docs.google.com/document/d/1X6Q9K28VNHhVRkVBFlbNa-ZMusxjCfmoA7--uJFHWx0/edit#):

> #418 - Config Profile - Ondro mentioned that Payara won’t be able to cope with the mp.config.profile specified on ordinary configsources. Emily suggests to go back to env or system property to start with and then add more support.

The reasoning is that we don't need it to implement tests and we don't yet have a usecase that would justify standardizing this. 

I suggest that we initially approve that `mp.config.profile` is only specified by as an ENV var or a system property. We also won't support specifying the profile in the config builder as there's also no usecase for it. If somebody finds a good usecase for specifying the property using any config source or in the builder, we should discuss it separately and add it with another PR any time later.

radcortez:
> I suggest that we initially approve that `mp.config.profile` is only specified by as an ENV var or a system property. 

I'm sorry, but I strongly disagree with this. This is not about requiring the property for testing or anything else, is about us being consistent with the spec we are proposing.

We have a spec that is all about Configuration and we try to convince our users to use it and have their own ConfigSources and manage their configuration. When we need to set our own configuration we are not even able to follow our own spec.

How can we possibly justify this to an end user that we are not able to support this?

smillidge:
Well then you are going to have to properly specify what occurs in a runtime that supports multiple deployments. What happens if the system properties config source has one value for profile. Application A has another and Application B another. Which profile value takes precedence? I assume that it is the ConfigSource with the higher Ordinal value i.e. System Properties?

radcortez:
We follow the exact same rules and we have specified for loading configurations.

Forget about profile for a second. In that scenario, if I have `foo.bar` specified, which one gets loaded?

smillidge:
So the value can change from one application deployment to another if it is only specified within the application . Therefore the TCK has to honour this fact (not saying it doesn't). 

If you specify the property as a standard MP Config property then it is therefore potentially mutable. I can't see why that is ruled out in your draft text. It would be possible to reread the profile on every property lookup albeit at a performance cost.

Emily-Jiang:
> So the value can change from one application deployment to another if it is only specified within the application . Therefore the TCK has to honour this fact (not saying it doesn't).

If multiple deployments are on the same runtime, they are potentially having a different profile stage. 

> If you specify the property as a standard MP Config property then it is therefore potentially mutable. I can't see why that is ruled out in your draft text. It would be possible to reread the profile on every property lookup albeit at a performance cost.

Though the value is potential mutuable, the value when reading at the applications starting should be used as I don't see the profile changes without application restarts.

Emily-Jiang:
Even though @OndroMih and I discussed on the hangout, we agreed that maybe supporting in environment variable or system properties to start with. However, I do agree it would be a lot of cleaner if we treat it similar to other config properties but recommend it being static. Runtime can choose whether they want to support dynamic or not. In order to move this forward, I will do a poll here for a decision. We need to draw an conclusion asap.

Emily-Jiang:
Poll:
Only specify `mp.config.profile` in environment variables or Config properties.

Emily-Jiang:
Poll: vote
Specify `mp.config.properties` in any config sources, similar to any other configuration properties.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,77 @@
...
+----
+@Inject @ConfigProperty(name="vehicle.name") String vehicleName;
+----
+
+[source, text]
+----
+String vehicleName = ConfigProvider.getConfig().getValue("vehicle.name", String.class);
+----
+
+If the property `mp.config.profile` is set to `dev`, the property `%dev.vehicle.name` is the Active Property. An active property overrides the properties in the same config source.
```
struberg, line 72: This is exactly the other way around as any other established solutions do it. Everyone else is using postfixing, e.g. vehicle.name.dev. The reason is the following: 
* you can easily group configuration that way
* you can have multiple lookup paths e.g. a dev.oracle and dev.postgres postfix.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,77 @@
...
+// Contributors:
+// Emily Jiang
+
+
+[[configprofile]]
+== Config Profile
+
+Config Profile indicates the project phase, such as dev, testing, live, etc. 
+
+=== Specify Config Profile
```
struberg, line 27: Why not use the term which is established in JavaEE: ProjectStage ?

This also has the benefit, that there is always a ProjectStage. Even if you don't specify any, you'll fallback to Production. This allows the following neat use case:

> myproject.security.enabled.Production=true
> myproject.security.enabled=false

Which means: enable security in production, but have it disabled anywhere else.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,77 @@
...
+
+[source, text]
+----
+java -jar myapp.jar -Dmp.config.profile=testing
+----
+
+The value of the property `mp.config.profile` shouldn't be updated after the application is started. It's only read once during the application start up. If the property value is modified afterwards, the behavior is undefined and any changes to its value made later can be ignored by the implementation. 
+
+The value of the property `mp.config.profile` specifies a single active profile. Only one profile can be active at a time. Commas in the value have no special meaning. For example, if the value of `mp.config.profile` is `testing,live`,  a single profile named `testing,live` is active instead of two active profiles. 
+If the property `mp.config.profile` is specified in multiple config sources, the value of the property is determined following the same rules as other configuration properties, which means the value in the config source with the highest ordinal wins.
```
struberg, line 40: This information is redundant. Standard rules are, well ... standard ;)

Emily-Jiang: In the earlier conversation, it was asked for clarification. If we treat this `mp.config.profile` exactly same as other properties, we can just say the rules on config property applies.

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: I think prefix or postfix both works. SmallRye Config went with prefix. I think what you listed above can be done using prefix as well.
`%dev.vehicle.name` stands out more than `vehicle.name.dev` though.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,77 @@
...
+
+=== How Config Profile works
```
jbee, line 42: This section isn't as clear as I'd hope. The way I understand it profiles do not have an effect beyond their source. If a source has either the original property or the profile specific property and the highest ordinal it wins (as usual) whereby naturally the profile specific property (if present) does override the original property value within that winning source.

spec/src/main/asciidoc/configprofile.asciidoc
radcortez: The issue with postfixing is that since a profile name can be anything you can encounter clashes. So `vehicle.name.something`, how do I know if `something` refers to a profile and I should set up `vehicle.name`, or it is an actual property I should setup `vehicle.name.something.something`?

spec/src/main/asciidoc/configprofile.asciidoc
radcortez: In my opinion, `profile` seems to be a more "popular" term to specify the configuration. It is used by Spring for instance.

Is there any particular reason why EE went with `ProjectStage` instead of `Profile`?

spec/src/main/asciidoc/configprofile.asciidoc
radcortez: I think we should be consistent and "eat our own dog food" :)

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: I think I will stick with `profile` for now.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
Emily-Jiang: @OndroMih I think only you argue for this property to be specified as env or system properties. @struberg @jbee both agreed to use normal property to set `mp.config.profile`. I will specify `mp.config.profile` as a normal property.

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: @jbee your undestanding is correct. Within the same config source, the active profile specific property overrides the original property value. Overall, the property from the highest ordinal wins no matter whether it is the original property or the active profile property.

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: I will keep `profile` for now.

Emily-Jiang:
I have done all of the necessary updates and provided explaination if your comments did not result in a update. I would like to merge this PR this week. Please provide your feedback in the next couple of days. If you are happy with the PR, please approve.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,77 @@
...
+
+The config profile can be specified via the property `mp.config.profile`, which can be set in any of the configuration sources. The value of the property can contain only characters that are valid for config property names. 
+This is because the name of the profile is directly stored in the name of the config property. It can be set when starting your application. e.g.
+
+[source, text]
+----
+java -jar myapp.jar -Dmp.config.profile=testing
+----
+
+The value of the property `mp.config.profile` shouldn't be updated after the application is started. It's only read once and will not be update once the `Config` object is constructed. If the property value of `mp.config.profile` is modified afterwards, the behavior is undefined and any changes to its value made later can be ignored by the implementation.
```
radcortez, line 37: be update -> be updated

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -0,0 +1,77 @@
...
+
+=== How Config Profile works
+
+The configuration property that utilizes the Config Profile is called a "profile-specific" property. A "profile-specific" property name consists of the following sequence: `% <profile name>.<original property name>`.
+
+Conforming implementations are required to search for a configuration source with the highest ordinal (priority) that provides either the property name or the "profile-specific" property name. 
+If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned. If it doesn't contain the "profile-specific" name, the value of the plain property will be returned. 
+
+
+For an instance, a config source can be specified as follows.
```
radcortez, line 50: Remove `an`?

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/CustomConfigProfileConfigSource.java
```diff
@@ -0,0 +1,65 @@
...
+        configValues.put("%prod.vehicle.name", "bus");
+        configValues.put("%test.vehicle.name", "van");
+        configValues.put("vehicle.name", "car");
+    }
+
+    @Override
+    public int getOrdinal() {
+        return 500;
+    }
+    @Override
```
radcortez, line 47: Add blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/CustomConfigProfileConfigSource.java
```diff
@@ -0,0 +1,65 @@
...
+    public String getValue(String key) {
+        return configValues.get(key);
+    }
+
+    @Override
+    public String getName() {
+        return "customConfigProfileSource";
+    }
+
+
```
radcortez, line 57: Remove additional line.

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/CustomConfigProfileConfigSource.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public Set<String> getPropertyNames() {
+        
+        return configValues.keySet();
+    }
+
+
```
radcortez, line 64: Remove blank lines.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/ProdProfileTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.testng.annotations.Test;
+
+/**
+ * Test cases for Config profile
+ * 
+ * @author Emily Jiang
+ */
+public class ProdProfileTest extends Arquillian {
+
+
```
radcortez, line 53: Remove additional blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/ProdProfileTest.java
```diff
@@ -0,0 +1,102 @@
...
+        WebArchive war = ShrinkWrap
+                .create(WebArchive.class, "ProdProfileTest.war")
+                .addAsLibrary(testJar);
+        return war;
+       
+    }
+
+
+
```
radcortez, line 80: Remove additional blank lines.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/ProdProfileTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @Test
+    public void testConfigProfileWithDev() {
+        ProfilePropertyBean bean = CDI.current().select(ProfilePropertyBean.class).get();
+
+        assertThat(bean.getConfigProperty(), is(equalTo("bus")));
+        assertThat(ConfigProvider.getConfig().getValue("vehicle.name", String.class), is(equalTo("bus")));
+    }
+
+
```
radcortez, line 89: Remove additional blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/ProdProfileTest.java
```diff
@@ -0,0 +1,102 @@
...
+        @Inject
+        @ConfigProperty(name="vehicle.name")
+        private String vehicleName;
+
+        public String getConfigProperty() {
+            return vehicleName;
+        }
+    }
+    
+
```
radcortez, line 101: Remove additional blank lines.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/TestConfigProfileTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.testng.annotations.Test;
+
+/**
+ * Test cases for Config profile
+ * 
+ * @author Emily Jiang
+ */
+public class TestConfigProfileTest extends Arquillian {
+
+
```
radcortez, line 53: Remove additional blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/TestConfigProfileTest.java
```diff
@@ -0,0 +1,100 @@
...
+        AbstractTest.addFile(testJar, "META-INF/microprofile-config.properties");
+
+        WebArchive war = ShrinkWrap
+                .create(WebArchive.class, "TestConfigProfileTest.war")
+                .addAsLibrary(testJar);
+        return war;
+       
+    }
+
+
```
radcortez, line 79: Remove additional blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/TestConfigProfileTest.java
```diff
@@ -0,0 +1,100 @@
...
+        @Inject
+        @ConfigProperty(name="vehicle.name")
+        private String vehicleName;
+
+        public String getConfigProperty() {
+            return vehicleName;
+        }
+    }
+    
+
```
radcortez, line 99: Remove additional blank lines.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/TestCustomConfigProfile.java
```diff
@@ -0,0 +1,104 @@
...
+import org.testng.annotations.Test;
+
+/**
+ * Test cases for Config profile
+ * 
+ * @author Emily Jiang
+ */
+public class TestCustomConfigProfile extends Arquillian {
+
+
```
radcortez, line 55: Remove additional blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/TestCustomConfigProfile.java
```diff
@@ -0,0 +1,104 @@
...
+        AbstractTest.addFile(testJar, "META-INF/microprofile-config.properties");
+
+        WebArchive war = ShrinkWrap
+                .create(WebArchive.class, "TestConfigProfileTest.war")
+                .addAsLibrary(testJar);
+        return war;
+       
+    }
+
+
```
radcortez, line 83: Remove additional blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/TestCustomConfigProfile.java
```diff
@@ -0,0 +1,104 @@
...
+        @Inject
+        @ConfigProperty(name="vehicle.name")
+        private String vehicleName;
+
+        public String getConfigProperty() {
+            return vehicleName;
+        }
+    }
+    
+
```
radcortez, line 103: Remove additional blank lines.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
```diff
@@ -0,0 +1,100 @@
...
+                        "%dev.vehicle.name=bike\n" +
+                        "%prod.vehicle.name=bus\n" +
+                        "%test.vehicle.name=van\n" +
+                        "vehicle.name=car"),
+                        "microprofile-config.properties")
+                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
+                .as(JavaArchive.class);
+
+                
+
```
radcortez, line 70: Remove additional blank lines.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
```diff
@@ -0,0 +1,100 @@
...
+        WebArchive war = ShrinkWrap
+                .create(WebArchive.class, "DevConfigProfileTest.war")
+                .addAsLibrary(testJar);
+        return war;
+       
+    }
+
+
```
radcortez, line 78: Remove additional blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
```diff
@@ -0,0 +1,100 @@
...
+    @Test
+    public void testConfigProfileWithDev() {
+        ProfilePropertyBean bean = CDI.current().select(ProfilePropertyBean.class).get();
+
+        assertThat(bean.getConfigProperty(), is(equalTo("bike")));
+        assertThat(ConfigProvider.getConfig().getValue("vehicle.name", String.class), is(equalTo("bike")));
+    }
+
+
```
radcortez, line 87: Remove additional blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
```diff
@@ -0,0 +1,100 @@
...
+        @Inject
+        @ConfigProperty(name="vehicle.name")
+        private String vehicleName;
+
+        public String getConfigProperty() {
+            return vehicleName;
+        }
+    }
+    
+
```
radcortez, line 99: Remove additional blank lines.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/InvalidConfigProfileTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.testng.annotations.Test;
+
+/**
+ * Test cases for Config profile
+ * 
+ * @author Emily Jiang
+ */
+public class InvalidConfigProfileTest extends Arquillian {
+
+
```
radcortez, line 53: Remove additional blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/InvalidConfigProfileTest.java
```diff
@@ -0,0 +1,102 @@
...
+        AbstractTest.addFile(testJar, "META-INF/microprofile-config.properties");
+
+        WebArchive war = ShrinkWrap
+                .create(WebArchive.class, "InvalidConfigProfileTest.war")
+                .addAsLibrary(testJar);
+        return war;
+       
+    }
+
+
```
radcortez, line 79: Remove additional blank line.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/InvalidConfigProfileTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @Test
+    public void testConfigProfileWithDev() {
+        ProfilePropertyBean bean = CDI.current().select(ProfilePropertyBean.class).get();
+
+        assertThat(bean.getConfigProperty(), is(equalTo("car")));
+        assertThat(ConfigProvider.getConfig().getValue("vehicle.name", String.class), is(equalTo("car")));
+    }
+
+
+
```
radcortez, line 89: Remove additional blank lines.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/InvalidConfigProfileTest.java
```diff
@@ -0,0 +1,102 @@
...
+        @Inject
+        @ConfigProperty(name="vehicle.name")
+        private String vehicleName;
+
+        public String getConfigProperty() {
+            return vehicleName;
+        }
+    }
+    
+
```
radcortez, line 101: Remove additional blank lines.

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: updated

spec/src/main/asciidoc/configprofile.asciidoc
Emily-Jiang: well spotted

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/CustomConfigProfileConfigSource.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/CustomConfigProfileConfigSource.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/ProdProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/ProdProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/ProdProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/TestConfigProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/TestConfigProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/TestCustomConfigProfile.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/DevConfigProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/InvalidConfigProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/InvalidConfigProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/InvalidConfigProfileTest.java
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/TestCustomConfigProfile.java
Emily-Jiang: done

Emily-Jiang:
@radcortez thanks for your feedback! I've addressed all of your comments, I think. Please double check and approve or fill more comments!

radcortez:
Thanks! +1

OndroMih:
Hi @Emily-Jiang, I'd be grateful if you allowed for more than 2 days to react. I don't work 100% on MicroProfile and there are also otehr PRs open that it's very easy to miss a 2 day deadline. I only noticed your deadline after you'd merged this PR actually.

@struberg gave a very reasonable feedback a few days ago, which I'd like to discuss further. When I review the discussion, most of the time it's just @struberg's or my opinion against @radcortez or @dmlloyd, both from the Quarkus team. So I don't understand why proposals from the Quarkus team always win without further discussion. 

Regarding support for setting the profile in any config source, I'm now OK with that since most of the people seem to support it or at least aren't against. Since that was the most important unresolved issue for me I'm OK with merging this PR. But I'm not OK with how it was finally merged with just an approval by a single person and with the lack of time to react before merging it.

Emily-Jiang:
@OndroMih this PR has been open for about 2 months. I have noted on the gitter and ask for feedback. Both @jbee and @radcortez are happy with the changes based on today's hangout. By the way, you can always raise further issues if you are unhappy with any issues. I talked with @struberg regarding the postfix and prefix and eventually he was ok and he is also happy with the profile property specified in any config sources.

radcortez:
Hi, @OndroMih I understand your frustration, but please don't feel that your opinion is being disregarded. If that happened, I'm sure everyone else is willing to review the changes and make any adjustments.

Also, I would like to tell you that I have a tremendous amount of respect for you and I value your opinions (and everyone else for that matter in this group). I hope I can convince you that this is not a Quarkus team vs everyone else. To make it clear, I'm in the SmallRye team, which provides the Config implementation not only to Quarkus, but also to Wildlfy, JBoss EAP, Thorntail, and now Websphere, so Quarkus is not my only concern.

For instance, I advocate that the config property for the profile to follow the same rules as everything else. Quarkus doesn't support that and works in the same way you were proposing. On the other hand,  I felt that this is was not consistent with our spec, so I've pushed back. In the end, Quarkus is probably going to have a similar issue with yours on how to implement it.

In the ConfigValue discussion, I've proposed a solution, you reviewed and made your proposal. I've updated the PR and it now reflects more of your proposal than mine. And that is fine because it made sense and we were able to collaborate on that.

I hope that you realize that I'm here to help and not to follow an agenda. Of course, we are not always going to agree and that is ok. As long as we can move forward its all good.

OndroMih:
Thanks for clarification, @radcortez and @emily-jiang. I respect both of you deeply but I was afraid that we're moving too fast with this because yhis is a huge change and the PR is huge by itself. I'm very busy these days and I can hardly find time to review one PR a week so having 3 or more PRs open at once is challanging to follow. I'd appreciate your understanding and give longer notice before merging. Since so many people eventually agreed with the final version of this PR, I'm OK with metging it although I don't fully agree with it. Especially my concerns about configuring the config by config itself may cause headaches to implementers and may also be confusing to users without much added value. Before we release the next version of Config I'd like to have a POC impl by both Quarkus and some app server, e. g. Payara, OpenLiberty or WildFly

OndroMih:
And BTW, I like the final version of the ConfigValue PR. Thanks, @radcortez. I believe we can still build on that and add additional capability to ConfigValue for more usecases.

radcortez:
Hey @OndroMih. I've already implemented this in SmallRye Config. If you want to have a look on how it is done, check here:

https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptorContext.java
https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java#L254

PR #559, opened 2020-04-30
Adds ConfigValueResolver fluent API

jbee:
### Summary
This PR introduces a fluent API which has multiple intents:

* cover more sophisticated use case of property resolution
* provide an API that is convenient to use, that means it assumes defaults that are likely to be intended when resolving configuration values and provides ways to disable individual defaults for untypical use cases

### Default Behaviours
The defaults here are not to throw exception and to never return null or empty `String`.
This should allow to replace `try`-`catch` constructs as well as `if (...isPresent())` constructs to resolve a configuration to the value finally used by the caller. The caller can use methods on the fluent API to opt-out of the defaults one-by-one to adopt the behaviour to the use case.

### Default Values
This support both defaults on the source level (`String`s) as well as typed defaults.
Source level defaults take precedence as they are considered to operate on the source level whereas typed defaults are considered to operate in connection with the later conversion stage. 

### Ad-hoc Conversion
The `asConvertedBy` method for ad-hoc conversion using lambdas intentionally does not use `Converter` type as argument to avoid binding the user to that API and because `Converter`s are required to be `Serializable` which is not needed for ad-hoc conversion.

### API Design Goals
The methods of  `ConfigValueResolver` can be used in any order and even multiple times without leading to conflicting definition. The actual value resolution can only occur once and forces the caller to either provide a default or to work with `Optional`s instead - avoiding exceptions, empty or null values unless intended by the caller.

Naming of the methods has been chosen so that when used fluently the code provides good readability, for example:
```java
// basic case: resolve value with typed default
int val1 = config.getValue("myint").as(Integer.class, 42);
// with empty list default
List<Integer> val2 = config.getValue("mylist").asList(Integer.class);
// or with explicit default
List<Integer> val3 = config.getValue("mylist").asList(Integer.class, Arrays.asList(42));
// same for sets with empty default
Set<Integer> val4 = config.getValue("myset").asSet(Integer.class);
// or explicit default
Set<Integer> val5 = config.getValue("myset").asSet(Integer.class, new HashSet<>(Arrays.asList(42)));

// opt-out of defaults
int val6 = config.getValue("myint").throwOnMissingProperty().as(Integer.class, 42);
int val7 = config.getValue("myint").throwOnFailedConversion().as(Integer.class, 42);
String spacer = config.getValue("spacer").acceptEmpty().as(String.class, ","); // can now be set empty explicitly

// use raw (source) level defaults
List<Integer> val8 = config.getValue("mylist").withDefault(annotation.defaultValue()).asList(Integer.class);
//... and combinations of above and more
```

### API Integration with other APIs and Concepts
I went ahead and removed methods from `Config` which I belief are not yet released and could be covered by the fluent API instead. This would also be the idea for the future that `Config` API is kept cleaner by not having to cover many variants of value resolution. This is extracted to the `ConfigValueResolver` which makes it easier for users to customise the logic of value resolution.

It should also be clear that the `ConfigValueResolver` could be implemented on top of `Config` by vendors or users using API provided by `Config` interface. The intent with adding a `ConfigValueResolver` to the standard is to make user code using such a "convenience API" portable to not lock in users to a vendor API or to build their own API on top of MP Config.

I also added a `withProfile` method to the API mostly to illustrate that this can also be used to handle profiles should those become a features further preventing the need to blur the responsibilities of the `Config` API.

The  `ConfigValueResolver` could even replace the existing `Config.getOptionalValue`; `getOptionalValue("myprop", MyClass.class)` would become `getValue("myprop").as(MyClass.class)`. For better backwards compatibility I left it in but removal could be considered as the replacement is equally short and readable.

This differs from the https://github.com/struberg/microprofile-config/blob/ConfigAccessor/api/src/main/java/org/eclipse/microprofile/config/ConfigAccessor.java API by being a pure convenience API that does not provide new concepts (like snapshots) or adds new capabilities (like caching). While the `ConfigAccessor` is meant to be created "statically" (once) and then be "applied" to an actual `Config` or snapshot the `ConfigValueResolver` is only a utility level on top of the actual `Config` that always needs to be used starting from the current `Config` instance. While the `ConfigAccessor` clearly has its benefits this should be seen as an alternative trying to make the most out of a simpler approach that purely focuses on a more convenient and readable resolution of values. Also keeping in mind that it still holds true, that configurations that are accessed frequently should be buffered by the caller.

Signed-off-by: Jan Bernitt <jaanbernitt@gmail.com>

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
@@ -30,6 +30,8 @@ import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
+import org.eclipse.microprofile.config.spi.Converter;
+
 /**
  * <p>
  * This is an abstraction to resolve {@link Config} values using a fluent API that is used
@@ -57,7 +59,7 @@ import java.util.function.Function;
  *
  * <p>
  * Usually conversion relies on registered {@link org.eclipse.microprofile.config.spi.Converter}s.
- * Ad-hoc conversion can be done using {@link #asConvertedBy(Function, Object)}.
+ * Ad-hoc conversion can be done using {@link #asConvertedBy(Converter, Object)}.
  * Values resolved as {@link String} are not converted.
  * Values resolved as {@code String[]} are only split but elements are not converted.
  *
@@ -77,15 +79,6 @@ import java.util.function.Function;
  * @author <a href="mailto:jan.bernitt@payara.fish">Jan Bernitt</a>
  */
 public interface ConfigValueResolver {
-
-    /**
-     * Adds a profile to property resolution which is applied when requesting a value.
-     * Multiple calls to this method on the same instance override prior profile values.
-     *
-     * @param value name of the profile to use, {@code null} is ignored
-     * @return This resolver for fluent API usage
-     */
-    ConfigValueResolver withProfile(String value);
 
     /**
      * Provides a raw property value default value.
@@ -180,7 +173,7 @@ public interface ConfigValueResolver {
      * @throws java.util.NoSuchElementException if the property isn't present in the configuration and throwing
      *         exceptions has been requested using {@link #throwOnMissingProperty()}
      **/
-    <T> T asConvertedBy(Function<String, T> converter, T defaultValue);
+    <T> T asConvertedBy(Converter<T> converter, T defaultValue);
 
     /**
      * Resolves the property as {@link List}.
--- a/api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
@@ -1,0 +1,294 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *   2020-04-30 - Jan Bernitt / Payara
+ *      Initial revision
+ *
+ *******************************************************************************/
+package org.eclipse.microprofile.config;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * <p>
+ * This is an abstraction to resolve {@link Config} values using a fluent API that is used
+ * when the basic {@link Config#getValue(String, Class)} does not suffice
+ * or the user wants to make sure a value is returned without an exception and excluding empty {@link String} values.
+ * The method names are chosen for good readability when used as a fluent API.
+ *
+ * <p>
+ * This API is designed so reliably result in a return value. This means by default it will not throw exceptions for
+ * missing properties or failed conversion but instead return a default value that callers have to provide.
+ * Alternatively to providing a default value a value can be resolved as {@link Optional}.
+ *
+ * <p>
+ * Arrays of values can be resolved as {@link List} or {@link Set}.
+ * By default these return empty lists or sets in case of missing property or failed conversion.
+ *
+ * <p>
+ * Simple value properties defined as empty string by default are considered missing. This can be changed
+ * using {@link #acceptEmpty()}.
+ *
+ * <p>
+ * Should exceptions be thrown for either missing properties or failed conversion the default of not throwing exception
+ * and using default values can be overridden using {@link #throwOnMissingProperty()} and
+ * {@link #throwOnFailedConversion()}.
+ *
+ * <p>
+ * Usually conversion relies on registered {@link org.eclipse.microprofile.config.spi.Converter}s.
+ * Ad-hoc conversion can be done using {@link #asConvertedBy(Function, Object)}.
+ * Values resolved as {@link String} are not converted.
+ * Values resolved as {@code String[]} are only split but elements are not converted.
+ *
+ * <p>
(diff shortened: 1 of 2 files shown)


genie-microprofile:

api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
```diff
@@ -0,0 +1,294 @@
...
+    ConfigValueResolver withDefault(String value);
+
+    /**
+     * Disables the default behaviour of considering properties defined as empty string as being not present (missing).
+     * Effectively this means {@code as}-methods might then return empty things in case the property is defined as empty
+     * in the source.
+     *
+     * @return This resolver for fluent API usage
+     */
+    ConfigValueResolver acceptEmpty();
```
dmlloyd, line 109: This is actually inconsistent with how empty values must work.  An empty string is an empty string.  They're always accepted.  They're always present (but empty).

api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
```diff
@@ -0,0 +1,294 @@
...
+     * Provides a raw property value default value.
+     * <p>
+     * The raw default is considered to be on the source level and therefore takes precedence over typed defaults
+     * provided to any of the {@code as}-methods. Should raw default be empty of fail to convert to the target type the
+     * typed default is used.
+     *
+     * @param value raw default value, not {@code null}
+     * @return This resolver for fluent API usage
+     */
+    ConfigValueResolver withDefault(String value);
```
dmlloyd, line 100: It is my view that having non-global default values will only add a layer of confusion.  In Quarkus, users were expecting that default values be included in property expression resolution, and we had to retrofit in order to support that.  This just reintroduces the problem again.  Default values should be configured when the configuration is being built, not at evaluation time (after all a configuration is defined to apply to a single application, so it should be possible to know what the default values would be at that stage).

api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
```diff
@@ -0,0 +1,294 @@
...
+    ConfigValueResolver acceptEmpty();
+
+    /**
+     * Disables the default behaviour of not throwing exceptions and instead returning default values for case of
+     * missing or empty raw value. If used in combination with {@link #acceptEmpty()} empty raw values will not throw an
+     * exception, otherwise they do.
+     *
+     * @return This resolver for fluent API usage
+     */
+    ConfigValueResolver throwOnMissingProperty();
```
dmlloyd, line 118: If a configuration value is an empty string, it is because a user has deliberately deleted that property value, explicitly overriding the default value.  See #531 for the use case.  It's not a question of throwing or not throwing.  Empty values should always be represented by `Optional*.empty()` which works for any value type.

api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
```diff
@@ -0,0 +1,294 @@
...
+
+    /**
+     * Disables the default behaviour of not throwing exceptions and instead returning default values for case of failed
+     * conversion or missing converter for the target type.
+     *
+     * @return This resolver for fluent API usage
+     */
+    ConfigValueResolver throwOnFailedConversion();
```
dmlloyd, line 126: What is the use case for this?  I think in most cases this would just lead to confusion.

api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
```diff
@@ -0,0 +1,294 @@
...
+     * @param defaultValue any value including null. It is returned in case of missing property or failed conversion
+     *                     unless throwing exceptions has been explicitly requested using
+     *                     {@link #throwOnMissingProperty()} or {@link #throwOnFailedConversion()}.
+     * @return the resolved and converted property value or the default value
+     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type and throwing
+     *         exceptions has been requested using {@link #throwOnFailedConversion()}
+     * @throws java.util.NoSuchElementException if the property isn't present in the configuration and throwing
+     *         exceptions has been requested using {@link #throwOnMissingProperty()}
+     **/
+    <T> T asConvertedBy(Function<String, T> converter, T defaultValue);
```
dmlloyd, line 183: Why are we suddenly using `Function<String,T>` for conversion?

api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
```diff
@@ -0,0 +1,294 @@
...
+public interface ConfigValueResolver {
+
+    /**
+     * Adds a profile to property resolution which is applied when requesting a value.
+     * Multiple calls to this method on the same instance override prior profile values.
+     *
+     * @param value name of the profile to use, {@code null} is ignored
+     * @return This resolver for fluent API usage
+     */
+    ConfigValueResolver withProfile(String value);
```
dmlloyd, line 88: Does it make sense to be able to change the profile as we run?  Are we causing problems for implementations that want to cache properties up front?

jbee:
For the benefit of those that didn't attend today's meeting: It might help to know that the API as suggested initially uses the philosophy of "best effort" possible to result in a value as opposed to the usual "defensive" philosophy of failing fast. This is purely chosen because it matches the usage I found in our code base which usually does "best effort" logic and because you can have an API that does "best effort" helping you to use it wrongly with the option of opting out into "fail fast" which I can't see working the other way around.

api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
jbee: This was explained in the PR description. Anyhow, it was discussed 2 weeks ago and I changed it to `Converter`.

api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
jbee: Removed.

api/src/main/java/org/eclipse/microprofile/config/ConfigValueResolver.java
Emily-Jiang: The best practices for mandating the global default is to specify the value in `microprofile-config.properties`. I am not sure whether all users give the same defaultValue on multiple property injections or not? Do you have any user feedback @dmlloyd on this?

dmlloyd: It is required in Quarkus. Nobody has complained.

jbee:
As discussed we are going to think about spec level first. PR on hold or we may even close it.

Emily-Jiang:
@jbee will do some protype first and open a new PR. Close this PR for now.

PR #560, opened 2020-05-05
Fix issue when iterating and asserting unknown env or system properties that may not have the expected result, for instance if property expansion is supported.

radcortez:


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
```diff
@@ -88,28 +88,18 @@ public static void setupCheck(){
...
-        Properties properties = System.getProperties();
-        for (Map.Entry<String, String> envEntry : env.entrySet()) {
-            String key = envEntry.getKey();
-            if(!properties.containsKey(key)) {
-                String value = envEntry.getValue();
-                String cfgValue = config.getValue(key, String.class);
-                Assert.assertEquals(value, cfgValue, "The property "+key+" did not have the expected value.");
-            }
-        }
+        String value = System.getenv().get("MP_TCK_ENV_DUMMY");
```
Emily-Jiang, line 91: I suggest to also include some builtin environment variables, such as `PATH` etc to demonstrate not only the properties we added but also the builtin ones.

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java
```diff
@@ -88,28 +88,18 @@ public static void setupCheck(){
...
-        for (Map.Entry<Object, Object> propEntry : properties.entrySet()) {
-            String propValue = (String) propEntry.getValue();
-            if (propValue != null && propValue.length() > 0) {
-                Assert.assertEquals(propValue, config.getValue((String) propEntry.getKey(), String.class));
-            }
-        }
+        String value = System.getProperties().getProperty("mp.tck.prop.dummy");
+        Assert.assertNotNull(value);
+        Assert.assertEquals("dummy", value);
+        Assert.assertEquals(value, config.getValue("mp.tck.prop.dummy", String.class));
```
Emily-Jiang, line 102: Same as above. Can you also check some built in jvm properties such as java.home, user.name?

Emily-Jiang:
@radcortez the PR looks good! As mentioned on the PR, can you log a bug issue for our release record?

Issue #549, opened 2020-03-19
URL.equals() is not a 100% reliable method to use in a TCK


tevans78:
There are a number of tests which rely on two URLs being equal and use the standard URL.equals() method to check that. However, the equals method for a URL does not just do a simple comparison of the URL String. The javadoc says...
```
Two URL objects are equal if they have the same protocol, reference equivalent hosts, have the same port number on the host, and the same file and fragment of the file.

Two hosts are considered equivalent if both host names can be resolved into the same IP addresses; else if either host name can't be resolved, the host names must be equal without regard to case; or both host names equal to null.

Since hosts comparison requires name resolution, this operation is a blocking operation. 
```
It is possible for two URL instances to be correctly created using the same String but for URL.equals to return false if, for some reason, one of the URLs can not be resolved at runtime. This failure would probably be due to a momentary network issue and nothing to do with the MP Config implementation.

Clearly, this is an extreme edge case that we have only seen fail once, on AIX. While there are several tests that use URL, the test that failed for us was https://github.com/eclipse/microprofile-config/blob/dea55bee3f61f40703fc759f04d374c3ed6a19f2/tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java#L931

The stack trace was
```
testUrlArrayInjection_EE7_FEATURES:
      java.lang.AssertionError: Lists differ at element [0]: http://microprofile.io != http://microprofile.io expected [http://microprofile.io] but found [http://microprofile.io]
at org.testng.Assert.fail(Assert.java:94)
at org.testng.Assert.failNotEquals(Assert.java:496)
at org.testng.Assert.assertEquals(Assert.java:125)
at org.testng.Assert.assertEquals(Assert.java:551)
at org.testng.Assert.assertEquals(Assert.java:670)
at org.testng.Assert.assertEquals(Assert.java:727)
at org.eclipse.microprofile.config.tck.ArrayConverterTest.testUrlArrayInjection(ArrayConverterTest.java:764)
```

dmlloyd:
Yes this is definitely a bug.  Thanks for the report!

Emily-Jiang:
Since URL.equals() performs DNS resolution, it was advised not to use it. What people think if we use URI.equals() instead?

PR #567, opened 2020-05-26
#549 fix URL equals

Emily-Jiang:

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
@@ -23,6 +23,7 @@ import static org.eclipse.microprofile.config.tck.base.AbstractTest.addFile;
 
 import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.time.Duration;
 import java.time.Instant;
@@ -33,9 +34,12 @@ import java.time.OffsetDateTime;
 import java.time.OffsetTime;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.IntStream;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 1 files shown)


Emily-Jiang:
@radcortez please review!

Emily-Jiang:
@jbee please review

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
```diff
@@ -1654,120 +1658,161 @@ public void testInstantSetInjection() {
...
+                return expectedValue[i].toURI().equals(value[i].toURI());
+        } 
+        catch (URISyntaxException e) {
+                e.printStackTrace();
+        }
+        return false;
+        }));
+    }
+
+    private void assertURLListEquals(List<URL> value) throws MalformedURLException {
```
jbee, line 1677: I'd think that either of the `assertURLArrayEquals` and `assertURLListEquals` could be implemented in terms of the other. I'd do `assertURLArrayEquals` => `assertURLListEquals(asList(value))`

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
```diff
@@ -1654,120 +1658,161 @@ public void testInstantSetInjection() {
...
+        }));
+    }
+
+
     @Test
-    public void testUrlArrayLookupProgrammatically() throws MalformedURLException {
+    public void testUrlArrayLookupProgrammatically() throws MalformedURLException, URISyntaxException {
         URL[] value = config.getValue("tck.config.test.javaconfig.converter.urlvalues", URL[].class);
         Assert.assertNotNull(value);
         Assert.assertEquals(value.length, 3);
```
jbee, line 1698: If there is a special method to assert equality I'd think the length and null check can or should be part of it so they do not need duplication.

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
```diff
@@ -1654,120 +1658,161 @@ public void testInstantSetInjection() {
...
-        Assert.assertEquals(value, new URL[]{
-            new URL("http://microprofile.io"),
-            new URL("http://openliberty.io"),
-            new URL("http://microprofile.io")});
+        assertURLArrayEquals(value);
         URL[] single = config.getConverter(URL[].class).get().convert("http://microprofile.io");
         Assert.assertNotNull(single);
         Assert.assertEquals(single.length, 1);
-        Assert.assertEquals(single, new URL[]{ new URL("http://microprofile.io") });
+        Assert.assertEquals(single[0].toURI(), new URL[]{ new URL("http://microprofile.io") }[0].toURI());
```
jbee, line 1715: Similarly to the extracted method to assert an array with 3 elements it seems worth it to extract one for the array with single element (including the null and length check) as it is a reoccurring assertion pattern.

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
```diff
@@ -1654,120 +1658,161 @@ public void testInstantSetInjection() {
...
-            new URL("http://openliberty.io"),
-            new URL("http://microprofile.io")));
+        assertURLListEquals(values);
         Optional<List<URL>> optionalSingle = config.getOptionalValues("tck.config.test.javaconfig.converter.urlvalue", URL.class);
         Assert.assertTrue(optionalSingle.isPresent());
         List<URL> single = optionalSingle.get();
         Assert.assertNotNull(single);
         Assert.assertEquals(single.size(), 1);
-        Assert.assertEquals(single, Arrays.asList(new URL("http://microprofile.io")));
+        Assert.assertEquals(single.get(0).toURI(), Arrays.asList(new URL("http://microprofile.io")).get(0).toURI());
```
jbee, line 1759: `Arrays.asList(new URL("http://microprofile.io")).get(0)` should just be `new URL("http://microprofile.io")`

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
```diff
@@ -1654,120 +1658,161 @@ public void testInstantSetInjection() {
...
-            new URL("http://microprofile.io"),
-            new URL("http://openliberty.io"),
-            new URL("http://microprofile.io")});
+    public void testUrlArrayInjection() throws MalformedURLException, URISyntaxException {
+        URL[] values = converterBean.getMyUrlArray();
+        Assert.assertEquals(values.length, 3);
+        assertURLArrayEquals(values);
         Assert.assertEquals(converterBean.getMySingleUrlArray().length, 1);
-        Assert.assertEquals(converterBean.getMySingleUrlArray(), new URL[]{ new URL("http://microprofile.io") });
+        Assert.assertEquals(converterBean.getMySingleUrlArray()[0].toURI(), new URL[]{ new URL("http://microprofile.io") }[0].toURI());
```
jbee, line 1769: Equally `new URL[]{ new URL("http://microprofile.io") }[0]` => `new URL("http://microprofile.io")`

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
```diff
@@ -1654,120 +1658,161 @@ public void testInstantSetInjection() {
...
-    public void testURLListInjection() throws MalformedURLException {
-        Assert.assertEquals(converterBean.getMyUrlList().size(), 3);
-        Assert.assertEquals(converterBean.getMyUrlList(), Arrays.asList(
-            new URL("http://microprofile.io"),
+    public void testURLListInjection() throws MalformedURLException, URISyntaxException {
+        List<URL> values = converterBean.getMyUrlList();
+        Assert.assertEquals(values.size(), 3);
+        assertURLListEquals(values);
+        Assert.assertEquals(converterBean.getMySingleUrlList().size(), 1);
+        Assert.assertEquals(converterBean.getMySingleUrlList().get(0).toURI(), Arrays.asList(new URL("http://microprofile.io")).get(0).toURI());
```
jbee, line 1778: unnecessary list wrapping again, maybe it is intentional? I find it more confusing.

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
```diff
@@ -1654,120 +1658,161 @@ public void testInstantSetInjection() {
...
+    }
+
+    private boolean assertURLSetEquals(Set<URL> valueSet, Set<URL> expectedURLSet) throws MalformedURLException, URISyntaxException {
+        if (valueSet.size()!= expectedURLSet.size()) {
+            return false;
+        }
+        Iterator<URL> it = valueSet.iterator();
+        boolean isEquals = true;
+        while (it.hasNext()) {
```
jbee, line 1787: Using a `while` and some `for` loops seems overly complicated and not very readable. I suggest to map to sets of URIs and use `uriSet1.equals(uriSet2)`

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
```diff
@@ -1654,120 +1658,161 @@ public void testInstantSetInjection() {
...
-        Assert.assertEquals(converterBean.getMySingleUrlList(), Arrays.asList(new URL("http://microprofile.io")));
-    }
-
-    @Test
-    public void testURLSetInjection() throws MalformedURLException {
-        Assert.assertEquals(converterBean.getMyUrlSet().size(), 2);
-        Assert.assertEquals(converterBean.getMyUrlSet(), new LinkedHashSet<>(Arrays.asList(
-                new URL("http://openliberty.io"),
-                new URL("http://microprofile.io"))));
+        Assert.assertTrue(assertURLSetEquals(values, expectedURLSet));
```
jbee, line 1811: As sets are used here I wonder if the lists and arrays used before are intentionally expecting an order or if those really should compare sets as well. Can say but wanted to point it out. 
Using a `LinkedHashSet` also bags the question if order does matter in which case these maybe should be compared as lists? Or use the default `HashSet` to leave no question that these are sets and order is not important.

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
Emily-Jiang: I think this is a short circuit. The whole test will be failed if the value is null or contains a different size of the items.

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
Emily-Jiang: ok. I will refactor.

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
Emily-Jiang: not really. We should not force ordering for sets.

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
Emily-Jiang: refactored.

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
Emily-Jiang: What you suggested will be equivalent URI comparison and defeat the purpose of the URL comparison, which means the set contains a collection of URLs.

Emily-Jiang: Most important thing is to test the right thing.

Emily-Jiang:
@jbee I have done most things you suggested and added explaintation for the comments I did not do. Please approve if you have no further comments.

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
jbee: > We should not force ordering for sets.

Then why use a `LinkedHashSet`?

tck/src/main/java/org/eclipse/microprofile/config/tck/ArrayConverterTest.java
jbee: The code already does `thisURL.toURI().equals(url.toURI())` (URI comparison) and is basically just a overly complicated way to compare two sets of URL in terms of their URI equality. Or I misread this code entirely. Anyhow. This was just a suggestion to make the code more readable.

Emily-Jiang:
Thanks @jbee !

PR #572, opened 2020-06-11
Fix TCK tests for Profile

radcortez:


genie-microprofile:

PR #576, opened 2020-06-24
Added unwrap API to Config.

radcortez:
This is a new proposal to add `unwrap` like proposed previously here:

https://github.com/eclipse/microprofile-config/issues/84
https://github.com/eclipse/microprofile-config/pull/85

I believe the use case is there, especially now where we want to experiment more in the implementation side of things.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -244,4 +244,19 @@ public interface Config {
      * @return an {@link Optional} containing the converter, or empty if no converter is available for the specified type
      */
     <T> Optional<Converter<T>> getConverter(Class<T> forType);
+
+    /**
+     * Returns an instance of the specific class, to allow access to the provider specific API.
+     * <p>
+     * If the MP Config provider implementation does not support the specified class, a {@link IllegalArgumentException}
+     * is thrown.
+     * <p>
+     * Unwraping to the provider specific API may lead to non-portable behaviour.
+     *
+     * @param type Class representing the type to unwrap to
+     * @param <T> The type to unwrap to
+     * @return An instance of the given type
+     * @throws IllegalArgumentException If the current provider does not support unwrapping to the given type
+     */
+    <T> T unwrap(Class<T> type);
 }


Emily-Jiang:
I am ok to add this. @struberg @jbee @OndroMih please comment.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -244,4 +244,17 @@
...
+     * <p>
+     * If the MP Config provider implementation does not support the specified class, {@link IllegalArgumentException}
+     * is thrown.
+     *
+     * @param type Class representing the type to unwrap to
+     * @param <T> The type to unwrap to
+     * @return An instance of the given type
+     * @throws IllegalArgumentException If the current provider does not support unwrapping to the given type
+     */
+    <T> T unwrap(Class<T> type);
```
jbee, line 259: Maybe the javadoc should mention if `T` must be a type that `extends Config` or can be any type. 

Maybe it should explain what the unwrap method is used for and if it should support unwrapping to any of the types wrapped around the innermost implementation class.

I assume that there are no limits to what the user can ask for, still the javadoc should make this clear and also that this is always non-portable behaviour. This might be obvious for most users but not for all.

Emily-Jiang: T should extends Config. The other thing is that in order to support this, the runtime need to expose the implementation class to the end users.

jbee: As long as it is clearly defined I am happy with it.

radcortez: Added some more detail to the java doc.

I don't think we should limit `T` to `extends Config`. First, to keep consistency on how other `unwrap` methods are implemented:

https://javaee.github.io/javaee-spec/javadocs/javax/persistence/Cache.html#unwrap-java.lang.Class-
https://javaee.github.io/javaee-spec/javadocs/javax/persistence/EntityManager.html#unwrap-java.lang.Class-
https://javaee.github.io/javaee-spec/javadocs/javax/validation/Validator.html#unwrap-java.lang.Class-
https://docs.oracle.com/en/java/javase/11/docs/api/java.sql/java/sql/Wrapper.html#unwrap(java.lang.Class)

And it will force the provided to return and actual implementation of `Config`. This shouldn't be an issue, because the caller already has an implementation of `Config` to call `unwrap`, so he could use one or the other if he wants to use proper `Config` API. This would allow the implementation to return specific implementations of APIs (maybe experimental), dealing with config semantics, that not necessary require a full implementation of `Config`.

Issue #396, opened 2019-01-14
Should Config.getValue() to support variable replacement


Emily-Jiang:
To address the issue #118, we have this PR #392 introducing a way on variable replacement. 

In the class ConfigAccessorBuilder.evaluateVariable(boolean evaluateVariables), we would like to suggest the evaluateVariable is on by default.

However, the above only impacts the variable values obtained via ConfigAccess. In ConfigProperty annotation, we also introduced a property evaluateVariables with the default value of true.

Base on this, it might make sense to support evaluateVaribles by default with Config.getValue(). Basically, if a config contains a variable replacement, the variable will be evaluated.

``` 
server.host=localhost
server.port=9080
server.url=http://${server.host}:${server.port}
```

If this issue is accepted, a backward incompatible change will be introduced. This will lead MP Config release to bump major version.

when calling config.getValue(server.url, String.class) will return
```
http://localhost:9080
```
If this issue is accepted, a backward incompatible change will be introduced. Hence, this issue will lead MP Config release to bump major version.

jmesnil:
I'd really like to avoid having a breaking change due to this feature.

As discussed during the last mp config meeting, we could move on to support variable evaluation *and* specifies that implementation MUST provide a way to support backwards compatibility.

My suggestion for this would be:
* starting with next release, variable evaluation is enabled by default
* implementation must disable variable evaluation if a well-known property (e.g. `mp-config.evaluate-variables.enabled` is configured on any of the default config sources (env vars, sys props & microprofile-config.properties)

This way an existing application could run without changing its code by adding the sys prop `-Dmp-config.evaluate-variables.enabled=false` or the env var `MP_CONFIG_EVALUATE_VARIABLES_ENABLED=false` and preserve its behaviour.

jmesnil:
We also need to have a proper spec for what variable evaluation.

I've starting working on a prototype and these are the questions I need to answer to move forward:
* Behaviour when a variable can not be evaluated
* Support for default value
* Support for recursiveness

Starting with your example `server.url=http://${server.host}:${server.port}`.

### What should happen when all variables can be evaluated?

```
server.host=localhost
server.port=9080
server.url=http://${server.host}:${server.port}
```
=> the expected value will be `http://localhost:9080`

### What should happen when some variables can not be evaluated?

```
server.host=localhost
# server.port is not defined
server.url=http://${server.host}:${server.port}
```
=> The API should throw a `NoSuchElementException` (or return an `Optional.empty()`).

[...]

Issue #118, opened 2017-04-03
Variable replacement


Emily-Jiang:
The issue is from Jeremie:
    my.property.version.major=1

    my.property.version.minor=0
    my.property.version=${my.property.version.major}.{my.property.version.minor}
    my.combined.property=Version ${my.property.version}

    my.combined.property.withdefault=Version ${my.undefined.property:1.0}

This is very effective for example the file structure of a service. The root directory configuration is the domain of the installation whereas the subdirectory is the domain of application.

This has an impact on the implementation because all properties must be known before solving them. In addition, all properties dependent on a mutable property become mutable as well.

struberg:
This already has been proposed in July last year
https://github.com/struberg/javaConfig/blob/configValue/api/src/main/java/javx/config/ConfigValue.java#L55

jerr:
Mark, you're absolutely right, in most cases that I've met, I also had properties with variables that were not to be computed.

But we can have two approaches for this:

- The first one, we can manage this on the API user side. As the code you specify, it is the user who specifies the management or how to calculate the property.

- The second, the management is transparent to the API user. You can restructure your properties hierarchy without changing a single line of code. This approach is perhaps also to be considered. It makes the properties management more flexible concerning the code.

Some examples that may be appropriate for the second approach:
```
my.propery.rootdirectory = /opt/myservice/data
my.propery.productdirectory = ${my.propery.rootdirectory}/${my.property.product.category:products}
my.property.product.templatefilename = ${product.id}-${product.name}.pdf
my.property.product.file = ${my.propery.productdirectory}/${my.property.product.templatefilename}

```
The result for the property my.property.product.file is "/opt/myservice/data/products/${product.id}-${product.name}.pdf"
With the default value of my.propery.productdirectory  and undefined properties of my.property.product.file we can manage that.

Sometimes you want to clean up your configuration files without affect the code. This situation has happened to me in several projects.

What do you think about this?

sgalles:
Hello,
Would this also work for replacement in the default values ?
In this case, I think this would automatically also allow this : https://github.com/eclipse/microprofile-config/issues/339
because you could simply define your alias as a second  property and just set the default value equals to the initial property
```java
@ConfigProperty(name = "my.foo")
...
@ConfigProperty(name = "my.fooalias", defaultValue = "${my.foo}")
```

Emily-Jiang:
@sgalles good suggestion. 
[...]

PR #577, opened 2020-06-25
Support Property Expressions.

radcortez:
Just the initial version to start the discussion. Once we have something that we agree on, I'll move with the TCK

This should handle:
- #118 
- #396

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
@@ -41,9 +41,16 @@ public interface ConfigValue {
     /**
      * The value of the property lookup.
      *
-     * @return the raw value of the property lookup or {@code null} if the property could not be found
+     * @return the value of the property lookup or {@code null} if the property could not be found
      */
     String getValue();
+
+    /**
+     * The value of the property lookup without any transformation (expansion, etc).
+     *
+     * @return the raw value of the property lookup or {@code null} if the property could not be found.
+     */
+    String getRawValue();
 
     /**
      * The {@link org.eclipse.microprofile.config.spi.ConfigSource} name that loaded the property lookup.
--- a/api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
@@ -39,14 +39,14 @@ public interface ConfigValue {
     String getName();
 
     /**
-     * The value of the property lookup.
+     * The value of the property lookup with transformations (expanded, etc).
      *
      * @return the value of the property lookup or {@code null} if the property could not be found
      */
     String getValue();
 
     /**
-     * The value of the property lookup without any transformation (expansion, etc).
+     * The value of the property lookup without any transformation (expanded , etc).
      *
      * @return the raw value of the property lookup or {@code null} if the property could not be found.
      */
--- a/spec/src/main/asciidoc/property-expressions.asciidoc
+++ b/spec/src/main/asciidoc/property-expressions.asciidoc
@@ -1,0 +1,76 @@
+//
+// Copyright (c) 2020 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// Contributors:
+// Roberto Cortez
+
+[[property-expressions]]
+== Property Expressions
+
+The value of a configuration property may contain an expression corresponding to another configuration property. An
+expression string is a mix of plain strings and expression segments, which are wrapped by the sequence `${ ... }`.
+
+Consider the following configuration properties file:
+
+[source,properties]
+----
+server.url=http://${server.host}/endpoint
+server.host=example.org
+----
+
+When looking up the `server.url` property, the value will be resolved and expanded to `http://example.org/endpoint`.
+All MicroProfile Config rules still apply. The `Config` is able to resolve expressions from different ConfigSources.
+
+Additionally, it is also possible to use the following syntax for property expressions:
+
+* `${expression:value}` - Provides a default value after the `:` if the expression doesn't find a value.
+* `${my.prop${compose}}` - Composed expressions. Inner expressions are resolved first.
+* `${my.prop}${my.prop}` - Multiple expressions.
+
+Consider the following configuration properties file:
+
+[source,properties]
+----
+server.url=http://${server.host:example.org}:${server.port}/${server.endpoint}
+server.port=8080
+server.endpoint=${server.endpoint.path.foo${server.endpoint.path.bar}}
+server.endpoint.path.foo=/foo
+server.endpoint.path.bar=/bar
+----
+
+The property `server.url` is expanded to `http://example.org:8080/foo/bar`.
+
+If an expression cannot be expanded and does not have a default value a `NoSuchElementException` is thrown. In the
+Optional case, returns an empty Optional.
+
+Composed expressions, are limited to the number of recurssion lookups performed. Implementations are encouraged to
+limit the number to `5`, but they can use higher limits if they wish to. When the number of allowed recursion lookups
+exceeds the limit, a `IllegalArgumentException` is thrown.
+
+Property expressions applies to lookup methods in `Config`, like `getValue`, `getConfigValue` and `getOptionalValue`.
+Lookup methods in `ConfigSource`, may support property expressions as well, but it is not required by the specification.
+
+The escape character is "\".
+
+=== Backwards Compatibility
+
+MicroProfile Config implementations MUST provide a way to disable variable evaluation to provide backwards
+compatibility. The property `mp.config.property.expressions` can be set to `false` and disable property expressions. By
+default, property expressions are enabled.
+
+If property expression expand is not desirable for a specific case, the raw value on a configuration property may be
+retrieved by calling `getRawValue()` in `ConfigValue`.
--- a/spec/src/main/asciidoc/property-expressions.asciidoc
+++ b/spec/src/main/asciidoc/property-expressions.asciidoc
@@ -47,30 +47,51 @@ Consider the following configuration properties file:
 ----
 server.url=http://${server.host:example.org}:${server.port}/${server.endpoint}
 server.port=8080
-server.endpoint=${server.endpoint.path.foo${server.endpoint.path.bar}}
+server.endpoint=${server.endpoint.path.${server.endpoint.path.bar}}
 server.endpoint.path.foo=/foo
-server.endpoint.path.bar=/bar
+server.endpoint.path.bar=foo
 ----
 
-The property `server.url` is expanded to `http://example.org:8080/foo/bar`.
+The property `server.url` is expanded to `http://example.org:8080/foo`.
 
-If an expression cannot be expanded and does not have a default value a `NoSuchElementException` is thrown. In the
-Optional case, returns an empty Optional.
+If an expression cannot be expanded and does not have a default value, a `NoSuchElementException` is thrown. In the
+Optional case, an empty Optional will be returned.
 
-Composed expressions, are limited to the number of recurssion lookups performed. Implementations are encouraged to
-limit the number to `5`, but they can use higher limits if they wish to. When the number of allowed recursion lookups
-exceeds the limit, a `IllegalArgumentException` is thrown.
+The number of recursion lookups is not infinite, but a limited number for composed expressions. Implementations are
+encouraged to limit the number to `5`, but they can use higher limits if they wish to. When the number of allowed
+lookups exceeds the limit, a `IllegalArgumentException` is thrown.
 
-Property expressions applies to lookup methods in `Config`, like `getValue`, `getConfigValue` and `getOptionalValue`.
-Lookup methods in `ConfigSource`, may support property expressions as well, but it is not required by the specification.
+Property expressions applies to all the methods in `Config` that performs resolution of a configuration property,
+including `getValue`, `getValues`, `getConfigValue`, `getValues`, `getOptionalValue`, `getOptionalValues`
+and `getConfigProperties`. The methods `getValue` and `getProperties` in `ConfigSource`, may support property
+expressions as well, but it is not required by the specification.
 
-The escape character is "\".
+Property expressions must also be applied to configuration properties injected via CDI. A default value
+defined via `org.eclipse.microprofile.config.inject.ConfigProperty#defaultValue` is not eligible to be expanded since
+multiple candidates may be available.
+
+If a configuration property value or default value requires the raw value without expansion, the expression may be
+escaped with a backslash ("\", double "\\" for property file based sources). For instance:
+
+[source,properties]
+----
+server.url=\\${server.host}
+server.host=localhost
+----
+
+The value of `server.url` is `${server.host}`.
 
 === Backwards Compatibility
 
 MicroProfile Config implementations MUST provide a way to disable variable evaluation to provide backwards
-compatibility. The property `mp.config.property.expressions` can be set to `false` and disable property expressions. By
-default, property expressions are enabled.
+compatibility. The property `mp.config.property.expressions.enabled` was introduced for this purpose. The value of the
+property determines whether the property expression is enabled or disabled. The value `false` means the property
+expression is disabled, while `true` means enabled.
 
-If property expression expand is not desirable for a specific case, the raw value on a configuration property may be
+If property expression expansion is not desirable for a specific case, the raw value on a configuration property may be
 retrieved by calling `getRawValue()` in `ConfigValue`.
+
+Specific sources may already use a similar or identical syntax to the one described in this specification. To preserve
+this usage, `ConfigSource#getValue()` should perform the expression substitution and then return the resolved value.
+Should such a source return a value with an expression from `ConfigSource#getValue()`, usual expression substitution
+does occur as described by this spec.
--- a/spec/src/main/asciidoc/release_notes.asciidoc
+++ b/spec/src/main/asciidoc/release_notes.asciidoc
@@ -27,9 +27,10 @@ A full list of changes delivered in the 2.0 release can be found at link:https:/
 === Incompatible Changes
 
 - ConfigSource#getPropertyNames is no longer a default method. The implementation of a ConfigSource must implement this method. (link:https://github.com/eclipse/microprofile-config/issues/431[431]
-- Property expansion is a MicroProfile Config 2.0 feature. As such, previous versions of the specification would not
-evaluate expresions. With property expansion available, there is a chance that previous working configuration won't
-work anymore.
+- Previous versions of the specification would not evaluate Property Expressions. As such, previous working
+configuration may behave differently (if the used configuration contains values with Property Expressions
+syntax). Check the link:property-expressions.asciidoc[Property Expressions] section on how to go back to the
+previous behaviour.
 
 === API/SPI Changes
 - Convenience methods have been added to Config allowing for the retrieval of multi-valued properties as lists instead of arrays (link:https://github.com/eclipse/microprofile-config/issues/496[#496])
@@ -43,6 +44,7 @@ work anymore.
 - Provide built-in Converters: OptinalInt, OptionalLong and OptionalDouble (link:https://github.com/eclipse/microprofile-config/issues/513[513])
 - Clarifies that Converters for primitive wrappers apply to primitive types as well (link:https://github.com/eclipse/microprofile-config/issues/520[520])
 - Clarify that nulls cannot be passed in to Converters (link:https://github.com/eclipse/microprofile-config/pull/542[542])
+- Support Property Expressions: This provides a way to set and expand variables in property values (link:https://github.com/eclipse/microprofile-config/issues/118[118])
 
 === Other Changes
 - Update to Jakarta EE8 APIs for MP 4.0 (link:https://github.com/eclipse/microprofile-config/issues/469[469])
--- a/spec/src/main/asciidoc/release_notes.asciidoc
+++ b/spec/src/main/asciidoc/release_notes.asciidoc
@@ -26,7 +26,10 @@ A full list of changes delivered in the 2.0 release can be found at link:https:/
 
 === Incompatible Changes
 
-- ConfigSource#getPropertyNames is no longer a default method. The implementation of a ConfigSource must implement this method. (link:https://github.com/eclipse/microprofile-config/issues/431[431])
+- ConfigSource#getPropertyNames is no longer a default method. The implementation of a ConfigSource must implement this method. (link:https://github.com/eclipse/microprofile-config/issues/431[431]
+- Property expansion is a MicroProfile Config 2.0 feature. As such, previous versions of the specification would not
+evaluate expresions. With property expansion available, there is a chance that previous working configuration won't
+work anymore.
 
 === API/SPI Changes
 - Convenience methods have been added to Config allowing for the retrieval of multi-valued properties as lists instead of arrays (link:https://github.com/eclipse/microprofile-config/issues/496[#496])
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/PropertyExpressionsTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/PropertyExpressionsTest.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright (c) 2016-2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.config.tck;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.testng.annotations.Test;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Set;
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.toList;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertThrows;
+
+public class PropertyExpressionsTest {
+    @Test
+    public void simpleExpression() {
+        Config config = buildConfig("my.prop", "1234", "expression", "${my.prop}");
+
+        assertEquals("1234", config.getValue("expression", String.class));
+    }
+
+    @Test
+    public void multipleExpressions() {
+        Config config = buildConfig("my.prop", "1234", "expression", "${my.prop}${my.prop}");
+
+        assertEquals("12341234", config.getValue("expression", String.class));
+    }
+
+    @Test
+    public void composedExpressions() {
+        Config config = buildConfig("my.prop", "1234", "expression", "${${compose}}", "compose", "my.prop");
+
+        assertEquals("1234", config.getValue("expression", String.class));
+    }
+
+    @Test
+    public void defaultExpression() {
+        Config config = buildConfig("expression", "${my.prop:1234}");
+
+        assertEquals("1234", config.getValue("expression", String.class));
+    }
+
+    @Test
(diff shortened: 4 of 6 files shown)


spec/src/main/asciidoc/property-expansion.asciidoc
```diff
@@ -0,0 +1,74 @@
...
+* `${my.prop${compose}}` - Composed expressions. Inner expressions are resolved first.
+* `${my.prop}${my.prop}` - Multiple expressions.
+
+Consider the following configuration properties file:
+
+[source,properties]
+----
+server.url=http://${server.host:example.org}:${server.port}/${server.endpoint}
+server.port=8080
+server.endpoint=${server.endpoint.path.foo$server.endpoint.path.bar}
```
Emily-Jiang, line 50: $server.endpoint.path.bar} -> ${server.endpoint.path.bar}

Emily-Jiang, line 50: $server.endpoint.path.bar} -> ${server.endpoint.path.bar}

spec/src/main/asciidoc/property-expansion.asciidoc
```diff
@@ -0,0 +1,74 @@
...
+
+The property `server.url` is expanded to `http://example.org:8080/foo/bar`.
+
+If an expression cannot be expanded and does not have a default value a `NoSuchElementException` is thrown. In the
+Optional case, returns an empty Optional.
+
+Property expansion applies to lookup methods in `Config`, like `getValue`, `getConfigValue` and `getOptionalValue`.
+Lookup methods in `ConfigSource`, may support property expansion as well, but it is not required by the specification.
+
+=== Backwards Compatibility
```
Emily-Jiang, line 63: Please this part towards the end under Release Note and then MicroProfile Config 2.0 Release Note.

spec/src/main/asciidoc/property-expansion.asciidoc
```diff
@@ -0,0 +1,74 @@
...
+
+Property expansion is a MicroProfile Config 2.0 feature. As such, previous versions of the specification would not
+evaluate expresions. With property expansion available, there is a chance that previous working configuration won't
+work anymore.
+
+MicroProfile Config implementations MUST provide a way to disable variable evaluation to provide backwards
```
Emily-Jiang, line 69: The following 2 paragraphs needs to stay here. Only the previous paragraph should be moved towards the release note.

spec/src/main/asciidoc/property-expansion.asciidoc
```diff
@@ -0,0 +1,74 @@
...
+compatibility. The property `mp.config.expansion` can be set to `false` and disable property expansion. By default,
```
Emily-Jiang, line 70: This bit info needs to be under Config class level javadoc and also talk about the methods that are affected by variable expansion.

radcortez: If we agree with the current state of the proposal, I'll add it.

Emily-Jiang: By the way, as for the property name, I suggest to rename to `mp.config.variable.expansion` as `config.expansion` sounds like a facility to expand MP Config.

spec/src/main/asciidoc/property-expansion.asciidoc
```diff
@@ -0,0 +1,70 @@
...
+----
+
+When looking up the `server.url` property, the value will be resolved and expanded to `http://example.org/endpoint`.
+All MicroProfile Config rules still apply. The `Config` is able to resolve expressions from different ConfigSources.
+
+Additionally, it is also possible to use the following syntax with for property expansion:
+
+* `${expression:value}` - Provides a default value after the `:` if the expansion doesn't find a value.
+* `${my.prop${compose}}` - Composed expressions. Inner expressions are resolved first.
+* `${my.prop}${my.prop}` - Multiple expressions.
```
ljnelson, line 42: What requirements on implementors are there regarding cycles?

radcortez: You mean recursion level in composed expressions?

spec/src/main/asciidoc/property-expansion.asciidoc
```diff
@@ -0,0 +1,70 @@
...
+server.host=example.org
+----
+
+When looking up the `server.url` property, the value will be resolved and expanded to `http://example.org/endpoint`.
+All MicroProfile Config rules still apply. The `Config` is able to resolve expressions from different ConfigSources.
+
+Additionally, it is also possible to use the following syntax with for property expansion:
+
+* `${expression:value}` - Provides a default value after the `:` if the expansion doesn't find a value.
+* `${my.prop${compose}}` - Composed expressions. Inner expressions are resolved first.
```
Emily-Jiang, line 41: We need to be careful to support the nested variable replacement. How deep do we want to go? 

The implementation must detect the infinite loop and exit with IllegalArgumentException. See the example below.

a=${${b}}
b=${a}
c=${d}
d=${c}

Emily-Jiang: Http spec has this stated:
10.3 Redirection 3xx

   This class of status code indicates that further action needs to be
   taken by the user agent in order to fulfill the request.  The action
   required MAY be carried out by the user agent without interaction
   with the user if and only if the method used in the second request is
   GET or HEAD. **A client SHOULD detect infinite redirection loops, since
   such loops generate network traffic for each redirection.**

      Note: previous versions of this specification recommended a
      maximum of **five** redirections. Content developers should be aware
      that there might be clients that implement such a fixed
      limitation.

Though it is not exactly same, it is quite close by nature. I am wondering whether we could borrow the similar approach.

Emily-Jiang:
Add \ as the escape character for escaping ${..} and : for default set.

radcortez:
> Add \ as the escape character for escaping ${..} and : for default set.

We need to analyze this more carefully, since it may clash with the escape of commas for arrays.

radcortez:
Added TCK's.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+server.endpoint.path.foo=/foo
+server.endpoint.path.bar=/bar
+----
+
+The property `server.url` is expanded to `http://example.org:8080/foo/bar`.
+
+If an expression cannot be expanded and does not have a default value a `NoSuchElementException` is thrown. In the
+Optional case, returns an empty Optional.
+
+Composed expressions, are limited to the number of recurssion lookups performed. Implementations are encouraged to
```
Emily-Jiang, line 60: should be recursion

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+server.url=http://${server.host:example.org}:${server.port}/${server.endpoint}
+server.port=8080
+server.endpoint=${server.endpoint.path.foo${server.endpoint.path.bar}}
+server.endpoint.path.foo=/foo
+server.endpoint.path.bar=/bar
+----
+
+The property `server.url` is expanded to `http://example.org:8080/foo/bar`.
+
+If an expression cannot be expanded and does not have a default value a `NoSuchElementException` is thrown. In the
```
Emily-Jiang, line 57: If an expression cannot be expanded and does not have a default value**,** a `NoSuchElementException` is thrown. miss a comma

api/src/main/java/org/eclipse/microprofile/config/ConfigValue.java
```diff
@@ -43,10 +43,17 @@
     /**
      * The value of the property lookup.
```
Emily-Jiang, line 44: It is better to say:
The value of the property lookup with transformation (expansion, etc).

It will be much clearer.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+* `${my.prop${compose}}` - Composed expressions. Inner expressions are resolved first.
+* `${my.prop}${my.prop}` - Multiple expressions.
+
+Consider the following configuration properties file:
+
+[source,properties]
+----
+server.url=http://${server.host:example.org}:${server.port}/${server.endpoint}
+server.port=8080
+server.endpoint=${server.endpoint.path.foo${server.endpoint.path.bar}}
```
Emily-Jiang, line 50: It should be:
server.endpoint=${server.endpoint.path.foo}${server.endpoint.path.bar}

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+server.endpoint.path.foo=/foo
+server.endpoint.path.bar=/bar
+----
+
+The property `server.url` is expanded to `http://example.org:8080/foo/bar`.
+
+If an expression cannot be expanded and does not have a default value a `NoSuchElementException` is thrown. In the
+Optional case, returns an empty Optional.
```
Emily-Jiang, line 58: => In the Optional case, an empty Optional will be returned.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+Composed expressions, are limited to the number of recurssion lookups performed. Implementations are encouraged to
+limit the number to `5`, but they can use higher limits if they wish to. When the number of allowed recursion lookups
+exceeds the limit, a `IllegalArgumentException` is thrown.
+
+Property expressions applies to lookup methods in `Config`, like `getValue`, `getConfigValue` and `getOptionalValue`.
```
Emily-Jiang, line 64: To make it clearer:
Property expressions applies to the methods `getValue`, `getConfigValue` and `getOptionalValue` in `Config`.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+Lookup methods in `ConfigSource`, may support property expressions as well, but it is not required by the specification.
```
Emily-Jiang, line 65: I would avoid the term Lookup methods as we have not defined anywhere.
How about
The methods `getValue()` in `ConfigSource` may support property expressions as well, but it is not required by the specification.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+The escape character is "\".
+
+=== Backwards Compatibility
+
+MicroProfile Config implementations MUST provide a way to disable variable evaluation to provide backwards
+compatibility. The property `mp.config.property.expressions` can be set to `false` and disable property expressions. By
```
Emily-Jiang, line 72: The property `mp.config.property.expressions` can be set to `false` and disable property expressions.
=>
The property `mp.config.property.expressions` was introduced for this purpose. The value of the property determines whether the property expression is enabled. The value 'false` means the property expression is disabled, while `true` means enabled.

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -142,6 +142,9 @@ A full list of changes delivered in the 2.0 release can be found at link:https:/
 
 === Spec Changes
 
+- Property expansion is a MicroProfile Config 2.0 feature. As such, previous versions of the specification would not
```
Emily-Jiang, line 145: You need to rebase this with the latest release note. You need to list this in Incompatible Changes to clearly spec the Config.getValue change.

tck/src/main/java/org/eclipse/microprofile/config/tck/PropertyExpressionsTest.java
```diff
@@ -0,0 +1,183 @@
+/*
+ * Copyright (c) 2016-2018 Contributors to the Eclipse Foundation
```
Emily-Jiang, line 2: 2016-2020?

tck/src/main/java/org/eclipse/microprofile/config/tck/PropertyExpressionsTest.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.NoSuchElementException;
+import java.util.Set;
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.toList;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertThrows;
+
+public class PropertyExpressionsTest {
+    @Test
```
Emily-Jiang, line 38: This is unit test. How can we test against the runtime to ensure the runtime support this? We need to use Arqullian way to package an app though.

spec/src/main/asciidoc/property-expressions.asciidoc
radcortez: Idea here was to add a sample of a composed expression.

tck/src/main/java/org/eclipse/microprofile/config/tck/PropertyExpressionsTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+    @Test
+    void expressionMissing() {
+        Config config = buildConfig("my.prop", "${expression}", "my.prop.partial", "${expression}partial");
+
+        assertThrows(Exception.class, () -> config.getValue("my.prop", String.class));
+        assertThrows(Exception.class, () -> config.getValue("my.prop.partial", String.class));
+    }
+
+    private static Config buildConfig(String... keyValues) {
```
radcortez, line 154: Programmatic Config.

tck/src/main/java/org/eclipse/microprofile/config/tck/PropertyExpressionsTest.java
radcortez: If you look here:
https://github.com/eclipse/microprofile-config/pull/577#discussion_r463547817

I just create Programatic Config. The Config Provider is available when running the TCK, so it does work. And since we don't need CDI or other features from Arquillian, we don't really need to packaged this in an app.

radcortez: BTW, I'm able to run the test just fine in SR Config TCK.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,79 @@
...
+lookups exceeds the limit, a `IllegalArgumentException` is thrown.
+
+Property expressions applies to all the methods in `Config` that performs resolution of a configuration property,
+including `getValue`, `getValues`, `getConfigValue`, `getValues`, `getOptionalValue`, `getOptionalValues`
+and `getConfigProperties`. The methods `getValue` and `getProperties` in `ConfigSource`, may support property
+expressions as well, but it is not required by the specification.
+
+The escape character is "\".
+
+=== Backwards Compatibility
```
jbee, line 71: I checked and we already use `${...}` syntax for a particular source. We cannot change that syntax as it  originates from other concepts that must stay as they are. 

I think the way to integrate is simply to have this source interpret the syntax with its own mechanism while all other sources use the MP mechanism. From outside of the source it will look like there was no expression used as it is already resolved when the values exists the source.

Maybe there should be something here in the spec saying that special sources might do something else when encountering this syntax but the sources named in the spec will honour it and in general sources try to follow the spec.

radcortez: Doesn't this piece covers it?

> Property expressions applies to all the methods in `Config` that performs resolution of a configuration property, including `getValue`, `getValues`, `getConfigValue`, `getValues`, `getOptionalValue`, `getOptionalValues` and `getConfigProperties`. The methods `getValue` and `getProperties` in `ConfigSource`, may support property expressions as well, but it is not required by the specification.

jbee: To me this is about the spec mechanism being implemented on the source level or not. What I mean is that there is a complete different mechanism in place that could be confused with MP Config expressions because of the same syntax.

spec/src/main/asciidoc/property-expressions.asciidoc
Emily-Jiang: It is incorrect example though. After the expansion, you have ${server.endpoint.path.foobar}. You will get NoSuchElementExists exception.

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -26,7 +26,10 @@ A full list of changes delivered in the 2.0 release can be found at link:https:/
 
 === Incompatible Changes
 
-- ConfigSource#getPropertyNames is no longer a default method. The implementation of a ConfigSource must implement this method. (link:https://github.com/eclipse/microprofile-config/issues/431[431])
+- ConfigSource#getPropertyNames is no longer a default method. The implementation of a ConfigSource must implement this method. (link:https://github.com/eclipse/microprofile-config/issues/431[431]
+- Previous versions of the specification would not evaluate Property Expressions. As such, previous working
+configuration may return behave differently (if the used configuration contains values with Property Expressions
```
Emily-Jiang, line 31: may return -> may

Emily-Jiang: As such, previous working
configuration may return behave differently (if the used configuration contains values with Property Expressions
=>
As such, previous property lookup will behave differently if the value of the property contains Property Expressions.

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -26,7 +26,10 @@ A full list of changes delivered in the 2.0 release can be found at link:https:/
...
+syntax). This new behaviour can be disabled. (link:property-expressions.asciidoc[Property Expressions])
```
Emily-Jiang, line 32: This new behaviour can be disabled. (link:property-expressions.asciidoc[Property Expressions]) =>
Check (link:property-expressions.asciidoc[Property Expressions]) on how to go back to the previous behaviour.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,79 @@
...
+
+MicroProfile Config implementations MUST provide a way to disable variable evaluation to provide backwards
+compatibility. The property `mp.config.property.expressions` was introduced for this purpose. The value of the property
```
Emily-Jiang, line 74: I thought this property name further. It is better to renamed to:
`mp.config.property.expressions.enabled`

spec/src/main/asciidoc/property-expressions.asciidoc
Emily-Jiang: @jbee you are asking for a config source level disablement. I assume the variable expansion is done at last. By then, it is difficult to work out where the config value is from.

jbee: I am not asking for a feature, just that the spec acknowledges that there might be sources where the same syntax has different effect.

spec/src/main/asciidoc/property-expressions.asciidoc
radcortez: Ah yes, you are right! Thanks!

spec/src/main/asciidoc/property-expressions.asciidoc
radcortez: @jbee, please feel free to propose what you think it makes sense.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,79 @@
...
+
+=== Backwards Compatibility
+
+MicroProfile Config implementations MUST provide a way to disable variable evaluation to provide backwards
+compatibility. The property `mp.config.property.expressions.enabled` was introduced for this purpose. The value of the
+property determines whether the property expression is enabled or disabled. The value `false` means the property
+expression is disabled, while `true` means enabled.
+
+If property expression expand is not desirable for a specific case, the raw value on a configuration property may be
+retrieved by calling `getRawValue()` in `ConfigValue`.
```
jbee, line 79: Proposal: Here a new paragraph could read:

> Specific sources may already use a similar or identical syntax to the one described in this standard for a source specific replacement mechanism. To stay backwards compatible such mechanisms take precedence. Usually this means the values returned by  `Source#getValue()` do no longer contain expressions present in the source itself. Should such a source return a value with an expression from `Source#getValue()` usual expression substitution does occur as described by this spec.

spec/src/main/asciidoc/property-expressions.asciidoc
Emily-Jiang: @jbee I understood your scenario for skipping the expansion for some special config sources. Yep, I am also ok for adding some relaxation sentence.

tck/src/main/java/org/eclipse/microprofile/config/tck/PropertyExpressionsTest.java
Emily-Jiang: I took a further look at the tests. You do need to test CDI, which is why I asked.

`@Inject @ConfigProperty(name="url") SomeType t;`
By the way, please also add another test to ensure the defaultValue can contain ${}, which will be expanded as well.

spec/src/main/asciidoc/property-expressions.asciidoc
Emily-Jiang: A slight rephrase @jbee @radcortez 
Specific sources may already use a similar or identical syntax to the one described in this standard for a source specific replacement mechanism. To preserve this usage, Source#getValue() should perform the expression substitution and then return the resolved value. Should such a source return a value with an expression from Source#getValue() usual expression substitution does occur as described by this spec.

tck/src/main/java/org/eclipse/microprofile/config/tck/PropertyExpressionsTest.java
```diff
@@ -0,0 +1,190 @@
...
+        if (keyValues.length % 2 != 0) {
+            throw new IllegalArgumentException("keyValues array must be a multiple of 2");
+        }
+
+        Map<String, String> properties = new HashMap<>();
+        for (int i = 0; i < keyValues.length; i += 2) {
+            properties.put(keyValues[i], keyValues[i + 1]);
+        }
+
+        return ConfigProviderResolver.instance().getBuilder()
```
Emily-Jiang, line 171: I took a further look at the tests. You do need to test CDI, which is why I asked.

@Inject @ConfigProperty(name="url") SomeType t;
By the way, please also add another test to ensure the defaultValue can contain ${}, which will be expanded as well.

Emily-Jiang: As talked in the call with @jbee @radcortez, the spec will state default value will not perform expression substituation.

otaviojava:
@radcortez 

One case that I did not see on the spec.
What happened when the property expression does not exist?
E.g.:
```
host=${KEY}
```
When there is no value to KEY.

IMHO: this value must be skipped and just be mapped when the KEY exists.

radcortez:
> @radcortez
> 
> One case that I did not see on the spec.
> What happened when the property expression does not exist?
> E.g.:
> 
> ```
> host=${KEY}
> ```
> 
> When there is no value to KEY.
> 
> IMHO: this value must be skipped and just be mapped when the KEY exists.

Hey @otaviojava, I believe this section covers it:

> If an expression cannot be expanded and does not have a default value, a `NoSuchElementException` is thrown. In the
 Optional case, an empty Optional will be returned.

What do you think?

radcortez:
In that case, you just use an `Optional`.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+
+Consider the following configuration properties file:
+
+[source,properties]
+----
+server.url=http://${server.host:example.org}:${server.port}/${server.endpoint}
+server.port=8080
+server.endpoint=${server.endpoint.path.${server.endpoint.path.bar}}
+server.endpoint.path.foo=/foo
+server.endpoint.path.bar=foo
```
Emily-Jiang, line 52: The value should be `bar`.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+----
+
+The property `server.url` is expanded to `http://example.org:8080/foo/bar`.
```
Emily-Jiang, line 55: The value should be `http://example.org:8080/foo`.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+including `getValue`, `getValues`, `getConfigValue`, `getValues`, `getOptionalValue`, `getOptionalValues`
+and `getConfigProperties`. The methods `getValue` and `getProperties` in `ConfigSource`, may support property
+expressions as well, but it is not required by the specification.
+
+Property expressions must also be applied to configuration properties injected via CDI. A default value
+defined via `org.eclipse.microprofile.config.inject.ConfigProperty#defaultValue` is not eligible to be expanded since
+multiple candidates may be available.
+
+If a configuration property value requires the raw value without expansion, the expression may be escaped with a
+double backslash ("\\"). For instance:
```
Emily-Jiang, line 74: Why double backslash? I recommend to use a single backslach to be consistent with the comma escape for array values.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+The value of `server.url` is `${server.host}`.
+
+=== Backwards Compatibility
+
+MicroProfile Config implementations MUST provide a way to disable variable evaluation to provide backwards
+compatibility. The property `mp.config.property.expressions.enabled` was introduced for this purpose. The value of the
+property determines whether the property expression is enabled or disabled. The value `false` means the property
+expression is disabled, while `true` means enabled.
+
+If property expression expand is not desirable for a specific case, the raw value on a configuration property may be
```
Emily-Jiang, line 91: expand -> expansion

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+retrieved by calling `getRawValue()` in `ConfigValue`.
+
+Specific sources may already use a similar or identical syntax to the one described in this standard for a source
```
Emily-Jiang, line 94: standard -> specification

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+specific replacement mechanism. To preserve this usage, `ConfigSource#getValue()` should perform the expression
+substitution and then return the resolved value. Should such a source return a value with an expression from
+`ConfigSource#getValue()` usual expression substitution does occur as described by this spec.
```
Emily-Jiang, line 97: usual expression -> , usual expression

spec/src/main/asciidoc/property-expressions.asciidoc
radcortez: Well sure, is just that property file based sources requires a double backslash.

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -26,7 +26,11 @@ A full list of changes delivered in the 2.0 release can be found at link:https:/
 
 === Incompatible Changes
 
-- ConfigSource#getPropertyNames is no longer a default method. The implementation of a ConfigSource must implement this method. (link:https://github.com/eclipse/microprofile-config/issues/431[431])
+- ConfigSource#getPropertyNames is no longer a default method. The implementation of a ConfigSource must implement this method. (link:https://github.com/eclipse/microprofile-config/issues/431[431]
+- Previous versions of the specification would not evaluate Property Expressions. As such, previous working
+configuration may return behave differently (if the used configuration contains values with Property Expressions
```
Emily-Jiang, line 31: my return behave -> may behave

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -26,7 +26,11 @@ A full list of changes delivered in the 2.0 release can be found at link:https:/
...
+syntax). Check the link:property-expressions.asciidoc[Property Expressions] specification on how to go back to the
```
Emily-Jiang, line 32: specification -> section

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -40,6 +44,7 @@ A full list of changes delivered in the 2.0 release can be found at link:https:/
 - Provide built-in Converters: OptinalInt, OptionalLong and OptionalDouble (link:https://github.com/eclipse/microprofile-config/issues/513[513])
 - Clarifies that Converters for primitive wrappers apply to primitive types as well (link:https://github.com/eclipse/microprofile-config/issues/520[520])
 - Clarify that nulls cannot be passed in to Converters (link:https://github.com/eclipse/microprofile-config/pull/542[542])
+- Support Property Expressions. This provides a way to set and expand variables in property values (link:https://github.com/eclipse/microprofile-config/issues/118[118])
```
Emily-Jiang, line 47: Support Property Expressions. -> Support Property Expressions:

Emily-Jiang (approved):
LGTM!

radcortez:
Do we want this for 2.0? or 2.1?

Emily-Jiang:
> Do we want this for 2.0? or 2.1?

2.0. It is in 2.0 milestone.

spec/src/main/asciidoc/property-expressions.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+including `getValue`, `getValues`, `getConfigValue`, `getValues`, `getOptionalValue`, `getOptionalValues`
+and `getConfigProperties`. The methods `getValue` and `getProperties` in `ConfigSource`, may support property
+expressions as well, but it is not required by the specification.
+
+Property expressions must also be applied to configuration properties injected via CDI. A default value
+defined via `org.eclipse.microprofile.config.inject.ConfigProperty#defaultValue` is not eligible to be expanded since
+multiple candidates may be available.
+
+If a configuration property value requires the raw value without expansion, the expression may be escaped with a
+backslash ("\", double "\\" for property file based sources). For instance:
```
Emily-Jiang, line 74: Can you also mention backslash is used to escape the :defaultvalue as well?

Emily-Jiang (approved):
LGTM!

Emily-Jiang:
@radcortez please go ahead to merge in. I would like to create a M2 or RC1 on 9th September.

PR #579, opened 2020-07-02
418 - config profile affection on config source level

Emily-Jiang:

### DIFF
--- a/spec/src/main/asciidoc/configprofile.asciidoc
+++ b/spec/src/main/asciidoc/configprofile.asciidoc
@@ -41,6 +41,7 @@ If the property `mp.config.profile` is specified in multiple config sources, the
 
 === How Config Profile works
 
+==== On Property level
 The configuration property that utilizes the Config Profile is called a "profile-specific" property. A "profile-specific" property name consists of the following sequence: `% <profile name>.<original property name>`.
 
 Conforming implementations are required to search for a configuration source with the highest ordinal (priority) that provides either the property name or the "profile-specific" property name. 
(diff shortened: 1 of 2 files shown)


Emily-Jiang:
@jbee @roberto @OndroMih please review. All comments welcome!

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -75,3 +76,19 @@ The properties `%live.vehicle.name` and `%testing.vehicle.name` are inactive con
 
 If `mp.config.profile` is set to `live`, the property `%live.vehicle.name` is the active property. The `vehicleName` will be `train`. Similarly, `bike` will be the value of `vehicleName`, if the profile is `testing`.
 
+==== On Config Source level
+
+Config Profile also affects the default config source `microprofile-config.properties`. If multiple config sources exist under the `META-INF` folder on the classpath with the name like `microprofile-config-<profile_name>.properties`, 
+the config source matching the active profile name will be loaded while the others are ignored. If a matching config source is not found, the `microprofile-config.properties` will be loaded as the fallback config source.
```
jbee, line 82: Does this lead to a hen-egg problem? As the profile is defined as property which e.g. could be defined in the `microprofile-config.properties` how is this supposed to work? First load the `microprofile-config.properties` when resolving the profile, then if a profile is present drop that source and add the `microprofile-config-<profile>.properties` ?

Emily-Jiang: It is a good question. What you suggested should work.

radcortez: I wonder if we should just load `microprofile-config.properties` always as a fallback. This will allow the implementation, to just merge all `microprofile-config-<profile>.properties` files in a single source and then use the specified behaviour for `%profile`.

Emily-Jiang: I've done the update accordingly and clarified that `mp.config.profile` property will be ignored if it was contained inside `microprofile-config-<profile>.properties`.

Emily-Jiang:
@jbee @radcortez I have added Tcks. Please review. Thanks

Emily-Jiang:
Thanks @jbee for approving this PR! @radcortez will you be able to take a quick look at the TCKs today?

Issue #536, opened 2020-03-06
Adopt Maven Formatter Plugin


radcortez:
We should probably use [Maven Formatter Plugin](https://github.com/revelc/formatter-maven-plugin) for consistent code styling and to avoid reformats on code changes.

PR #584, opened 2020-07-24
Consistent javadoc formatting

radcortez:
There are a lot of changes here, but I didn't modify the content itself.

This work was mostly to make the docs consistent between all sources:

- Removed duplicated empty lines.
- Removed double spaces.
- Wrap at 120 chars.
- Same indentations for li and ol.
- Use paragraphs where required.
- Fixed javadoc warnings.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
@@ -1,5 +1,4 @@
 /*
- *******************************************************************************
  * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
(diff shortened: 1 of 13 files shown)


Emily-Jiang:
@radcortez I think it might be better to change the formatting rules. If we use different rules, we will be run into constant battle of changing.

api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
```diff
@@ -16,35 +15,35 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *******************************************************************************/
+ */
 
 /**
- * <p>CDI Support for Microprofile Config
+ * CDI Support for Microprofile Config
```
Emily-Jiang, line 21: MicroProfile

api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
```diff
@@ -16,35 +15,35 @@
...
+ *
+ * <p>
+ * Microprofile Config also supports injection via a JSR-330 DI container:
```
Emily-Jiang, line 24: MicroProfile

radcortez:
> @radcortez I think it might be better to change the formatting rules. If we use different rules, we will be run into constant battle of changing.

Yes, I did suggest this previously here:
https://github.com/eclipse/microprofile-config/issues/536

Actually let me do a PR with that and how the code would look like with the changes so we can discuss.

PR #589, opened 2020-08-04
Fix #588. Consistent use of prefix.

### LINKED ISSUE #588: Prefix in @ConfigProperties
**Describe the bug**

I think we need to revise the way that we specced the behaviour of prefix in @ConfigProperties. Right now the prefix does not behave like other implementations of configurations (like Spring, Micronaut or Quarkus).

We force the user to add the dot in the end of the configuration prefix. This is counter intuitive, compared to the other platforms and even other APIs with similar behaviours. 

We should fix this and specify that the ending dot is not required in the prefix to perform the lookup.

radcortez:


Emily-Jiang:
@radcortez can you solve the conflicts?

api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperties.java
```diff
@@ -72,7 +72,7 @@
      * Support inline instantiation of the {@link ConfigProperties} qualifier.
      */
     public final static class Literal extends AnnotationLiteral<ConfigProperties> implements ConfigProperties {
-        public static final Literal NOPREFIX = of("");
+        public static final Literal NO_PREFIX = of("");
```
jbee, line 75: Is this a not yet released constant?

jbee (approved):

api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperties.java
Emily-Jiang: correct @jbee

radcortez:
Fixed.

spec/src/main/asciidoc/configexamples.asciidoc
```diff
@@ -238,17 +238,25 @@ details.endpoint -> endpoint -> book
...
+** Use `@ConfigProperty` to provide a default value.
 ** Use `Optional<T>` or `OptionalInt`, `OptionalDouble`, `OptionalLong` as the type.
 
 2. The property value cannot be converted to the specified type
 
+If any of the property cannot be found and there is neither default value nor property is not optional, `java.util.NoSuchElementException` will be thrown.
+In order to avoid this, you can supply a default value when defining the field. Alternatively, you can use `@ConfigProperty` to provide a default value.
+You can also use `Optional<T>` or OptionalInt, OptionalDouble, OptionalLong as the type. If any property values cannot be converted to the specified type, `java.lang.IllegalArgumentException` will be thrown.
+
+When using programmatic lookup of the configuration properties, the configuration properties class does not need to be annotated with
```
Emily-Jiang, line 257: The lines between 257 - 260 should be removed as we don't have the spi any more.

PR #590, opened 2020-08-05
Fix TCK tests.

radcortez:


tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigPropertiesTest.java
```diff
@@ -256,12 +251,11 @@ public void testConfigPropertiesDefaultOnBeanPL() {
         Config config = ConfigProvider.getConfig();
         BeanFour myBeanFour = config.getConfigProperties(BeanFour.class);
         Assert.assertEquals("mycloud.org", myBeanFour.getHost());
-        Assert.assertEquals("9080", myBeanFour.port);
+        Assert.assertEquals(9080, myBeanFour.port);
         Assert.assertFalse(myBeanFour.location.isPresent());
     }
 
     @ConfigProperties(prefix="customer.")
-    @Dependent
```
Emily-Jiang, line 264: I thought the Bean defining annotation are required by Quarkus. In FT, we had to explicitly add the scope annotation even though there is a beans.xml. Is this not the case any more?

radcortez: I guess it depends on how you implement it, but we don't need it.

We just scan for annotated beans with `@ConfigProperties` and create a configurator bean with `@Dependent` scope directly. We don't honor any scope defined in the class itself.

Emily-Jiang: What if the class with `@ConfigProperties` annotation also declares a scope, say `@RequestScoped`, the end user would like the lifecyle to be with `@RequestScoped` not `@Dependent` scoped. I am talking about the bean class with `@ConfigProperties`.

radcortez: I think we should not support it for now. It will become confusing with `@ConfigProperty` which is only Dependent scope.

Emily-Jiang: `@ConfigProperty` has no scope associated with as it is just a qualifier not a stereotype.

Emily-Jiang: same thing for `@ConfigProperties`.

radcortez: Ok, it's fine.

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigPropertiesTest.java
```diff
@@ -279,7 +273,6 @@ public String getName() {
     }
 
     @ConfigProperties
-    @RequestScoped
```
Emily-Jiang, line 282: Can you put this annotation back as per our conversation?

PR #594, opened 2020-09-07
Further update ConfigProperties based on the feedback

Emily-Jiang:

Since the class of `@ConfigProperties` is a CDI bean, it is not correct to provide programmatic look from Config, which has no dependency on CDI and it might be called before CDI is fully up. Therefore, we need to remove the methods added to the Config. For programmatic lookup, use CDI.current().select. 

Besides, `ConfigProperty` is only working on CDI beans. It will be odd to look up the bean annotated with `@ConfigProperties` in a non-CDI way, which might get a different object or null if it is called earlier.


radcortez:
The reason we added these methods was to cover the use case for non CDI environments. `CDI.current().select` will not help if you don't have CDI available.

Also, I think that the separator change should be done in a separate PR.

Emily-Jiang:
@radcortez MP Config spec only support the CDI usage. As you know, `@ConfigProperties` is a CDI qualifier. It is odd to be used for non-CDI case.

radcortez:
Until now, we never had any issues in using MP Config in non CDI environments, using the programmatic API.

Anyway, we can remove the annotation inspection from the programmatic API. I think it is only used to read the prefix if the prefix is not provided.

Emily-Jiang:
@radcortez let me redo this PR to leave separtor issue out and do it in a separate PR based on your feedback.

Emily-Jiang:
@radcortez @jbee please double check.

jbee (approved):
LGTM. 

As we discussed it makes sense to me that the programmatic access to the feature is done via `CDI` as you describe it in the spec and not by an additional method `Config.getConfigProperties` assuming that we always want to demand that the filled been is a CDI bean.

radcortez (commented):
I still think it is useful to provide an API to retrieve the mapping object in non CDI environments. For CDI implementations, `Config.getConfigProperties` may delegate to `CDI.current().select`.

Anyway, implementations may provide that method. I'm neutral on this.

Emily-Jiang:
@radcortez technically you can do it but it is very confusing as the annotation is a CDI quarlifier. Making the class usable as a non CDI bean or the instance not from CDI does not make sense. Since you are neutral on this, I will go ahead to merge this in.

PR #595, opened 2020-09-08
240 - config properties with configurable separator

Emily-Jiang:


Emily-Jiang:
will proceed with #589 . We will revisit this after we have got more feedback.

PR #596, opened 2020-09-08
Move ConfigValueBean into it's own class

tevans78:
Signed-off-by: tevans <tevans@uk.ibm.com>

@Emily-Jiang Since Arquillian adds the test class to the archive by default, the nested bean class appears twice and leads to an ambiguous bean error.


genie-microprofile:

Emily-Jiang (commented):
@tevans78 can you try to remove the line 49 of addClasses to see whether it fixes your problem? I think it is much simpler.

tevans78:
Will try to address this in another way

PR #597, opened 2020-09-09
Extract configuration names to constants.

radcortez:


api/src/main/java/org/eclipse/microprofile/config/Names.java
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.config;
+
+/**
+ * Constants for the names of the MP-config properties that implementations must support.
+ */
+public interface Names {
```
Emily-Jiang, line 24: I would rename this to Constants. Alternatively, put the constants on the Config interface.

radcortez: Well, I was following what other projects are using to make it consistent. For instance: 
https://github.com/eclipse/microprofile-jwt-auth/blob/e0157e5a4c2a9d01b02176e08995f88eacf011f4/api/src/main/java/org/eclipse/microprofile/jwt/config/Names.java

jbee (approved):

Emily-Jiang (approved):
LGTM!

PR #598, opened 2020-09-11
Empty value handling - rebase from #541

### LINKED ISSUE #397: Need a way to specify empty list/array in microprofile-config.properties

### LINKED ISSUE #446: Consistent empty/missing behavior

### LINKED ISSUE #531: Enable explicit clearing of values

### LINKED ISSUE #532: Consistent empty/missing value handling (use case)

Emily-Jiang:
Fixes #446, Fixes #531, Fixes #532, Fixes #397.


Emily-Jiang:
I will update the tests and accommodate the comments in my next commit.

Emily-Jiang:

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -99,15 +99,22 @@
...
-     * @throws java.util.NoSuchElementException if the property isn't present in the configuration
+     * @param <T>
+     *             The property type
+     * @param propertyName
+     *             The configuration property name
+     * @param propertyType
+     *             The type into which the resolved property value should get converted
+     * @return the resolved property value as an instance of the requested type (not {@code null})
+     * @throws IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws NoSuchElementException if the property name or the value of the property does not exist
```
jbee, line 115: I think the phrasing _if the property name or the value of the property does not exist_ could be more clear. Something like: _if the property is not defined or is defined as an empty/blank string_

I assume blank is considered empty?

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -114,13 +114,29 @@
  * All <em>built in</em> Converters have a priority value of {@code 1}. <em>Implicit</em> converters are only created
  * when no other converter was found; therefore, they do not have a priority.
  *
+ * <h3 id="empty">Empty values</h3>
+ *
+ * For all converters, the empty string {@code ""} <em>must</em> be considered an empty value. Some converters <em>may</em>
+ * consider other values to be empty as well.
+ * <p>
+ * When a conversion results in an empty value, most converters should normally return {@code null}.  All of the built-in
```
jbee, line 122: I find this confusing an unnecessary. Does that mean even if source lookup ends up with `null` (not present) a converter could turn this into some value (present)? Effectively a programmed default value?

I suggest converters must not be burdened with handling `null` or empty/blank and `null` after source lookup should always result in the _not present_ result of the `Config` API. Should a user resolve a converter directly and pass an empty string or `null` to it it would be undefined behaviour. Or if we want to support this as a feature I feel the converter should behave consistent with the `Config` API, not with the `Source` API and throw an exception.

I understand that for some types there does exist a value that could represent an empty value or null but as it is impossible to formulate such a value using a source I don't see why converter would need to deal with them.

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -26,11 +26,15 @@ A full list of changes delivered in the 2.0 release can be found at link:https:/
 
 === Incompatible Changes
 
-- ConfigSource#getPropertyNames is no longer a default method. The implementation of a ConfigSource must implement this method. (link:https://github.com/eclipse/microprofile-config/issues/431[431]
+- ConfigSource#getPropertyNames is no longer a default method. The implementation of a ConfigSource must implement this method. (link:https://github.com/eclipse/microprofile-config/issues/431[431])
 - Previous versions of the specification would not evaluate Property Expressions. As such, previous working
 configuration may behave differently (if the used configuration contains values with Property Expressions
 syntax). Check the link:property-expressions.asciidoc[Property Expressions] section on how to go back to the
 previous behaviour.
+- Empty value or other special characters is no longer a valid config value for a particular return type. Refer to the earlier section of this spec for more details.
```
jbee, line 34:
```suggestion
- Empty value or other special characters are no longer a valid config value for a particular return type. Refer to the earlier section of this spec for more details.
```

spec/src/main/asciidoc/configsources.asciidoc
```diff
@@ -34,7 +34,7 @@ A higher `ordinal` means that the values taken from this `ConfigSource` will ove
 This allows a configuration to be customized from outside a binary, assuming that external `ConfigSource` s have higher `ordinal` values than the ones whose values originate within the release binaries.
 
 It can also be used to implement a drop-in configuration approach.
-Simply create a jar containing a `ConfigSource` with a higher ordinal and override configuration values in it.
+Simply create a jar containing a `ConfigSource` with a higher ordinal and override configuration values in it. Specifying an empty string as the value erases the property.
```
jbee, line 37:
```suggestion
Simply create a jar containing a `ConfigSource` with a higher ordinal and override configuration values in it. Specifying an empty string as the value effectively erases the property.
```

tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValue/EmptyValuesTestProgrammaticLookup.java
```diff
@@ -0,0 +1,229 @@
...
+                .addAsLibrary(jar)
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    @Inject
+    private Config config;
+
+    @Test(expectedExceptions = NoSuchElementException.class)
+    public void testMissingStringGetValueArray() {
+        String[] values = config.getValue("missing.string", String[].class);
```
jbee, line 68: In such cases I like to create a helper, like `void assertConfigurationNotPresent(String property);` which would use all the different ways to check that so that you get better readability and more coverage as you will cover the matrix with different ways properties are defined and different ways they are resolved.

jbee (commented):
Dropped some thoughts. The only crucial one to me is the question of how converters behave when they are given `null` or empty/blank string values. 

I'd think 

* either a _not present_ source lookup is never passed to the converter level and doing so manually is considered undefined behaviour to lift the burden of implementing these corner cases 
* or any result from source lookup is passed to the converters but any empty value (not present) results in the same behaviour as the `Config` API would show, throwing an exception of some type.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -99,15 +99,22 @@
...
      */
-    <T> T getValue(String propertyName, Class<T> propertyType);
+    <T> T getValue(String propertyName, Class<T> propertyType) throws IllegalArgumentException, NoSuchElementException;
```
radcortez, line 117: I think we shouldn't declare the exceptions in the method signature. We were not doing it before and they are unchecked.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -162,9 +169,10 @@
      * @param propertyName The configuration property name
      * @param propertyType The type into which the resolved property value should be converted
      * @return The resolved property value as an {@code Optional} wrapping the requested type
-     * @throws java.lang.IllegalArgumentException if the property cannot be converted to the specified type
+     *
+     * @throws IllegalArgumentException if the property cannot be converted to the specified type
      */
-    <T> Optional<T> getOptionalValue(String propertyName, Class<T> propertyType);
+    <T> Optional<T> getOptionalValue(String propertyName, Class<T> propertyType) throws IllegalArgumentException;
```
radcortez, line 175: I think we shouldn't declare the exceptions in the method signature. We were not doing it before and they are unchecked.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
radcortez: > I find this confusing an unnecessary. Does that mean even if source lookup ends up with null (not present) a converter could turn this into some value (present)? Effectively a programmed default value?

I don't think so. The `org.eclipse.microprofile.config.spi.Converter#convert` states that callers should not pass `null` or you may get a `NullPointerException`.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -114,13 +114,29 @@
...
+ * converters except the primitive ones <em>must</em> return {@code null} for an empty value, <em>except</em> for the built-in converters for
```
radcortez, line 123: Except used twice. Seems confusing.

radcortez (changes requested):

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
jbee: In which case I think for consistency empty/blank should be no different. And as it would be a pain to tax each converter to check for blank strings I suggest to assume they never get passed such as this is already an identified _not present_ result. If user still pass such values manual behaviour would be undefined.

api/src/main/java/org/eclipse/microprofile/config/Config.java
Emily-Jiang: I would just say empty string as blank string term is not commonly used.

api/src/main/java/org/eclipse/microprofile/config/Config.java
Emily-Jiang: ok, I agree with you. I think I inherited from the previous PR.

api/src/main/java/org/eclipse/microprofile/config/Config.java
Emily-Jiang: removed.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
Emily-Jiang: I am ok to remove this as it is implementation details based on what @jbee said.  @radcortez ok?

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
Emily-Jiang: updated

spec/src/main/asciidoc/configsources.asciidoc
Emily-Jiang: done

spec/src/main/asciidoc/release_notes.asciidoc
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValue/EmptyValuesTestProgrammaticLookup.java
Emily-Jiang: good idea. done

jbee:
@Emily-Jiang I believe you have not yet pushed what you commented as _done_

tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValue/EmptyValuesTest.java
```diff
@@ -55,25 +55,7 @@ public static Archive deployment() {
...
-        String stringValue = config.getValue(EMPTY_PROPERTY, String.class);
-        assertEquals(stringValue, "");
-        System.clearProperty(EMPTY_PROPERTY);
-    }
-
     @Test
-    public void testEmptyStringPropertyFromConfigFile() {
-        String stringValue = config.getValue(PROP_FILE_EMPTY_PROPERTY, String.class);
-        assertEquals(stringValue, "");
+    public void test() {
```
jbee, line 59: Is there nothing to test?

jbee (approved):
I think everything is fine. Just confused about the empty `test()` method.

tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValue/EmptyValuesTest.java
Emily-Jiang: This is just a negative test case and the deployment will fail. The test there is for reporting true. This is the pattern we use for deployment failures (e.g. [this](https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/broken/MissingConverterOnInstanceInjectionTest.java) one)

Emily-Jiang:
@radcortez I think I addressed all of your comments as well. Please take another look at it and I would like to merge in tomorrow.

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
radcortez: Sure.

radcortez (changes requested):

tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValue/EmptyValuesTest.java
```diff
@@ -32,15 +33,14 @@
...
-import static org.testng.Assert.assertEquals;
-
 public class EmptyValuesTest extends Arquillian {
 
     private static final String EMPTY_PROPERTY = "my.empty.property";
     private static final String PROP_FILE_EMPTY_PROPERTY = "my.empty.property.in.config.file";
     public static final StringAsset EMPTY_STRING_ASSET = new StringAsset(PROP_FILE_EMPTY_PROPERTY + "=");
 
     @Deployment
+    @ShouldThrowException(DeploymentException.class)
```
radcortez, line 43: I believe this won't throw the expected exception because `EmptyValuesBean#stringValue` sets a default in the annotation.

radcortez (changes requested):

tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValue/EmptyValuesTest.java
Emily-Jiang: The default is an empty value though so this is to verify this empty value won't make the property exist.

tck/src/main/java/org/eclipse/microprofile/config/tck/EmptyValue/EmptyValuesTest.java
```diff
@@ -17,10 +17,11 @@
  * limitations under the License.
  *
  */
-package org.eclipse.microprofile.config.tck;
+package org.eclipse.microprofile.config.tck.EmptyValue;
```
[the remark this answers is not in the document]
radcortez: The package name was changed, but the file is still in the wrong directory.

tck/src/main/java/org/eclipse/microprofile/config/tck/emptyvalue/EmptyValuesTestProgrammaticLookup.java
```diff
@@ -0,0 +1,233 @@
+/*
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
```
tevans78, line 2: wrong date?

tck/src/main/java/org/eclipse/microprofile/config/tck/emptyvalue/EmptyValuesTestProgrammaticLookup.java
```diff
@@ -0,0 +1,233 @@
...
+import org.testng.annotations.Test;
+
+import javax.inject.Inject;
+
+public class EmptyValuesTestProgrammaticLookup extends Arquillian {
+
+    @Deployment
+    public static Archive deployment() {
+        JavaArchive jar = ShrinkWrap.create(JavaArchive.class, "emptyValues.jar")
+                .addClasses(EmptyValuesTest.class, EmptyValuesBean.class)
```
radcortez, line 44: These should not be included here.

radcortez (changes requested):

radcortez (approved):

Issue #573, opened 2020-06-17
Clarification about default ConfigSources ordinal


NicoNes:
Hi,

The spec states that:

> a special property config_ordinal can be set within any built-in ConfigSource implementation

Does it means that default config sources MUST honor this special property when present otherwise use the respective default ?

Thanks

radcortez:
Hi @NicoNes 

From my understanding yes. We should not treat default sources differently, or we will run into consistency issues :)

NicoNes:
Hey @radcortez 

Thanks for the anwser. I agree with you. 

Actually I was asking for this clarification because SmallRye config implementation you seems to contribute to does not honor this this spec requirement:
- `SysPropConfigSource` ordinal is always 400 
- `EnvConfigSource` is always 300.
- `PropertiesConfigSource` is OK and behave as expected.

It's maybe the wrong place to talk about it, let me know if I'm right or not and if you want me to open and issue on SmallRye config project.

Thanks

radcortez:
Ahaha, I was actually wondering if SmallRye was doing that when I've answered. I guess there is no TCK for this, so we probably need to add it as well.

I'll have a look to see if I can come up with a fix.

Emily-Jiang:
The spec allows end users to set the ordinal for the default config sources. It suggests we should add TCKs to ensure this behaviour.

Joseph-Cass:
I've made a SmallRye issue for this: https://github.com/smallrye/smallrye-config/issues/401

PR #605, opened 2020-09-23
573: clarify default config source ordinal

Emily-Jiang:
Fixes: #573

### DIFF
--- a/tck/running_the_tck.asciidoc
+++ b/tck/running_the_tck.asciidoc
@@ -112,7 +112,7 @@ The following environment variable and their values must be present in the test 
 * `MY_STRING_PROPERTY` with the value `woohoo`
 * `MP_TCK_ENV_DUMMY` with the value `dummy`
 * `config_ordinal` with `45`
-* `customer_name` with Bob`
+* `customer_name` with `Bob`
 
 The following system properties and their values must be present in the test runner environment:
 
@@ -160,8 +160,8 @@ If you use Apache Maven then the tests are run via the `maven-surefire-plugin`
                     <MY_BOOLEAN_PROPERTY>true</MY_BOOLEAN_PROPERTY>
                     <my_string_property>haha</my_string_property>
                     <MY_STRING_PROPERTY>woohoo</MY_STRING_PROPERTY>
-                    <CONFIG_ORDINAL>45</CONFIG_ORDINAL>
-                    <CUSTOMER_NAME>Bob</CUSTOMER_NAME>
+                    <config_ordinal>45</config_ordinal>
+                    <customer_name>Bob</customer_name>
                 </environmentVariables>
                 <systemProperties>
                     <mp.tck.prop.dummy>dummy</mp.tck.prop.dummy>
--- a/tck/running_the_tck.asciidoc
+++ b/tck/running_the_tck.asciidoc
@@ -111,6 +111,8 @@ The following environment variable and their values must be present in the test 
 * `my_string_property` with the value `haha`
 * `MY_STRING_PROPERTY` with the value `woohoo`
 * `MP_TCK_ENV_DUMMY` with the value `dummy`
+* `config_ordinal` with `45`
+* `customer_name` with Bob`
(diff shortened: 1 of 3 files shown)


radcortez:
We need to update the TCK running instructions to add the additional properties.

radcortez (changes requested):
We need to update the TCK running instructions to add the additional properties.

Emily-Jiang:
Thanks @radcortez ! I realised that this morning as well. I also added it to the release note. Please take a look at it! @radcortez @jbee

tck/running_the_tck.asciidoc
```diff
@@ -111,6 +111,8 @@ The following environment variable and their values must be present in the test
 * `my_string_property` with the value `haha`
 * `MY_STRING_PROPERTY` with the value `woohoo`
 * `MP_TCK_ENV_DUMMY` with the value `dummy`
+* `config_ordinal` with `45`
+* `customer_name` with Bob`
```
radcortez, line 115: Add these into the surefire configuration as well.

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/DefaultConfigSourceOrdinalTest.java
```diff
@@ -0,0 +1,95 @@
...
+       //check whether the environment variables were populated by the executor correctly
+
+        if (!"45".equals(System.getenv("config_ordinal"))) {
+         Assert.fail("Before running this test, the environment variable \"config_ordinal\" must be set with the value of 45");
+        }
+        if (!"Bob".equals(System.getenv("customer_name"))) {
+         Assert.fail("Before running this test, the environment variable \"customer_name\" must be set with the value of Bob");
+        }
+        System.setProperty("customer.hobby", "Tennis");
+        System.setProperty("config_ordinal", "120");
```
radcortez, line 76: These should be part of the surefire configuration and not set here.

Since ordinal is very specific property, when we get here, the `Config` object may already be built and in that case the `config_ordinal` was already read to define the source order and there is no effect here.

radcortez (changes requested):

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/DefaultConfigSourceOrdinalTest.java
Emily-Jiang: Good point. done

tck/running_the_tck.asciidoc
Emily-Jiang: done

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/DefaultConfigSourceOrdinalTest.java
```diff
@@ -0,0 +1,95 @@
...
+            .create(WebArchive.class, "DefaultConfigSourceOrdinalTest.war")
+            .addAsLibrary(testJar);
+        return war;
+    }
+
+    @BeforeClass
+    public void checkSetup() {
+       //check whether the environment variables were populated by the executor correctly
+
+        if (!"45".equals(System.getenv("config_ordinal"))) {
```
radcortez, line 69: These checks don't match the properties set in the TCK instructions (upper case). In these case, since we are doing a direct check to `System.env` the name rules don't apply unfortunately.

radcortez (changes requested):

tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/DefaultConfigSourceOrdinalTest.java
Emily-Jiang: I got the info wrong in the TCK instructions. I've corrected it now.

radcortez (approved):

PR #611, opened 2020-09-30
608 - add tck to honour null returned by converter

### LINKED ISSUE #608: Clarify when defaultValue is used on ConfigProperty
**Clarify when defaultValue is used**
`@ConfigProperty(name="my.property", defaultValue="123") int myProp;`
In the [javadoc](https://github.com/eclipse/microprofile-config/blob/master/api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperty.java#L142) of `ConfigProperty`, it has the following:
The default value if the configured property value does not exist.

What if in a config source, the following config property is defined.
`my.property=abc`.

Normally you will get IllegalArgumentException if the converter throws that. However, what if a converter did not throw an exception, it returns `null`. Will `defaultValue` or `null` be assigned to the variable?

- Option 1: defaultValue is not used and `null` -> myProp. Add a tck to verify this.
Pros: honour the converter. In this way, converter has an option to assign `null`.

- Option 2: defaultValue is used. Javadoc needs to be updated to say: if converter return `null`, this value will be used.
Pros: non null is returned.
Cons: behaviour change. Not possible to return a null. Could be confusing if a property is defined with a value and the defaultValue is still used.

Thoughts?

Emily-Jiang:

Fixes #608

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ConvertedNullValueTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ConvertedNullValueTest.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package org.eclipse.microprofile.config.tck;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.config.spi.Converter;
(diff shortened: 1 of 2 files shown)


Emily-Jiang:

tck/src/main/java/org/eclipse/microprofile/config/tck/ConvertedNullValueTest.java
```diff
@@ -0,0 +1,90 @@
...
+        WebArchive war = ShrinkWrap
+                .create(WebArchive.class, "ConvertedNullValueTest.war")
+                .addAsLibrary(testJar);
+        return war;
+    }
+
+    @Test
+    public void testDefaultValue() {
+        // This should return null as the converter returns null if the property does not contain :
+        Assert.assertNull(myBean.getPizza());
```
radcortez, line 71: Hum, shouldn't this fall into the empty rules here: https://github.com/eclipse/microprofile-config/issues/446?

radcortez: And in that case, throw an Exception?

Emily-Jiang: not really. The empty rull is to throw exception when a `null` passed to converter. This one is different: `cheese` was passed to converter. Converter decides to return null after parsing the input.

jbee: Happy to see you confused. This is why I think allowing converters return `null` is no good idea as it is very confusing.

radcortez: In my opinion, empty rules must apply to conversion result as well, or we fall again into the same issue where we are going to see inconsistent results when looking up configuration values.

Emily-Jiang: I did say this was an experiment :). In this case, I will do a separate PR for Option 2 to see whether it is any better.

PR #615, opened 2020-10-06
Release the config which was created in PropertyExpressionsTest

### LINKED ISSUE #614: PropertyExpressionsTest.tearDown issues
**Describe the bug**

`PropertyExpressionsTest.tearDown` tries to release the config for the current context classloader. I have two problems with this:
1. Shouldn't it be relasing the config which was built by the tests, rather than getting the current config and releasing that?
2. When arquillian runs in a managed or remote container, `@AfterMethod` methods are run both on the client and on the server. The config implementation may not be present on the client so `ConfigProviderResolver.instance()` may fail

Azquelt:
The tearDown method should release the config which was created by the
test, rather than trying to release the current config for the thread
context classloader.

This change has the side effect of not trying to call
ConfigProviderResolver.instance() from the test process where it may
fail.

Fixes #614


Emily-Jiang (approved):

radcortez:
I guess this was a side-effect of this test not being an Arquillian test and later converted to become Arquillian.

radcortez (approved):

PR #616, opened 2020-10-06
Avoid duplicate beans in deployments

### LINKED ISSUE #613: TCK tests which define beans as inner classes within test class cause ambiguous bean errors
**Describe the bug**

The current tests use beans defined as inner classes within the test class:
```
CDIPropertyExpressionsTest
ConfigPropertiesTest
ConfigProviderTest
ConfigValueTest
ConfigPropertyFileProfileTest
DevConfigProfileTest
InvalidConfigProfileTest
ProdProfileTest
TestConfigProfileTest
TestCustomConfigProfile
```

In each of these tests, the test class and its inner classes are deployed in a .jar with a beans.xml, which is then included in a .war as a library.

If you're using a non-embedded container, Arquillian then adds the test class and its inner classes to the .war.

When this final archive is deployed, on Open Liberty we get an ambiguous beans error because the test beans are deployed twice (in the .war and in the .jar) and we find both classes when we scan for beans.

When I brought this up in my team, @Emily-Jiang pointed out `CDIPlainInjectionTest` which does the same thing and has been in the TCK for several versions. This test does not fail because it does not package the application as a .jar within a .war and so the test class and its test beans are only deployed once.

I propose fixing this issue by changing the deployment for the other tests to just use a .war rather than a .jar within a .war.

Azquelt:
Where a test uses beans which are inner classes of the test class,
use a single .war deployment rather than a jar within a war.

This prevents the test beans from being included twice (once in the .jar
and again in the .war).

Fixes #613

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPropertyExpressionsTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPropertyExpressionsTest.java
@@ -35,23 +35,17 @@ import org.jboss.arquillian.testng.Arquillian;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
-import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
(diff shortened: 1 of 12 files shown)


Emily-Jiang (approved):

radcortez:
Great! I wanted to do this for a while! Thanks!

radcortez (approved):

PR #624, opened 2020-10-09
Fixes for @ConfigProperties

radcortez:


api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperties.java
```diff
@@ -65,7 +64,6 @@
      *
      * @return the configuration property prefix
      */
-    @Nonbinding
```
Emily-Jiang, line 68: Can you explain why you need to delete this? Once you delete, you will force impl to create different beans with different prefix.

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigPropertiesTest.java
```diff
@@ -191,6 +192,7 @@ public String getHost() {
         }
     }
 
+    @Dependent
```
Emily-Jiang, line 195: Just want to add a note to clarify this.. Sincer there is a beans.xml, all classes are beans. I guess this is for Quarkus to work since it relies on Bean Defining annotations to optimistation. I thought I put there in my original PR but was asked to be removed. It is ok to specify this.

api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperties.java
radcortez: I think not.

The issue is that with @Nonbinding, the prefix value is not taken into account when matching Qualifiers, so if you have:

```java
@ConfigProperties(prefix="foo")
SomeProperties propertiesFoo;

@ConfigProperties
SomeProperties properties;

@ConfigProperties
class SomeProperties {
}
```

You would create 2 instances beans for `SomeProperties`, one using the `@ConfigProperties` qualifier and the other 
the `@ConfigProperties(prefix="foo")` qualifier. In this situation you get an ambiguous resolution.

The prefix does identify the config namespace we want to use associated with the bean. It seems to me that the `prefix` should be binded to the qualifier as well. What do you think?

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigPropertiesTest.java
radcortez: We removed it because BeanThree was only being used for the programmatic API, that we later removed :)

api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperties.java
Emily-Jiang: This depends on the implementation. If you ask for adding `@NonBinding`, you force implementations to provide multiple beans for each prefix used in the code. If you don't add `@NonBinding`, the implementation can just provide one bean and give out different instance depending on the prefix. I think it is much more flexible. Besides, the name attribute is also `NonBinding` in the `ConfigProperty`.

tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigPropertiesTest.java
Emily-Jiang: ah, ok. I might have remembered wrong.

api/src/main/java/org/eclipse/microprofile/config/inject/ConfigProperties.java
radcortez: I think it makes some sense.

It the case of `ConfigProperty` we are talking about a potentially high number of combinations, so it doesn't make sense to provide a bean for each.

In the case of `ConfigProperties`, the number will be much lower, plus it does simplify the implementation and possible extensions, because consumers have to check the injection point prefix, qualifier prefix and finally annotated type prefix when providing the new instance to get the right configuration.

Emily-Jiang: How do you create CDI beans? Do you base on the config properties? Do you add beans based on the injection points?  It will be helpful to get some insights from other implementations as well. @jbee thoughts?

radcortez: I do both. I collect the annotated classes first and provide beans for them. Then I also look into the injection points and provide beans for them.

And completely unrelated, we may want to revisit ao we do injection of a `ConfigProperties` class. Right now, we only need to use `@Inject`, unless we want to override the prefix and it is fine.

On the other hand, after trying it for a while, I feel that adding a `@ConfigProperties` all the time at the injection points, makes the code more clear, in the sense that I can tell straight away what type of class that is. Also, this made me question if we should override the injection point with an empty prefix. Right now, adding a simple `@ConfigProperties` will always override the prefix to use an empty one.

radcortez: I'll open a separate issue so we can discuss a few things. Right now, I think we may have too many options.

radcortez:
I've changed this PR to reflect the discussion in #629.

Emily-Jiang (changes requested):
Hi @radcortez , 
After you left the call, @jbee and I discussed this further. It is fine to support what you proposed. There is one more use case about how to override prefix.
`@Inject @ConfigProperties(prefix="") Details serverDetails`. 
In the current form, we cannot defferentiate whether the value was set to "" or the default value. In order to differentiate this, the prefix needs to use the similar form as `ConfigProperty.defaultValue` -> UNCONFIGURED_VALUE. Please update your PR to reflect that.

spec/src/main/asciidoc/configexamples.asciidoc
```diff
@@ -466,18 +471,6 @@ clientDetails.port -> client.port -> 9081
 clientDetails.endpoint -> client.endpoint -> shelf
 clientDetails.getLocation() -> client.old.location -> Dublin
 ```
-If `@ConfigProperties` has no associated prefix, it means the prefix is empty. e.g.
```
Emily-Jiang, line 469: can you add the condition when prefix is empty, e.g. `@ConfigProperties(prefix="")`

radcortez:
I've did the changes.

I'm wondering if we should align with `@ConfigProperty` behaviour and when you don't set a `prefix` in `@ConfigurationProperties` in the bean class, we default to the FQN as a prefix?

spec/src/main/asciidoc/configexamples.asciidoc
```diff
@@ -429,57 +429,97 @@ public class Details {
...
 clientDetails.host -> client.host -> myHost
 clientDetails.port -> client.port -> 9081
 clientDetails.endpoint -> client.endpoint -> shelf
 clientDetails.getLocation() -> client.old.location -> Dublin
-```
-If `@ConfigProperties` has no associated prefix, it means the prefix is empty. e.g.
 ----
-@Inject @ConfigProperties Details details; //prefix is empty
+
+If `@ConfigProperties` has no associated prefix at the injection point, it defaults to the prefix set inthe `Details`
```
Emily-Jiang, line 481: inthe -> in the

spec/src/main/asciidoc/configexamples.asciidoc
```diff
@@ -429,57 +429,97 @@ public class Details {
...
+----
 serverDetails.host -> server.host -> localhost
 serverDetails.port -> server.port -> 9080
 serverDetails.endpoint -> server.endpoint -> query
 serverDetails.getLocation() -> server.old.location -> London
-```
-* Specify `@ConfigProperties` when injecting the bean annotated with `@ConfigProperties`
+----
+
+Specify `@ConfigProperties` when injecting the bean annotated with `@ConfigProperties`
```
Emily-Jiang, line 460: I suggest to update to:
Specify `prefix` on the annotation `@ConfigProperties` when injecting the bean

Emily-Jiang (approved):

Issue #627, opened 2020-10-13
NullConvertersTest is not set up to run inside a container


Emily-Jiang:
NullConvertersTest has a couple of issues:
1. It uses junit while the other MP Config uses testng. We need to ensure just use one framework.
2. The setup is not set up to run inside a container.

All MP Config TCK tests need to be run "server-side", inside the runtime container. If they are not then there is no guarantee that a ConfigProvider is available. This test needs to extend the Arquillian framework and be deployed to the runtime as a test application (following the same pattern as all of the other TCK tests). similar to #606

PR #628, opened 2020-10-13
627 fix deployment

Emily-Jiang:
Fixes: #627


radcortez:
It's failing the CI.

I've added a check so the CI fails if the code is not formatted :)

Just run the build locally and it will do it for you automatically.

radcortez (approved):

Azquelt (approved):

PR #635, opened 2020-10-19
Fix server.url example in Property Expressions

### LINKED ISSUE #634: Property Expressions spec mistake
**Description**

In the Property Expressions part of the spec, I believe the `server.url` would be expanded to `http://example.org:8080//foo`, not `http://example.org:8080/foo`: https://download.eclipse.org/microprofile/microprofile-config-2.0-RC1/microprofile-config-spec.html#property-expressions

Joseph-Cass:
Fixes #634


genie-microprofile:

radcortez (approved):

PR #643, opened 2020-10-29
633 - fix backslack comma behaviour

Emily-Jiang:

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/emptyvalue/EmptyValuesTestProgrammaticLookup.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/emptyvalue/EmptyValuesTestProgrammaticLookup.java
@@ -123,7 +123,7 @@ public class EmptyValuesTestProgrammaticLookup extends Arquillian {
     }
 
     @Test
-    public void testBackslashCommaStringGetOptionalValueAsArrayorList() {
+    public void testBackslashCommaStringGetOptionalValueAsArrayOrList() {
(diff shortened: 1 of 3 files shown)


Emily-Jiang:

tck/src/main/java/org/eclipse/microprofile/config/tck/emptyvalue/EmptyValuesTestProgrammaticLookup.java
```diff
@@ -120,6 +122,12 @@ public void testBackslashCommaStringGetValue() {
         Assert.assertEquals(value, "\\,");
     }
 
+    @Test
+    public void testBackslashCommaStringGetOptionalValueAsArrayorList() {
```
jbee, line 126:
```suggestion
    public void testBackslashCommaStringGetOptionalValueAsArrayOrList() {
```

jbee (approved):
Just a typo.

Shouldn't there also be an entry in the changelog or was this added after last release?

radcortez (approved):

Emily-Jiang:
> Just a typo.
> 
> Shouldn't there also be an entry in the changelog or was this added after last release?

There was an entry on the release note.

radcortez:
I think I've messed up when I tried this PR and didn't notice that we need to remove this piece:

https://github.com/eclipse/microprofile-config/blob/8d3ab2c6c41e74108754e7a6d713cb414e3aa6f0/tck/src/main/java/org/eclipse/microprofile/config/tck/emptyvalue/EmptyValuesTestProgrammaticLookup.java#L134

This is testing that `getOptionalValue` and `getOptionalValues` are not present if the value is a `\,` but with the updated rules, they now return `,`.

PR #646, opened 2020-11-03
Move bean out of test class in null injection test

### LINKED ISSUE #617: TCK tests which expect deployment failures must not use have beans as inner classes of the test class
**Describe the bug**

This is a subtly different problem to #613 

If a test's deployment method is annotated with `@ShouldThrowException`, then in a managed or remote container, arquillian won't add all the arquillian support classes to the jar, including the `Arquillian` class which the test class extends.

This means that the test class and all of its inner classes failing to load since they depend on the `Arquillian` class.

At the moment, it looks like the only affected test is `ConfigPropertiesMissingPropertyInjectionTest`

On OpenLiberty, I get this warning in the logs when running this test:
```
[06/10/2020, 15:23:41:641 BST] 0000003e org.jboss.weld.Bootstrap                                     I WELD-000119: Not generating any bean definitions from org.eclipse.microprofile.config.tck.broken.ConfigPropertiesMissingPropertyInjectionTest$BeanOne because of underlying class loading error: Type org.jboss.arquillian.testng.Arquillian not found.  If this is unexpected, enable DEBUG logging to see the full error.
```

Azquelt:
When a deployment uses ShouldThrowException, arquillian does not include
the arquillian support classes in the deployed app, including the
Arquillian class which the test extends.

In this case, if arquillian is not on the class path on the server, this
test will fail because the test bean cannot be loaded because it depends
on the Arquillian class and so doesn't cause the expected deployment
exception.

Fixes #617 (again)

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/converters/convertToNull/ConvertedNullValueBrokenInjectionBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/converters/convertToNull/ConvertedNullValueBrokenInjectionBean.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Emily-Jiang (approved):
LGTM!

PR #649, opened 2020-11-04
Deploy spec files to Nexus.

radcortez:


spec/pom.xml
```diff
@@ -61,6 +60,7 @@
                             <goal>process-asciidoc</goal>
                         </goals>
                         <configuration>
+                            <outputFile>${project.build.directory}/generated-docs/${project.build.finalName}.pdf</outputFile>
```
Emily-Jiang, line 63: why do you define another `<outputFile>`?

spec/pom.xml
```diff
@@ -72,6 +72,7 @@
                             <goal>process-asciidoc</goal>
                         </goals>
                         <configuration>
+                            <outputFile>${project.build.directory}/generated-docs/${project.build.finalName}.html</outputFile>
```
Emily-Jiang, line 75: Is this line needed? We defined `<outputFile>` twice.

spec/pom.xml
Emily-Jiang: I suggest to define a property 
`<config-spec-pdf>${project.build.directory}/generated-docs/${project.build.finalName}.pdf</config-spec-pdf>` so that the following places can refer to that property

spec/pom.xml
Emily-Jiang: I suggest to define a property
<config-spec-html>${project.build.directory}/generated-docs/${project.build.finalName}.pdf</config-spec-html> so that the following places can refer to that property

spec/pom.xml
```diff
@@ -89,8 +90,55 @@
...
+                    <execution>
+                        <id>deploy-spec-pdf</id>
+                        <phase>deploy</phase>
+                        <goals>
+                            <goal>deploy-file</goal>
+                        </goals>
+                        <configuration>
+                            <repositoryId>ossrh</repositoryId>
+                            <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+                            <file>${project.build.directory}/generated-docs/${project.build.finalName}.pdf</file>
```
Emily-Jiang, line 117: <file>${config-spec-pdf}</file>

spec/pom.xml
```diff
@@ -89,8 +90,55 @@
...
+                    <execution>
+                        <id>deploy-spec-html</id>
+                        <phase>deploy</phase>
+                        <goals>
+                            <goal>deploy-file</goal>
+                        </goals>
+                        <configuration>
+                            <repositoryId>ossrh</repositoryId>
+                            <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+                            <file>${project.build.directory}/generated-docs/${project.build.finalName}.html</file>
```
Emily-Jiang, line 133: <file>${config-spec-html}</file>

spec/pom.xml
radcortez: One if for the pdf and the other one is for html.

spec/pom.xml
radcortez: One if for the pdf and the other one is for html.

spec/pom.xml
radcortez: Ah nvm I think when I did the change I didn't got yours so when it rebase it merged both.

spec/pom.xml
Emily-Jiang: I meant the line 77 also defines the element `<outputFile>`.

Emily-Jiang (approved):

spec/pom.xml
radcortez: Yes. I didn't notice you did a similar change and it got duplicated with the rebase.

Emily-Jiang:
Doing a 2.0-RC6 to test this out.

radcortez:
Ok :)

PR #655, opened 2020-11-11
Fix copyright

Emily-Jiang:

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -62,8 +62,55 @@
                 </dependencies>
             </plugin>
             <plugin>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>3.2.0</version>
+                <executions>
+                    <execution>
+                        <id>copy-resources</id>
+                        <phase>validate</phase>
+                        <goals>
+                            <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${basedir}/target/apidocs/doc-files/</outputDirectory>
+                            <overwrite>true</overwrite>
+                            <resources>
+                                <resource>
+                                    <directory>src/main/javadoc/licenses/${revremark}/</directory>
+                                    <filtering>true</filtering>
+                                </resource>
+                            </resources>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
+                <configuration>
+                        <failOnError>false</failOnError>
+                        <additionalparam>${javadoc.options}</additionalparam>
+                        <docfilessubdirs>true</docfilessubdirs>
+                        <attach>true</attach>
+                        <doclint>none</doclint>
+                        <doctitle>${project.name}</doctitle>
+                        <windowtitle>${project.name}</windowtitle>
+                        <header><![CDATA[<br>${project.name} v${project.version}]]></header>
+                        <bottom>
+<![CDATA[
+        Copyright &#169; ${inceptionYear} &#x2013; {currentYear} {organizationName}. All rights reserved.
+        <br>
+        Use is subject to <a href="{@docRoot}/doc-files/license.html" target="_top">license terms</a>.
+    ]]>
+                        </bottom>
+                        <tags>
+                            <tag>
+                                <name>implSpec</name>
+                                <placement>a</placement>
+                                <head>Implementation Specification:</head>
+                            </tag>
+                        </tags>
+                    </configuration>
                 <executions>
                     <execution>
                         <id>attach-javadocs</id>
--- a/spec/src/main/asciidoc/microprofile-config-spec.asciidoc
+++ b/spec/src/main/asciidoc/microprofile-config-spec.asciidoc
@@ -1,27 +1,10 @@
-//
-// Copyright (c) 2016-2019 Eclipse Microprofile Contributors:
-// Mark Struberg
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-//
-
 = Configuration for MicroProfile
 :authors: Mark Struberg, Emily Jiang, John D. Ament, Roberto Cortez, Jan Bernitt
 :email: struberg@apache.org, emijiang@uk.ibm.com, john.d.ament@gmail.com, radcortez@yahoo.com, jaanbernitt@gmail.com
 :version-label!:
 :sectanchors:
 :doctype: book
-:license: Apache License v2.0
+:license: :license: Eclipse Foundation Specification License v1.0
 :source-highlighter: coderay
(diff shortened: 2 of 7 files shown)


Emily-Jiang:
@radcortez I think we need to remove the Apache License v2 from the pdf and html as it looks very clumsy. This PR is in draft status and will need to work out the dates etc with Kevin.

radcortez:
Yes, @kwsutter already confirmed that the spec and java doc files need to drop the ASF licence.

api/pom.xml
```diff
@@ -64,6 +64,29 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
+                <configuration>
+                        <failOnError>false</failOnError>
+                        <additionalparam>${javadoc.options}</additionalparam>
+                        <docfilessubdirs>true</docfilessubdirs>
+                        <javadocDirectory>${project.basedir}/../src/main/javadoc</javadocDirectory>
```
radcortez, line 71: This can be dropped, since the default is `src/main/javadoc`.

api/pom.xml
```diff
@@ -64,6 +64,29 @@
...
+                        <attach>true</attach>
+                        <doclint>none</doclint>
+                        <doctitle>${project.name}</doctitle>
+                        <windowtitle>${project.name}</windowtitle>
+                        <header><![CDATA[<br>${project.name} v${project.version}]]></header>
+                        <bottom>
+<![CDATA[
```
radcortez, line 78: We can replace this with:

```xml
<bottom>
    <![CDATA[
        Copyright &#169; {inceptionYear}&#x2013;{currentYear} {organizationName}. All rights reserved.
        <br>
        Use is subject to <a href="{@docRoot}/doc-files/speclicense.html" target="_top">license terms</a>.
    ]]>
</bottom>
```

And set the `<inceptionYear>2017</inceptionYear>` in the Parent POM. The other properties are retrieved from the `organization` section.

radcortez (changes requested):

radcortez:
From what I understood from @kwsutter comment https://github.com/microprofile/microprofile-wg/issues/4#issuecomment-725555493, the spec source should use EPL but when doing the final release we need to change it for the EFSL. There is also a reference about it in Jakarta here: https://github.com/eclipse-ee4j/jakartaee-api.

Apparently in Jakarta that is done with Jenkins, but I think we can improve it and just set a maven property pointing to one license or the other and set the default to EPL, but change it in the `release` profile to the EFSL.

api/pom.xml
Emily-Jiang: Good suggestions! As for the inceptionYear, since most of the spec asciidoc for the specs started 2016. I have set the inceptionYear to 2016.

Emily-Jiang:
> From what I understood from @kwsutter comment [microprofile/microprofile-wg#4 (comment)](https://github.com/microprofile/microprofile-wg/issues/4#issuecomment-725555493), the spec source should use EPL but when doing the final release we need to change it for the EFSL. There is also a reference about it in Jakarta here: https://github.com/eclipse-ee4j/jakartaee-api.
> 
> Apparently in Jakarta that is done with Jenkins, but I think we can improve it and just set a maven property pointing to one license or the other and set the default to EPL, but change it in the `release` profile to the EFSL.

@radcortez In our spec pdf/html, the source are using Apache V2, as you can see from each individual ascii doc. The generated pdf/html has to use EFSL, which is what I read from @kwsutter . Kevin's comments of final release vs. non-final release is for javadoc. Can you double check?

kwsutter:
@Emily-Jiang and @radcortez,  There are many ways to accomplish the switch from Apache v2 to EFSL.  We are currently doing it via our Jenkins build for the Jakarta EE Platform, but we have an Issue to modify this approach and just do this via a build property.  Or, as Emily pointed out, you could piggy back on the "revmark".  Whatever works is fine.  There's no requirement on how this accomplished, it's just a requirement that the final release version of the Specification and Javadoc use the EFSL license.  Non-final versions of the Spec and Javadoc can use Apache v2, EPL, or whatever the project is currently using for the source.

radcortez:
> Non-final versions of the Spec and Javadoc can use Apache v2, EPL, or whatever the project is currently using for the source.

So, for instance a `RC` release does not have to update, right?

Emily-Jiang:
> > Non-final versions of the Spec and Javadoc can use Apache v2, EPL, or whatever the project is currently using for the source.
> 
> So, for instance a `RC` release does not have to update, right?

I think so. It is not called final release.

kwsutter (approved):
These changes look like the right approach.  Thanks!

api/pom.xml
radcortez: Great :)

radcortez:
Ok, I think these clear all the questions. @Emily-Jiang please proceed with the remaining suggestions. Thanks :)

spec/src/main/asciidoc/microprofile-config-spec.asciidoc
```diff
@@ -30,8 +13,11 @@ ifdef::backend-pdf[]
 :pagenums:
 endif::[]
 
+// == License
+:sectnums!:
+include::license-efsl.adoc[]
 
-include::license-alv2.asciidoc[]
```
radcortez, line 34: Shouldn't we do the same thing here and just change the license when doing a Final release?

Emily-Jiang: From my understanding, Jakarta Specs use EFSL licenses all the time. Kevin did say the generated doc/html not source files uses EFSL. The generated pdf/htmls means all releases including the local generated files. @kwsutter please can you confirm?

radcortez: @kwsutter mentioned this:

> it's just a requirement that the final release version of the Specification and Javadoc use the EFSL license. Non-final versions of the Spec and Javadoc can use Apache v2, EPL, or whatever the project is currently using for the source.

If we can use the EFSL all the time in the spec files it is fine for me.

kwsutter: This is kind of confusing...  And, to be honest, you may not find consistency across all of the Jakarta EE specs... The requirement is that the Final generated Specifications and Javadocs have to contain the EFSL.  These are the Final specs and javadoc available via our Specifications page:  https://jakarta.ee/specifications/ .  When we produced our Milestone1 release for Jakarta EE 9, most of the Specifications and Javadocs also used the EFSL license, but we were not as picky with verifying that.  Most Specification projects, if you just clone the repo and rebuild the specs as a developer, you'll end up with EPL (or Apache in some cases) in the generated Specs and Javadoc.

The key requirement is that the Final released version of the Specs and Javadocs have to contain the EFSL.

kwsutter: > Most Specification projects, if you just clone the repo and rebuild the specs as a developer, you'll end up with EPL (or Apache in some cases) in the generated Specs and Javadoc.

@Emily-Jiang pointed out that this is not always the case (at least for the Platform specs).  For the Platform, we always use the EFSL in the generated Spec pdf and html.  But, for the API, we will use the EPL for normal development builds of the Javadoc and use the EFSL when producing the Final version of Javadoc.

Emily-Jiang: ok, thank you @kwsutter ! I did clone a few Jakarta-EE specs and realised that the specs always use EFSL. @kwsutter confirmed that he was a bit confused and confirmed that spec pdf/html always use EFSL and javadoc is the one that need to be differentiated.

Emily-Jiang: Based on the previous comments, no new actions are to be performed.

PR #663, opened 2020-11-27
Use MP Parent POM.

radcortez:
Build is expected to fail, since Parent POM is not yet available. This is to showcase the differences when using the Parent POM.


Emily-Jiang (changes requested):
@radcortez I guess we can only merge this after the parent pom.xml was done. I'll take a look at that later. I suggest we wait till MP 4.0 is out of door, which should happen soon.

radcortez:
Yes, I'm not going to merge it now. This was mostly to test the Parent POM and show how the project is going to look like in advance.

api/pom.xml
```diff
@@ -17,183 +17,28 @@
...
-        </dependency>
         <dependency>
             <groupId>jakarta.enterprise</groupId>
             <artifactId>jakarta.enterprise.cdi-api</artifactId>
-            <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>org.osgi</groupId>
-            <artifactId>org.osgi.service.cdi</artifactId>
+            <groupId>biz.aQute.bnd</groupId>
```
Emily-Jiang, line 35: Can you move this dependency to the parent pom as well, becuase all of the specs need this?

api/pom.xml
```diff
@@ -17,183 +17,28 @@
...
+            <artifactId>biz.aQute.bnd.annotation</artifactId>
         </dependency>
         <dependency>
             <groupId>org.osgi</groupId>
-            <artifactId>osgi.annotation</artifactId>
+            <artifactId>org.osgi.service.cdi</artifactId>
```
Emily-Jiang, line 40: I think you swapped the dependency around without any change. Is it a ordering thing?

api/pom.xml
```diff
@@ -17,183 +17,28 @@
...
-                    </archive>
-                </configuration>
-            </plugin>
-            <plugin>
-                <groupId>biz.aQute.bnd</groupId>
-                <artifactId>bnd-baseline-maven-plugin</artifactId>
-                <version>${bnd.version}</version>
-                <configuration>
-                    <base>
-                        <version>1.4</version>
```
Emily-Jiang, line 162: How did you specify the baseline version?

api/pom.xml
radcortez: Hum, I had a look into the other projects and I couldn't find any usages. Can you point me to where it is being used other than config?

api/pom.xml
radcortez: Yes.

api/pom.xml
radcortez: The version defaults to the highest version of the project's artifact that is less than the version of the project's artifact. Meaning that with this change it will be `2.0`. If we want to use another version, we can just override the plugin definition and configuration coming from Parent.

Emily-Jiang: I tried to use the latest version from maven central before. However, once you deliver RCx to maven central, the RCx will be used. Between the RCm, RCn, we are not allowed to update mistakes wihout version updates. I raised an [issue](https://github.com/bndtools/bnd/issues/4240) with bnd to ask for a facility on just pulling in released versions. The issue was solved in the bnd plugin 5.2. Let me create a PR on parent.pom to pull in the latest version and specify a property and then this can be done. [missed the bnd release notes, hence my previous question.]

api/pom.xml
Emily-Jiang: This was needed by OSGi runtime. They planned to integrate with all MP specs and consumed Config first. I have not seen more update then. Let's leave it for now and move the dependency to parent when they resume their activities.

api/pom.xml
Emily-Jiang: fixed the parent pom.xml via this [PR](https://github.com/eclipse/microprofile-parent/pull/37)

radcortez: Thanks!

radcortez:
@Emily-Jiang are you ok if we push a Parent Release and update Config?

Emily-Jiang (approved):
LGTM!

Emily-Jiang:
oops. You need to fix the GH Actions first before you merge.

radcortez:
Yes, the issue is that the Parent POM is not released in Central yet.

PR #679, opened 2021-03-12
Spec doesn't mention String as a built-in converter

tevans78:
Spec doesn't mention String as a built-in converter ... the Converter JavaDoc does
https://github.com/eclipse/microprofile-config/blob/4bda7c776e00a3baf41af728215900e02485de2c/api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java#L70

### DIFF
--- a/spec/src/main/asciidoc/converters.asciidoc
+++ b/spec/src/main/asciidoc/converters.asciidoc
@@ -41,6 +41,7 @@ The following ``Converter``s are provided by MicroProfile Config by default:
 * `double`, `java.lang.Double`, and `java.util.OptionalDouble`; a dot '.' is used to separate the fractional digits
 * `char` and `java.lang.Character`
 * `java.lang.Class` based on the result of `Class.forName`
+* `java.lang.String`
 
 All built-in ``Converter``s have the `@Priority` of `1`.
 


radcortez:
Hi @tevans78 . Thanks!

Can you include the "Signed-off-by" footer: https://accounts.eclipse.org/legal/eca/validation/62168? We can't merge without it.

Emily-Jiang:
`String` is an automatic converter as it has a public constructor taking a string parameter. The spec does not mention String as a default convertor either. Therefore, I don't think any changes are needed here.

spec/src/main/asciidoc/converters.asciidoc
```diff
@@ -41,6 +41,7 @@ The following ``Converter``s are provided by MicroProfile Config by default:
 * `double`, `java.lang.Double`, and `java.util.OptionalDouble`; a dot '.' is used to separate the fractional digits
 * `char` and `java.lang.Character`
 * `java.lang.Class` based on the result of `Class.forName`
+* `java.lang.String`
```
Emily-Jiang, line 44: It is an automatic converter. No need to be mentioned here. I think the javadoc needs to be updated to have that line removed instead.

Emily-Jiang (changes requested):

tevans78:
@radcortez The signoff is missing because I used the github.com web UI to create the PR automatically.
Since @Emily-Jiang thinks the change is not the right one anyway, I think I will close this PR and create an issue instead.

PR #696, opened 2021-05-14
Remove remaining instances of URL.equals()

### LINKED ISSUE #695: URL.equals() is still being used in TCKs
Issue #549 highlighted the problem with using `URL.equals()` and PR #567 attempted to resolve it. 

However, [ConverterTest.java](https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/ConverterTest.java) is still using the method here:
https://github.com/eclipse/microprofile-config/blob/0127e22c37149a46b8222cb79a29ab6355d24208/tck/src/main/java/org/eclipse/microprofile/config/tck/ConverterTest.java#L716-L720

and here: 
https://github.com/eclipse/microprofile-config/blob/0127e22c37149a46b8222cb79a29ab6355d24208/tck/src/main/java/org/eclipse/microprofile/config/tck/ConverterTest.java#L727-L731

I suggest we move the `assertURL` methods from #567 to a common package, perhaps [AdditionalMatchers.java](https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/matchers/AdditionalMatchers.java)?

Joseph-Cass:
Resolves #695

(note: I haven't yet tested this against an implementation of the specification)


Azquelt (changes requested):
I think you've missed out actually updating `ConverterTest` to use your new equality methods.

Additionally, your URL assertion methods will not give helpful error messages when the URLs are not equal.

tck/src/main/java/org/eclipse/microprofile/config/tck/util/AdditionalAssertions.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private AdditionalAssertions() {
+        // utility class
+    }
+
+    public static void assertURLArrayEquals(URL[] value, URL[] expectedValue) throws MalformedURLException {
+        assertURLListEquals(Arrays.asList(value), Arrays.asList(expectedValue));
+    }
+
+    public static void assertURLListEquals(List<URL> value, List<URL> expectedValue) throws MalformedURLException {
```
otaviojava, line 47: What do you think if we repleace `List` to `Iterable`

tck/src/main/java/org/eclipse/microprofile/config/tck/util/AdditionalAssertions.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
```
Emily-Jiang, line 2: 2017, 2021
Since you are refactoring the code instead of writing from scratch, you need to use the original copyright year and copy the authors from the original file as well.

tck/src/main/java/org/eclipse/microprofile/config/tck/util/AdditionalAssertions.java
Joseph-Cass: The rest of the class only does Array, List and Set comparisons, where the order of the List comparisons matters, but the order of the set comparisons doesn't (despite being a `LinkedHashSet`) so I'm not sure if this is worth it.

Azquelt (approved):

Emily-Jiang (approved):
LGTM! Thanks @Joseph-Cass !

Emily-Jiang:
The build break should be fixed by the checkstyle update https://github.com/eclipse/microprofile-parent/pull/43.

Azquelt:
The build is actually failing because it's running against an older commit on this PR.

When the build runs, it checks out a commit which is a merge of the PR branch and the target branch. Looking at the "Checkout" section of the build, we can see it checks out dad14562ddde03a069a398a5038beaebe6912252

However, if we fetch this commit from the repository and look at the log, we can see its parent is actually b2c1694, rather than the expected bfed586:

```
$ git fetch origin dad14562ddde03a069a398a5038beaebe6912252
From github.com:eclipse/microprofile-config
 * branch            dad14562ddde03a069a398a5038beaebe6912252 -> FETCH_HEAD

$ git log dad14562ddde03a069a398a5038beaebe6912252 --oneline
dad1456 (HEAD) Merge b2c1694bd58f85bb2b56d2a8c2b73a3e1d7436f7 into 0127e22c37149a46b8222cb79a29ab6355d24208
b2c1694 Address feedback
[...]
```

If I check out b2c1694 and run the build locally, I do see the same failure that is reported in the build.

I've tried re-running the build job but it still checks out the wrong commit and consequently fails. I'm not sure why the github action is pulling down the wrong commit. Possibly something has gotten cached incorrectly somewhere and amending the last commit and repushing would fix it?

Joseph-Cass:
Hmm good spot. Thank you @Azquelt, that seems to have done the trick!

Azquelt:
Worth noting that checkstyle still reports errors, so the change in eclipse/microprofile-parent#43 is still needed (though possibly we should be making the build fail if there are checkstyle issues).

Issue #668, opened 2021-01-07
Upgrade Eclipse MicroProfile with Jakarta EE 9


otaviojava:
Hello everyone, is there any roadmap to Eclipse MicroProfile that supports Jakarta EE 9?

The Working Group brings a new perspective to Jakarta EE specs such as JPA, NoSQL, JMS, etc.

It will be great to either standardize or define good practices to create complex objects from Eclipse MIcroProfile Config, such as an EntityManager or a NoSQL database connection.

ederks85:
Hi Otávio,

This is one of the topics I will bring up for discussion as soon as
possible with the steering committee. A few work items have been defined,
like creating a document where we will collect roadmap items, actions, and
future plans for MicroProfile. We discussed this on the last Hangout, you
can watch this part in the recording if you like.

Edwin

keilw:
@ederks85/all,

Based on the discussions in https://www.eclipse.org/lists/cn4j-alliance/msg00001.html there are several barriors for any MP API like Config to ever be consumed by Jakarta EE even in an implementation (and I am not convinced a Jakarta Spec or API would even need them anyway) bu this highlights the biggest problem aside from a **maturity level** and the general question of **namespace** (again implementations have very little problem to consume the same top level package they also got in many cases)

brunobat:
I am totally against this. 
MicroProfile should only change to Jakarta EE 10 and in the case that release has meaningful features. 
Updating all the namespaces will be a big overhead to whoever is using MicroProfile.
I can tell you I won't commit resources to this until I know what's in for Jakarta EE 10. I don't want to be in a position where I need to revert my services to use something else in order not to be stuck in a potential dead end.

keilw:
Then MP is dead at least for Jakarta EE, but so far you haven't even made it to Jakarta EE 9 while the first implementations already exist for Jakarta 9 but not a single vendor has implemented or applied MP 4 yet.
It is becoming the break block.

radcortez:
> Then MP is dead at least for Jakarta EE (even on an implementation level), but so far you haven't even made it to Jakarta EE 9 while the first implementations already exist for Jakarta 9 but not a single vendor has implemented or applied MP 4 yet.
> This way it is becoming the break block and legacy instead of an innovator.

This is incorrect. 

Wildfly just released a Beta last week with full support for MP 4.0:
https://twitter.com/bestansberry/status/1364935448464867328?s=20

Both Liberty and Quarkus are on their way to support it as well:
[...]

PR #699, opened 2021-05-26
Alignment with Jakarta EE 9

Emily-Jiang:
Fixes: #668

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
@@ -44,5 +44,5 @@
  * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  */
-@org.osgi.annotation.versioning.Version("1.1")
+@org.osgi.annotation.versioning.Version("2.0")
 package org.eclipse.microprofile.config.inject;
--- a/api/src/main/java/org/eclipse/microprofile/config/package-info.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/package-info.java
@@ -82,5 +82,5 @@
  * @author <a href="emijiang@uk.ibm.com">Emily Jiang</a>
  * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
  */
-@org.osgi.annotation.versioning.Version("2.0.0")
+@org.osgi.annotation.versioning.Version("3.0.0")
 package org.eclipse.microprofile.config;
--- a/api/src/main/java/org/eclipse/microprofile/config/spi/package-info.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/spi/package-info.java
@@ -32,5 +32,5 @@
  * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
  *
  */
-@org.osgi.annotation.versioning.Version("2.0")
+@org.osgi.annotation.versioning.Version("3.0")
 package org.eclipse.microprofile.config.spi;
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <groupId>org.eclipse.microprofile.config</groupId>
         <artifactId>microprofile-config-parent</artifactId>
-        <version>2.1-SNAPSHOT</version>
+        <version>3.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>microprofile-config-api</artifactId>
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -51,7 +51,7 @@ import org.eclipse.microprofile.config.spi.Converter;
  * the programmatically created ones are not required to be serializable.
  * <p>
  * If one or more converters are registered for a class of a requested value then the registered
- * {@link org.eclipse.microprofile.config.spi.Converter} which has the highest {@code @javax.annotation.Priority} is
+ * {@link org.eclipse.microprofile.config.spi.Converter} which has the highest {@code @jakarta.annotation.Priority} is
  * used to convert the string value retrieved from the config sources.
  *
  * <h2>Usage</h2>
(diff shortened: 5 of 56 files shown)


api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
```diff
@@ -44,5 +44,5 @@
  * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
  * @author <a href="mailto:emijiang@uk.ibm.com">Emily Jiang</a>
  */
-@org.osgi.annotation.versioning.Version("1.1")
+@org.osgi.annotation.versioning.Version("2.0")
```
radcortez, line 47: How about if we align these versions?

api/src/main/java/org/eclipse/microprofile/config/package-info.java
```diff
@@ -82,5 +82,5 @@
  * @author <a href="emijiang@uk.ibm.com">Emily Jiang</a>
  * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
  */
-@org.osgi.annotation.versioning.Version("2.0.0")
+@org.osgi.annotation.versioning.Version("3.0.0")
```
radcortez, line 85: How about if we align these versions?

radcortez (commented):
Don't we need changes in the spec?

api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
Emily-Jiang: This is the semantic version for the corresponding package. They are all individually version to indicate whether there are any backward incompatible changes between teh releases. There should not be any alignment with the different packages. As you might have noticed, not all of the packages have changed their versions. For more [info](http://docs.osgi.org/whitepaper/semantic-versioning/040-semantic-versions.html) regarding semantic versioning. We use bnd versioning plugin to work out all of the versions.

api/src/main/java/org/eclipse/microprofile/config/package-info.java
Emily-Jiang: @radcortez please see my comments [here](https://github.com/eclipse/microprofile-config/pull/699#discussion_r640165314)

api/src/main/java/org/eclipse/microprofile/config/spi/package-info.java
```diff
@@ -32,5 +32,5 @@
  * @author <a href="mailto:struberg@apache.org">Mark Struberg</a>
  *
  */
-@org.osgi.annotation.versioning.Version("2.0")
+@org.osgi.annotation.versioning.Version("3.0")
```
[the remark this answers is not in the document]
Emily-Jiang: same as above.

api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
radcortez: I know, but this is almost like starting fresh, and I think it wouldn't hurt to align all versions (even if there are no changes, but to mark a baseline with the Jakarta rename) and then follow the regular versioning scheme going forward.

Emily-Jiang: It's not like starting from scratch though. We have been follow the regular versioning scheme from the beginning. I think it is expecially important to obey the rules as you can see some packages are not even affected by the changes so their version will not be updated. If we directly jump from 1.1->3.0, this will confuse end users.

Emily-Jiang: @radcortez do you have further comments? If not, can you approve this PR so that I can make further progress. At the moment, I am blocked by this.

ederks85 (approved):
LGTM. I have zero knowledge that might help the OSGI alignment discussion., unfortunately.

Emily-Jiang:
> LGTM. I have zero knowledge that might help the OSGI alignment discussion., unfortunately.

Thank you @ederks85 ! MP uses OSGi's semantic versioning policy and its plugin to work out backward incompatible changes. MP has benefited from this tool along the way to spell out backward incompatible changes and version the releases appropriately.

radcortez:
Feel free to do as you wish. I still we could align all version (it is ok to jump versions) and it would be easier to maintain in the future, then keeping track different versions for all the separate packages (and since we are breaking the API is ok to push new versions).

Emily-Jiang:
> Feel free to do as you wish. I still we could align all version (it is ok to jump versions) and it would be easier to maintain in the future, then keeping track different versions for all the separate packages (and since we are breaking the API is ok to push new versions).

There is difference between api package version and spec version. For instance, one package has not changed since the first release of MP Config, there is no reason to randomly bump up its version if no change was made since then. In this way, end users will immediately find out that this package has not got any changes since the previous releases. This is what semantic versioning is about. For spec version, if there is backward incompatible changes in any packages, the spec version will need to be bumped to indicate backward incompatible changes are in the current release.

radcortez:
We are breaking the package and we are bumping the versions it not like we are randomly bumping the version. My suggestion was to just align all versions (not necessarily align with the project version).

Emily-Jiang:
> We are breaking the package and we are bumping the versions it not like we are randomly bumping the version. My suggestion was to just align all versions (not necessarily align with the project version).

@radcortez I appreciate your thoughts on this. For the packages that do not rely on any jakarta namespace, we should just leave them alone. We have smenatic versioning plugin to help us, we can do it precisely correctly. Therefore, I am against the idea of bumping all packages unnecessarily.

api/src/main/java/org/eclipse/microprofile/config/inject/package-info.java
bjhargrave: Package versions should be semantic. Bundle versions and specification document versions are free to change. 

If you change the major version of a package then you affect all consumers of the package by breaking their imports as they, properly, import to exclude a major version bump as that means a breaking change.

Each package has an independent semantic versioning life. There is no reason to keep all package versions in lock step and there is no reason to change a major version unless there is a breaking change you need to communicate to users of the package.

A package's version is a message to the future. Make sure the message is clear and not noise.

PR #701, opened 2021-06-04
Avoid throwing Exception for Optional types and ConfigValue if expression cannot be expanded

### LINKED ISSUE #700: Avoid throwing Exception for Optional types and ConfigValue if expression cannot be expanded
**Description**

When an expression cannot be expanded, a `NoSuchElementException` is expected to be thrown however it should not be thrown in case of Optional types (Optional, OptionalInt, OptionalDouble and OptionalLong) and ConfigValue as it should rather behave like the configuration property was missing and thus give respectively an empty Optional or an empty ConfigValue.

Indeed, a typical use case, is to set a configuration property using an expression and only provide a value to that expression when needed since the configuration property is declared as optional.

essobedo:
fixes #700

## Motivation

The TCK doesn't cover the case where an expression cannot be expanded but the target type is an Optional types or ConfigValue that should not throw `NoSuchElementException`, the goal of this PR is to add new tests to ensure that the expected behavior is consistent. 

## Modifications

* Indicates that in case of ConfigValue, an empty ConfigValue is expected if the expression cannot be expanded.
* Adds new tests to cover this particular use case.

### DIFF
--- a/spec/src/main/asciidoc/property-expressions.asciidoc
+++ b/spec/src/main/asciidoc/property-expressions.asciidoc
@@ -55,7 +55,8 @@ server.endpoint.path.bar=foo
 The property `server.url` is expanded to `http://example.org:8080/foo`.
 
 If an expression cannot be expanded and does not have a default value, a `NoSuchElementException` is thrown. In the
-Optional case, an empty Optional will be returned.
+Optional case, an empty Optional will be returned. In the `ConfigValue` case, an `ConfigValue` with only
+the name of the property will be returned.
 
 The number of recursion lookups is not infinite, but a limited number for composed expressions. Implementations are
 encouraged to limit the number to `5`, but they can use higher limits if they wish to. When the number of allowed
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPropertyExpressionsTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/CDIPropertyExpressionsTest.java
@@ -19,9 +19,16 @@
 package org.eclipse.microprofile.config.tck;
 
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertNull;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
+import java.util.OptionalLong;
 import java.util.Set;
 
 import org.eclipse.microprofile.config.ConfigValue;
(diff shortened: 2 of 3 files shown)


Emily-Jiang (approved):
LGTM! Thanks @essobedo for your contribution!

radcortez (approved):

radcortez:
This is our first PR after the jakarta rename in master.

We agreed to not push any modifications to the jakarta rename, so I suggest to make a new branch (maybe `config-mp6`) to accept any changes from now on.

Emily-Jiang:
> This is our first PR after the jakarta rename in master.
> 
> We agreed to not push any modifications to the jakarta rename, so I suggest to make a new branch (maybe `config-mp6`) to accept any changes from now on.
This change could end up in mp5.x. My personal view is that we should quickly release mp 5.0 with namespace updates and then add more functionalities. Maybe we should open a branch post-jakarta to list all of the changes to be merged after the jakarta namespace update. I have added an item on the next technical calls to discuss across the specs.

essobedo:
@radcortez @Emily-Jiang If this change end up in mp5.x, I would go a little bit further. 

Indeed, I believe that this part of the spec should be clarified a little bit more, I quote `If an expression cannot be expanded and does not have a default value, a `NoSuchElementException` is thrown`. This statement makes perfect sense, if the expression cannot be expanded, it should be seen as not set, so to avoid getting a `NoSuchElementException`, we can either use Optional types or ConfigValue or we can provide a default value. So far so good, the problem is that the TCK seems to assume that the only way to provide a default value is through the expression itself while it is also possible thanks to the annotation ConfigProperty like in the next example:

```
@Inject
@ConfigProperty(name = "bad.property.expression.prop", defaultValue = "my-default")
String expressionDefault;
```

Where `bad.property.expression.prop` is set to `${missing.prop}` and `missing.prop` is missing.

In SmallRye Config, the use case described above will throw a `NoSuchElementException` and IMHO it should not.
Don't you agree?
WDYT?

radcortez:
One of the issues with the `defaultValue` in the annotations is that we never specified if this should apply to the injection point only or if this should apply globally. And what if you do a programatic lookup? See https://github.com/eclipse/microprofile-config/issues/475. We should definitely figure out this one first.

essobedo:
> One of the issues with the `defaultValue` in the annotations is that we never specified if this should apply to the injection point only or if this should apply globally. And what if you do a programatic lookup? See #475. We should definitely figure out this one first.

Whaoo thank you for the link, I would have never imagined that a default value defined in one specific injection point could affect another one (the butterfly effect?), IMHO it would be really error prone, in term of maintenance of a big project it could have a significant impact but as you are the experts it is up to you 😄  

Anyway thank you again for sharing the link, I will follow it with caution as it could impact my projects.

Emily-Jiang:
> @radcortez @Emily-Jiang If this change end up in mp5.x, I would go a little bit further.
> 
> Indeed, I believe that this part of the spec should be clarified a little bit more, I quote `If an expression cannot be expanded and does not have a default value, a `NoSuchElementException` is thrown`. This statement makes perfect sense, if the expression cannot be expanded, it should be seen as not set, so to avoid getting a `NoSuchElementException`, we can either use Optional types or ConfigValue or we can provide a default value. So far so good, the problem is that the TCK seems to assume that the only way to provide a default value is through the expression itself while it is also possible thanks to the annotation ConfigProperty like in the next example:
> 
> ```
> @Inject
> @ConfigProperty(name = "bad.property.expression.prop", defaultValue = "my-default")
> String expressionDefault;
> ```
> 
> Where `bad.property.expression.prop` is set to `${missing.prop}` and `missing.prop` is missing.
> 
> In SmallRye Config, the use case described above will throw a `NoSuchElementException` and IMHO it should not. Don't you agree? WDYT?

@essobedo it is a little bit tricky. Once a value was provided in a config source, it takes precedence. The `defaultValue` will be overridden. For the `bad.property.expression.prop` case, any unexpandable cases should cause an error. Otherwise, it will be very difficult to debug if we treated as if the property is not defined.

Emily-Jiang:
I will merge this PR as it is. Any related changes should be noted in a separate issue with a separate PR.

Emily-Jiang (approved):

PR #711, opened 2021-08-26
Tree node support.

tomas-langer:
This PR shows how tree structure could be supported using MP config.
I have updated the proposal with a bit simplified version.

## API changes
Config now has a method `get(String name)` to obtain a child node of the current node.
It also has several typed methods to get the current node as a typed value

## SPI changes
Config source can now provide the full tree instead of a single node
Converter can convert from a node (object node, list node, value node) to a type

This PR was created as a result of discussion on Jakarta Config meeting, so we have a vehicle to discuss tree structure in configuration.

Some decisions that I made and explanation:
- there are two new methods (`isMutable` and `changeListener`) on `ConfigSource`. Unfortunately the tree structure is connected with mutability, as if a source provides the full tree, it is no longer queried for each node separately, and so the existing approach of asking each source in sequence (and thus supporting some kind of "maybe mutability") is no longer possible. The new approach is allowing a config source to declare its support for mutations, and an event support to notify configuration of such a change. The configuration may or may not be interested in changes - if not, method `changeListener` is not called...
- instead of creating three interfaces, I have added a method `getType()` and default methods to the config source that support each type. This is for backward compatibility (and also to make it easy for config source implementation - you just need to know about a single interface). This is a bit more complicated for config implementation, as it needs to check for source type, and then call the correct method to get values

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/config/Config.java
+++ b/api/src/main/java/org/eclipse/microprofile/config/Config.java
@@ -385,7 +385,7 @@ public interface Config {
      * server.ports.1=8082
      * </pre>
      * <p>
-     * {@link Type#VALUE} - the nodes {@code server.port} and {@code server.host} are values
+     * {@link Type#LEAF} - the nodes {@code server.port} and {@code server.host} are values
      * <pre>
      * server.port=8080
      * server.host=localhost
@@ -474,7 +474,7 @@ public interface Config {
         /**
          * Value node is a leaf node - it does not have any child nodes, only direct value.
          */
-        VALUE,
+        LEAF,
         /**
          * Node is missing, it will return only empty values.
          */


api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -291,4 +303,181 @@
...
+     *      (in an unknown order)
+     *
+     * @return list of child nodes, or empty if this node does not exist
+     */
+    Optional<List<Config>> asNodeList();
+
+    /*
+     * Shortcut helper methods
+     */
+    default Optional<String> asString() {
```
Emily-Jiang, line 451: This is only callable if the hasValue() returns true. I am wondering whether we layer this further and move some of the methods to LeafNode.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -291,4 +303,181 @@
...
+    }
+
+    /**
+     * Config node type.
+     */
+    enum Type {
+        /**
+         * Object node with named members and a possible direct value.
+         */
+        OBJECT,
```
Emily-Jiang, line 466: I like the Type structure, but not the enum value names. I think it is pretty much: Parent, List, Leaf, Empty

api/src/main/java/org/eclipse/microprofile/config/Config.java
tomas-langer: This is practically not true - there can be a converter from config node to string that could work even for list nodes (e.g. converting a list to a comma separated string)

api/src/main/java/org/eclipse/microprofile/config/Config.java
tomas-langer: Parent is not a very good name, as it can have a value (it is basically a tree node that may contain child nodes and a value).
Leaf is OK. Empty vs. Missing - I think missing is more spot on (as this config node is missing from configuration). Empty may mean something different (e.g. the node exists, but is empty).
The values should still be all upper case (best practice for naming enum values)

tomas-langer: Renamed value to `LEAF`

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -198,4 +201,103 @@ default int getOrdinal() {
...
+     */
+    default void changeListener(Runnable changeRunnable) {
+    }
+
+    /**
+     * Type of the config source.
+     */
+    enum Type {
+        /**
+         * This is for backward compatibility - request each property separately.
```
Emily-Jiang, line 271: Should this be `FLAT` to state all properties are visible with a flat structure?

api/src/main/java/org/eclipse/microprofile/config/spi/Converter.java
```diff
@@ -150,4 +152,18 @@
      *             if the given value was {@code null}
      */
     T convert(String value) throws IllegalArgumentException, NullPointerException;
+
+    /**
+     * Convert the given config node (may be an object, list or value) to the specified type.
```
Emily-Jiang, line 157: If this config node is a node instead of leaf, I assume an `IllegalArugumentException` should be thrown.

tomas-langer: A converter may be capable of converting object nodes or list nodes as well. If the converter converts to a POJO for example, accepting an Object node is absolutely fine

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
tomas-langer: The naming could be anything, I just wanted to have an option that is backward compatible. If `FLAT` sounds better, no problem.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -198,4 +201,103 @@ default int getOrdinal() {
...
+
+    /**
+     * If this config source is {@link #isMutable()} and its shape changes, it should notify
+     * the configuration implementation of such a change.
+     * Once the change listener is registered by configuration implementation, this source can
+     * start checking for changes. If any change occurs, the source should invoke the runnable.
+     *
+     * @param changeRunnable runnable to call when this config source changes
+     */
+    default void changeListener(Runnable changeRunnable) {
```
Emily-Jiang, line 263: I think `onChange` is a better name.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
```diff
@@ -198,4 +201,103 @@ default int getOrdinal() {
...
+        return true;
+    }
+
+    /**
+     * If this source {@link #getType()} is {@link Type#TREE}, this method is used to load the
+     * tree of the data.
+     *
+     * @return node content or empty optional if the underlying config source does not exist
+     */
+    default Optional<ConfigNode> load() {
```
Emily-Jiang, line 240: I think we should need this method if we update the following method `load(String key)` to state that if the `key` is null, it will load the whole tree. Otherwise, load the tree with the specified `root`. I would change the signature `load(String root)` while `null` means the whole root.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigNode.java
```diff
@@ -0,0 +1,135 @@
...
+    Integer sourcePriority();
+
+    /**
+     * Base types of config nodes.
+     */
+    enum NodeType {
+        /**
+         * An object (complex structure), optionally may have a value.
+         */
+        OBJECT,
```
Emily-Jiang, line 80: I still find this confusing. `LIST` and `VALUE` are all `OBJECT`. How about `STRUCTURE` or `COMPLEX`  to differentiate them.

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -291,4 +303,181 @@
...
+    }
+
+    default Optional<Integer> asInt() {
+        return as(Integer.class);
+    }
+
+    /**
+     * Config node type.
+     */
+    enum Type {
```
Emily-Jiang, line 462: We have `NodeType` under `ConfigNode` and this `Type` again for config node type. Can we list one out and refer to it in both `Config` and `ConfigNode`?

api/src/main/java/org/eclipse/microprofile/config/Config.java
```diff
@@ -291,4 +303,181 @@
...
+    /**
+     * A detached node removes prefixes of each sub-node of the current node.
+     * <p>
+     * Let's assume this node is {@code server} and contains {@code host} and {@code port}.
+     * The method {@link #key()} for {@code host} would return {@code server.host}.
+     * If we call a method {@link #key()} on a detached instance, it would return just {@code host}.
+     *
+     * @return a detached config instance
+     */
+    Config detach();
```
Emily-Jiang, line 347: I am not sure about the use case here. When do you need to do this? Is this purly about shorter the key string?

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java
tomas-langer: I am against using `null` as a valid parameter value (anywhere, anytime). 
I can see two options here:
1. use empty string instead (as the root node does not have a name, so the key could be an empty string)
2. use a dedicated `Key` class that would have the concept of a root and use that

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigNode.java
tomas-langer: This is paritally taken from JSON, where `JsonObject` is the specific type (something that contains other nodes) and `JsonStructure` is the superset.
I am using `List` instead of `Array`, as I don't like arrays much in general. But otherwise this copies JSON-P naming

`JsonObject` -> OBJECT
`JsonArray` -> LIST (because collections are better than arrays ;) )
`JsonValue` -> VALUE

api/src/main/java/org/eclipse/microprofile/config/Config.java
tomas-langer: Because these are different scopes and different meanings.
One is intended for SPI (ConfigNode.Type) that supports just three types.
The other is for API (Config.Type) that supports also `MISSING`.
User's do not see `ConfigNode` unless they implement `ConfigSource`; SPI does not use Config. 
So separation of concerns.

api/src/main/java/org/eclipse/microprofile/config/Config.java
tomas-langer: this is to create a config that "cuts" the prefix from all of its nodes. Use case is when you need to configure a component that expects a map (such as JSON-B, CDI).
If you would not have such a method, you would need to create a map, and then rebuild the map by doing a substring on each key

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigNode.java
Emily-Jiang: I didn't make to the correlation with Json before you mentioned. I think if we prefix with Config, it might be less confusing, but refering to ConfigNode.OBJECT as ConfigNode.ConfigOBJECT is mouthful. It is not ideal, but I have not got a better name as yet:(. Maybe someone else has.

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigNode.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+
+package org.eclipse.microprofile.config.spi;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Marker interface identifying a config node implementation.
```
JHahnHRO, line 24: Doesn't "marker interface" usually refer to interfaces without methods?

api/src/main/java/org/eclipse/microprofile/config/spi/ConfigNode.java
```diff
@@ -0,0 +1,135 @@
...
+     * @return key of this node
+     */
+    String key();
+
+    /**
+     * Get the type of this node.
+     *
+     * @return NodeType this node represents
+     */
+    NodeType nodeType();
```
JHahnHRO, line 39: Some time, in the far, far future, the spec will require a minimum of Java 17. Should we prepare for this day by declaring the type enum to be a temporary workaround until this interface can be `sealed` in a future version?

Issue #716, opened 2021-09-17
Remove OSGi annotations from the API


radcortez:
The OSGi annotations added in https://github.com/eclipse/microprofile-config/pull/381/, cause annoying compiler warnings. 

The issue is that we mark the OSGi dependencies as provided and a reference to `org.osgi.annotation.bundle.Requirement.Resolution#OPTIONAL` is missing from the dependencies when compiling a project that uses `@ConfigProperty`.

Adding the OSGi dependency as compile does fix the issue, but I believe an MP API should not transitively push this kind of dependency to their consumers (especially if they are not using OSGi).

This can be reproduced just by using the MP Config API JAR, use the `@ConfigProperty` annotation and compile code with warnings on: `-Dmaven.compiler.showWarnings=true`. It generates the following warning:

```
[WARNING] unknown enum constant org.osgi.annotation.bundle.Requirement.Resolution.OPTIONAL
  reason: class file for org.osgi.annotation.bundle.Requirement$Resolution not found
```

This causes confusion to users, and we should remove this warning.

Emily-Jiang:
@rotty3000 this is the annotation you added for OSGi CDI. If I remembered correctly, this annotation should not cause any issues for non-OSGi environament. Apparently it caused problems for Quarkus. We might have to remove these annotations if this can't be resolved.

radcortez:
Just a clarification, this has nothing to do with Quarkus. You can reproduce the warning just with the MP Config API jar and with compiler warnings turned on.

rotty3000:
Though this is _not a runtime limitation in any way_, I don't think I know a way to prevent the compiler warning based on how the annotation is designed (to use a enum as property value).

I would ask @bjhargrave for advice on this.

bjhargrave:
This is because the the osgi annotations jar is not a compile scope dependency of the eclipse microprofile config jar. Generally tools need to see annotation definitions to see if they may be meta-annotated with annotations the tool needs to process. Javadoc wants to look for Documented meta-annotations, Bnd wants to look for meta-annotations it must process. Javac is likely also wanting to do the same. Thus the warning to indicate there is an annotation whose definition cannot be examined.

The solution is for the eclipse microprofile config jar to have a compile scope dependency on the osgi.annotation jar which would provide javac, javadoc, etc. which visibility to the annotation definitions.

[...]

PR #732, opened 2022-02-09
deps: Update osgi.annotation dependency to 8.1.0

bjhargrave:
Version 8.1.0 removes the use of enum types in the annotations which
avoids compiler warnings for downstream users of the api jar.

We also include osgi.annotation in the tck project to make the
annotation classes available to javadoc to avoid warnings.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,12 @@
         <dependencies>
             <!-- TODO - Investigate BND Dependencies-->
             <dependency>
+                <groupId>org.osgi</groupId>
+                <artifactId>osgi.annotation</artifactId>
+                <version>8.1.0</version>
+                <scope>provided</scope>
+            </dependency>
+            <dependency>
                 <groupId>biz.aQute.bnd</groupId>
                 <artifactId>biz.aQute.bnd.annotation</artifactId>
                 <version>5.3.0</version>
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -32,6 +32,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>osgi.annotation</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>jakarta.enterprise</groupId>
             <artifactId>jakarta.enterprise.cdi-api</artifactId>
         </dependency>


bjhargrave:
@Emily-Jiang This addresses the compiler warning issue in #716.

radcortez (approved):

radcortez:
Great. Thank you!

radcortez:
I believe this should be backported to 2.x.

bjhargrave:
> I believe this should be backported to 2.x.

I can make a PR for that. What branch?

radcortez:
> > I believe this should be backported to 2.x.
> 
> I can make a PR for that. What branch?

We don't have one. Let me create one. Thanks!

radcortez:
This should cover it: https://github.com/eclipse/microprofile-config/tree/2.x

Emily-Jiang:
Thank you @bjhargrave !

Emily-Jiang:
@radcortez we can do a service release once @bjhargrave makes a PR against 2.x.

brunolmfg:
Unfortunately that fix was merged into `master` after the `3.0.x` was branched. So the version `3.0.2` doesn't contain the upgrade of **osgi.annotation** to `8.1.0`.

Looking in the history I've notice some confusing versioning:
 * 33617450: `3.0` tagged;
 * 66b9611e: `3.0.1` tagged a first time;
 * d4a6e84e: `3.0.x` branched;
 * 8f2df48a: `3.0.1` tagged again in `master`;
 * 18d1b929: The fix merged in `master`;
 * 19ca7d88: `3.0.2` tagged in `3.0.x` branch.

@radcortez, @bjhargrave and @Emily-Jiang I believe this should be backported to `3.0.x`.

@turing85 It seems to me that [the problem you posted on Quarkus Zulip](https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/Quarkus.203.2E0.2E0.2EAlpha2.20and.20an.20old.20friend) is due to the upgrade of MP Config from `2.0.1` (**osgi.annotation** fixed) to `3.0` (**osgi.annotation** not fixed) present in Quarkus `3.0.0.Alpha1` and later.

radcortez:
We need to do a release to fix other stuff, so we might as well just release it as is and let consumers update to the new version.

PR #733, opened 2022-02-11
deps: Update osgi.annotation dependency to 8.1.0

bjhargrave:
Version 8.1.0 removes the use of enum types in the annotations which
avoids compiler warnings for downstream users of the api jar.

We also configure the maven-javadoc-plugin to be able to access
the package-list file for the osgi.annotation javadoc. The default
Java user agent value causes the server to deny the request.


radcortez:
Great! Thank you!

PR #743, opened 2022-04-19
Make sure TCKs are CDI-version agnostic

### LINKED ISSUE #742: Make the TCK tests CDI 4 compatible
As you might know, CDI 4 (EE 10) will bring some breaking changes one of which is the change of discovery mode in case of empty beans.xml. The original tracking issue can be seen at https://github.com/eclipse-ee4j/cdi/issues/500

The change is that such bean archive will now use annotated discovery mode and CDI container will only pick up beans with [bean defining annotations](https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#bean_defining_annotations). Previously, all classes were picked up and processes, so this mostly affects classes that had no CDI annotation but were expected to become beans.

This change affects the TCKs in this project.

Note that this can be fixed rather easily and in a compatible manner so that the TCKs will keep working in the current version (CDI 3) as well as with the updated version (CDI 4) with no further changes needed once the migration hits.

I have changes prepared locally and will submit a PR once I have it verified against an impl (using SR impl with updated Weld 5 container).

manovotn:
Fixes #742 

This doesn't change the TCK behavior in any way. It simply makes it CDI-version agnostic so that the TCK works with CDI 3 and CDI 4 implementations. It will ease transition for any early adopters and will eliminate future fraction when actual migration to EE 10 happens. See linked issue for more information.

This change was tested with SR impl customized to run the TCKs on Weld 5 container and the tests passed.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ConverterTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ConverterTest.java
@@ -62,6 +62,7 @@ import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
+import jakarta.enterprise.context.Dependent;
 import jakarta.inject.Inject;
 
 /**
@@ -98,8 +99,7 @@ public class ConverterTest extends Arquillian {
     }
 
     @Inject
-    @ConfigProperty(name = "tck.config.test.javaconfig.converter.duckname")
-    private Duck namedDuck;
+    InjectingBean bean;
 
     @Test(expectedExceptions = IllegalArgumentException.class)
     public void testDonaldNotConvertedByDefault() {
@@ -664,7 +664,7 @@ public class ConverterTest extends Arquillian {
 
     @Test
     public void testCustomConverter() {
-        Assert.assertEquals(namedDuck.getName(), "Hannelore");
+        Assert.assertEquals(bean.getNamedDuck().getName(), "Hannelore");
     }
 
     @Test
@@ -805,5 +805,18 @@ public class ConverterTest extends Arquillian {
         }
         Assert.assertEquals(((Converter<Duck>) readObject).convert("Donald").getName(),
                 original.convert("Donald").getName(), "Converted values to be equal");
+    }
+
+    // declare a proper bean with bean defining annotation as an injection target for config-related injection
+    @Dependent
+    public static class InjectingBean {
+
+        @Inject
+        @ConfigProperty(name = "tck.config.test.javaconfig.converter.duckname")
+        private Duck namedDuck;
+
+        public Duck getNamedDuck() {
+            return namedDuck;
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


manovotn:
I should add that I was frankly surprised to see this few tests are affected, but the rest was working just fine for me with CDI 4/Weld 5 (372 tests executed in total).

tck/src/main/java/org/eclipse/microprofile/config/tck/ConverterTest.java
```diff
@@ -98,8 +99,7 @@ public static WebArchive deploy() {
     }
 
     @Inject
-    @ConfigProperty(name = "tck.config.test.javaconfig.converter.duckname")
-    private Duck namedDuck;
+    InjectingBean bean;
```
Emily-Jiang, line 102: I think a simplier update is to make this Test class CDI bean as you see line 75 also uses `Inject`.

manovotn: @Emily-Jiang I see what you mean and noticed it myself but I disagree.

Currently, the test class is treated as a bean "by accident" as there is discovery mode `all` and the class will be processed as if it were a bean - all events fired from `ProcessAnnotatedType` to `ProcessInjectionTarget`.
However, Arquillian itself will never even use this bean and will instead create an instance of test class manually and then proceeds to detect injection points and use CDI to set the fields (this is why I say it works "by accident").
Marking the test as bean just re-introduces this side effect that has no real use. Or, to put it differently, marking the test as a bean doesn't resemble a real world scenario.

Looking at MP Config impl, I saw that a CDI extension is used to detect PAT and PIT and other events to see what beans are needed and then registering them synthetically based on discovered injection points - this is perfectly fine but it requires actual meaningful CDI bean which is what I added in this PR.

As for like 75 - this is fine because the `Config` bean is always present (whenever there is MP config impl) by definition and it won't interfere with anything. You could inject such bean anywhere or even resolve it directly from `CDI.current()` inside a non-contextual object which is what makes it different from the injection points I moved into a separate bean.

manovotn: Does my explanation make sense @Emily-Jiang?

Emily-Jiang: I am not sure whether you can inject the `Config` anywhere though. For supporting injection, it would have to be CDI beans or Java/Jakarta EE component classes. How would CDI container discover this test class and perform the injection? I am just worried about not being able to pass the tests from a Jakarta EE context. I completely understand you can get hold of the bean from CDI.current() but unsure about the support of the injection.

manovotn: @Emily-Jiang the injection into Arq. test instances is governed by Arq. adapter itself, see https://github.com/arquillian/arquillian-core/tree/master/testenrichers/cdi-jakarta
If memory serves, this will actually use something like `CDI.current()` to perform the injection into test instance.

The thing about `Config` is that the bean for it is registered without detecting an injection point requiring it. That's not the case for the rest of beans I moved elsewhere.

Emily-Jiang: ah, okay. Thanks for the explanation! I was worried about that being injected by the CDI container.

PR #750, opened 2022-08-31
add support for Map injection

### LINKED ISSUE #562: Allow to fetch multiple config parameters as a Map
**Description**

As a:

- [] API user (application developer)
- [x] SPI user (container or runtime developer)

...I need to be able to:
fetch multiple configuration parameters with same type as a Map
``` java
@ConfigProperty("my.prefix")
Map<String, String> configs
```
or fetch multiple configuration parameters with different type as a scoped Config
```
@ConfigProperty("my.prefix")
Config configs
```

...which enables me to:

Avoid the more generic
```
@inject
Config config

config.get("my.prefix" + myDynamicKey)
``

### PR DESCRIPTION
**Pull request by rmanibus**
fix #562

### CODE REVIEW

File: spec/src/main/asciidoc/configexamples.asciidoc [hunk 1]
```diff
@@ -121,6 +121,11 @@ public class InjectedConfigUsageSample {
     @ConfigProperty(name="myprj.some.supplier.timeout", defaultValue="100")
     private java.util.function.Supplier<Long> timeout;
 
+    //Injects a Map to resolve multiple configuration parameters under the same configuration key.
```
Emily-Jiang, line 124: For the Map injection, what types for Key, Value do you propose to support? We need to clarify that.

spec/src/main/asciidoc/configexamples.asciidoc
```diff
@@ -121,6 +121,11 @@ public class InjectedConfigUsageSample {
...
+    @Inject
+    @ConfigProperty(name = "myprj.some.reasons", defaultValue = "200=OK;201=Created")
+    private java.util.Map<Integer, String> reasons;
```
otaviojava, line 127: ´´´suggestion
private java.util.Map<String, String> reasons;
´´´
Once we're talking about configuration I believe the key must be a `String`

Emily-Jiang: I tend to agree to narrow down to Map<String, String>. Otherwise, it will be overwhelming with the tests and it might have more issues to deal with regarding conversion errors.

Issue #762, opened 2022-12-10
Clarify the required behaviour of @ConfigProperties beans


JHahnHRO:
The Spec does not explicitly state what kind of CDI Bean a class annotated with `@ConfigProperties` should give rise to. The way I understand it, it MUST be a dependent bean. My reasons are:
1. The class itself from the example is annotated with `@Dependent`
2. The `prefix` attribute in `@ConfigProperties` is `@NonBinding`. That means that the two example injection points
```
@Inject
@ConfigProperties
Details serverDetails;
```
and
```
@Inject
@ConfigProperties(prefix="client")
Details clientDetails;
```
are expected to resolve to the same bean. The only way the CDI spec allows for one bean to create two different instances depending on the injection point is for the bean to be `@Dependent` scoped.
3. Programmatic lookup with arbitrary prefixes only works if it is a `@Dependent` bean for the same reason.

HOWEVER: In the TCK there is [a test](https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigPropertiesTest.java#L185) with a `@RequestScoped @ConfigProperties` bean.
The only way for a normal scoped bean to give different instances for different injection points is to have the `prefix` attribute be binding again. And in fact, that is what smallrye does: It removes the `@NonBinding` annotation during `BeforeBeanDiscovery` and creates many, many synthetic beans from one `@ConfigProperties` class - one for each prefix that is discovered at injection points.

But that has another consequence: Programmatic lookup of `@ConfigProperties` no longer works as I would expect it. Again, the spec is not 100% clear here. The only example given is
```
Details details = CDI.current().select(Details.class, ConfigProperties.Literal.NO_PREFIX).get();
```
Does that mean that this is the *only* case of programmatic lookup that is required to work? Or is it required that programmatic lookup with arbitrary prefixes works? I would have expected the latter to be the case (and indeed I have filed this as a [smallrye bug](https://github.com/smallrye/smallrye-config/issues/860) )

radcortez:
Yes, I agree that `@ConfigProperties#prefix` should be binding. I've discussed that at some point in https://github.com/eclipse/microprofile-config/issues/629, but I ended up moving in the wrong direction :(

I also think that `@ConfigProperties` should not be `@RequestScoped`.  While technically possible, I think it would only generate uncertainty and frustration because the config result could be very different (because MP Config, leaves the door open for dynamic configuration).

[...]

PR #764, opened 2022-12-14
remove @RequestScoped @ConfigProperties bean and the associated test cases

JHahnHRO:
The test cases are fundamentally broken (see issue #762) and cannot possibly work the way they are written.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigPropertiesTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigPropertiesTest.java
@@ -21,6 +21,9 @@ package org.eclipse.microprofile.config.tck;
 
 import java.util.Optional;
 
+import jakarta.enterprise.context.Dependent;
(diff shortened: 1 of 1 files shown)


Emily-Jiang (changes requested):
The TCK is absolutely fine. Removing the scope support is a big change. What you raised is an implementation issue.

Emily-Jiang:
I am going to close this PR by 5th July if there is no other responses on this PR.

Emily-Jiang:
Close this PR as per my comments above. Please reopen if you still think this should be discussed again.

radcortez:
Yes, I believe that making these beans `@RequestScoped` was a mistake. In most cases, you are not going to observe any changes in the injected values. If we solely consider the specified sources of MP Config, both Env, and Properties are immutable, and System Properties could be changed programmatically (but not really usual). We have to question if this is really useful.

Yes, users could write their own dynamic source and observe new values being loaded on each request, but how many times does that happen on an app lifecycle? Is it worth paying the price for this?

PR #767, opened 2023-01-18
Avoid throwing Exception for Optional types and ConfigValue if expression cannot be expanded

jbescos:
After upgrading [Helidon to Weld 5.x](https://github.com/helidon-io/helidon/pull/5830) I found some Config TCK issues with the version of the master, that are related to this [PR](https://github.com/eclipse/microprofile-config/pull/701).

I am not expert on microprofile-config, but does these changes in one test make sense?.


jbescos:
@essobedo could you take a look please?.

essobedo (approved):
Good catch, it was a long time ago for me, but at least it makes sense 👍

radcortez (approved):

radcortez:
This helps with some errors, but we may have other issues.

radcortez:
Thanks for the PR!

PR #770, opened 2023-02-24
Update to MP Parent 3.0

radcortez:


Emily-Jiang:
Do we need to update the mp parent as we don't have dependency on the new core profile? I thought a spec should not update the dependency unless they need to utilise a particular feature.

radcortez:
It requires the parent to update plugins. Old versions do not work with Maven 3.9.x.

Emily-Jiang:
It might make sense to fix the parent-2.x without pulling in Jakarta EE 10 but with Maven 3.9.x update.

radcortez:
At some point, all modules will need to be updated, and maintaining two streams at this point is just extra work for little gain.

If we want to avoid an update on the dependency side, then we need to move the Jakarta dependencies outside the Parent POM to another project, so they can be updated independently.

PR #778, opened 2023-04-15
[Fix 777] Add javax.inject.Qualifier to allow integration with Guice.

scr-oath:
Add `@javax.inject.Qualifier` to `@ConfigProperty` in order to allow integration with Guice; make the dependency optional so that the dependency is not transitive.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -39,6 +39,11 @@
             <groupId>org.osgi</groupId>
             <artifactId>org.osgi.service.cdi</artifactId>
         </dependency>
+        <!-- https://github.com/eclipse/microprofile-config/issues/777 -->
+        <dependency>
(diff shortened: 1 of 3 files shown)


Emily-Jiang:
All MicroProfile specs have updated from javax to jakarta. If you want to use javax namespace, you should use the previous version of MicroProfile specs. For config, it will be version 2.x or earlier.

Emily-Jiang:
@scr-oath in light of the above conversation, I am going to close this PR by the end of this week if there is no more update on this. By the way, if you want to contribute to Eclipse projects, you will need to sign a ECA as mentioned in the above check failures above.

Emily-Jiang:
close this PR as per the above comment. Please reopen if you think we need to discuss more on this.

Issue #779, opened 2023-05-11
Some Assert parameters are backwards


tevans78:
In TestNG assertions, the actual value comes first, then the expected value (this is the opposite to JUnit).

The example I have found is in DefaultConfigSourceOrdinalTest but I am sure there will be others. Will update this issue when I find them and then do a PR.

https://github.com/eclipse/microprofile-config/blob/08955cc3c7dbd5b125299fec0d1ee404ed8e967a/tck/src/main/java/org/eclipse/microprofile/config/tck/configsources/DefaultConfigSourceOrdinalTest.java#LL89C21-L89C21

PR #780, opened 2023-05-15
Switch expected and actual values to match TestNG assertEquals

tevans78:
fixes https://github.com/eclipse/microprofile-config/issues/779

In many cases, when using `Assert.assertEquals`, the expected value argument was first rather than second. While this is correct when using JUnit Assertions, MP Config uses TestNG where the actual value comes first and the expected value is second.

This should make no functional difference to the tests but it will mean that the messages are correct should the test fail.


Emily-Jiang:
@eclipse-microprofile-bot please build

Emily-Jiang (approved):

PR #782, opened 2023-06-06
Clarify how source ordinal applies with an active profile

### LINKED ISSUE #781: Spec of precedence for sources when using profiles seems vague
**Description**

When activating profiles, the [3.0 spec says this about precedence of sources](https://download.eclipse.org/microprofile/microprofile-config-3.0/microprofile-config-spec-3.0.html#_on_property_level):

> Conforming implementations are required to search for a configuration source with the highest ordinal (priority) that provides either the property name or the “profile-specific” property name. If the configuration source provides the “profile-specific” name, the value of the “profile-specific” property will be returned. If it doesn’t contain the “profile-specific” name, the value of the plain property will be returned.

Let's set up a context before we head to why this might be a non-complete definition.

We have 2 sources activated:
1. microprofile-config.properties, source ordinal = 100: 
    ```
    my.foobar=test
    %test.my.foobar=notest
    ```
2. Environment variable, source ordinal = 300: 
    ```
    MY_FOOBAR=BBQ
    ```

There are two ways to interpret the spec when activating the profile `test`:

A) Lookup of `my.foobar` should return `notest`. The higher ordinal source does not provide a profiled value (which can be set using `_TEST_MY_FOOBAR=BBQ2`), thus the value from the properties file wins (despite it's lower ordinal).
B) Lookup of `my.foobar` should return `BBQ` because although the higher ordinal source has no profiled setting, it shall still win because of the higher ordinal.

When testing with SmallRye Config (@radcortez), it seems that variant B is implemented.
When testing with Payara (@smillidge), it seems that variant A is implemented.

[I tried to reach out to Payara in their forum](https://forum.payara.fish/t/microprofile-config-not-using-value-from-higher-ordinal-source-with-profile/618), but never received an answer. It looked like this aspect was discussed when profiles were first added [in the Github Issue](https://github.com/eclipse/microprofile-config/issues/418#issuecomment-618678228).

May I kindly request clarification which way it was intended to be @Emily-Jiang  ?

If it helps, I'm happy to create a reproducer to demonstrate the behaviour. Thank you very much y'all!

poikilotherm:
Closes #781 

This pull request tries to rephrase parts of the profile on property level spec to clarify how ordinal of sources and the profiled vs. unprofiled values work together.

### DIFF
--- a/spec/src/main/asciidoc/configprofile.asciidoc
+++ b/spec/src/main/asciidoc/configprofile.asciidoc
@@ -45,7 +45,9 @@ If the property `mp.config.profile` is specified in multiple config sources, the
 The configuration property that utilizes the Config Profile is called a "profile-specific" property. A "profile-specific" property name consists of the following sequence: `% <profile name>.<original property name>`.
 
 Conforming implementations are required to search all configuration sources that provide either the property name or the "profile-specific" property name.
-If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned. If it doesn't contain the "profile-specific" name, the value of the plain property will be returned.
+If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned.
+If it doesn't contain the "profile-specific" name, the value of the plain property will be returned.
+Thus, the "profile-specific" concept only influences which value is to be surfaced per config source.
 
 For instance, a config source (ordinal 100) can be specified as follows.
 
@@ -76,8 +78,7 @@ The properties `%live.vehicle.name` and `%testing.vehicle.name` are inactive con
 If `mp.config.profile` is set to `live`, the property `%live.vehicle.name` is the active property. The `vehicleName` will be `train`. Similarly, `bike` will be the value of `vehicleName`, if the profile is `testing`.
 If no profile is active, `lorry` will be the value of `vehicleName`.
 
-When another configuration source provides the same property, the final resulting value of the lookup depends on the ordinal (priority) of all providing sources.
-
+As explained above, the override scope is per configure source.
 For instance, an additional source (ordinal 300) may be specified as follows:
 
 [source, text]
--- a/spec/src/main/asciidoc/configprofile.asciidoc
+++ b/spec/src/main/asciidoc/configprofile.asciidoc
@@ -44,11 +44,10 @@ If the property `mp.config.profile` is specified in multiple config sources, the
 ==== On Property level
 The configuration property that utilizes the Config Profile is called a "profile-specific" property. A "profile-specific" property name consists of the following sequence: `% <profile name>.<original property name>`.
 
-Conforming implementations are required to search for a configuration source with the highest ordinal (priority) that provides either the property name or the "profile-specific" property name.
+Conforming implementations are required to search all configuration sources that provide either the property name or the "profile-specific" property name.
 If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned. If it doesn't contain the "profile-specific" name, the value of the plain property will be returned.
 
-
-For instance, a config source can be specified as follows.
+For instance, a config source (ordinal 100) can be specified as follows.
 
 [source, text]
 ----
@@ -75,6 +74,18 @@ In more details, if `mp.config.profile` is set to `dev`, the property `%dev.vehi
 The properties `%live.vehicle.name` and `%testing.vehicle.name` are inactive config properties and don't override the property `vehicle.name`.
 
 If `mp.config.profile` is set to `live`, the property `%live.vehicle.name` is the active property. The `vehicleName` will be `train`. Similarly, `bike` will be the value of `vehicleName`, if the profile is `testing`.
+If no profile is active, `lorry` will be the value of `vehicleName`.
+
+When another configuration source provides the same property, the final resulting value of the lookup depends on the ordinal (priority) of all providing sources.
+
+For instance, an additional source (ordinal 300) may be specified as follows:
+
+[source, text]
+----
+vehicle.name=helicopter
+----
+
+The value of `vehicleName` will be `helicopter` no matter which profile might be active, as the higher ordinal (300 versus 100) makes the later source win.
 
 ==== On Config Source level
 
--- a/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/OverrideConfigProfileTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/OverrideConfigProfileTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
  *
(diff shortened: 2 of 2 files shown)


poikilotherm:
> @poikilotherm you need to sign ECA form first. Please click on the Details above and follow instructions there.

Was already on it :wink: Successfully signed and re-validated just now.

Please let me know what you think about the extended spec. Thank you very much for looking into this @Emily-Jiang !

Emily-Jiang:
Thank you @poikilotherm for the PR! Can you also add a couple of sentence on the spec to state the behavior?

poikilotherm:
> Thank you @poikilotherm for the PR! Can you also add a couple of sentence on the spec to state the behavior?

Hi @Emily-Jiang thanks for the feedback. Unfortunately, I'm not sure what you are referring to? Is there some other place in the spec that I should edit besides the profile details that needed clarification? Any hints are much appreciated!

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -75,6 +74,18 @@ In more details, if `mp.config.profile` is set to `dev`, the property `%dev.vehi
...
+When another configuration source provides the same property, the final resulting value of the lookup depends on the ordinal (priority) of all providing sources.
+
+For instance, an additional source (ordinal 300) may be specified as follows:
+
+[source, text]
+----
+vehicle.name=helicopter
+----
+
+The value of `vehicleName` will be `helicopter` no matter which profile might be active, as the higher ordinal (300 versus 100) makes the later source win.
```
Emily-Jiang, line 88: For the above sentences, I suggest to just say:
The config profile controls which property value will be surfaced beyond each configure source. In other words, the override scope is per configure source. You can then add an example to explain that if you like.

spec/src/main/asciidoc/configprofile.asciidoc
```diff
@@ -44,11 +44,10 @@ If the property `mp.config.profile` is specified in multiple config sources, the
 ==== On Property level
 The configuration property that utilizes the Config Profile is called a "profile-specific" property. A "profile-specific" property name consists of the following sequence: `% <profile name>.<original property name>`.
 
-Conforming implementations are required to search for a configuration source with the highest ordinal (priority) that provides either the property name or the "profile-specific" property name.
+Conforming implementations are required to search all configuration sources that provide either the property name or the "profile-specific" property name.
 If the configuration source provides the "profile-specific" name, the value of the "profile-specific" property will be returned. If it doesn't contain the "profile-specific" name, the value of the plain property will be returned.
```
Emily-Jiang, line 48: After this line, I suggest adding:
The "profile-specific" concept only influence which value is to be surfaced per config source.

Emily-Jiang:
> > Thank you @poikilotherm for the PR! Can you also add a couple of sentence on the spec to state the behavior?
> 
> Hi @Emily-Jiang thanks for the feedback. Unfortunately, I'm not sure what you are referring to? Is there some other place in the spec that I should edit besides the profile details that needed clarification? Any hints are much appreciated!

Sorry for not being clear. I have added more detailed thoughts inline. Will you be able to take at this PR?

poikilotherm:
@Emily-Jiang thank you for your additional feedback! 

I rephrased a little as you suggested. Please let me know if this is better now.

I already worked on adding a new TCK test locally, but struggeling to find a property not altered by other tests, so it can be used here. Will continue coding, maybe push something preliminary, too, so y'all could have a look.

spec/src/main/asciidoc/configprofile.asciidoc
poikilotherm: Thanks, added.

spec/src/main/asciidoc/configprofile.asciidoc
poikilotherm: Thanks, refactored.

poikilotherm:
@Emily-Jiang @radcortez @smillidge I just added the new TCK test in this PR.

If you want to see it in action, I took notes how to execute the test in addition to the others: https://notes.desy.de/iDjhL8UESlO-_RGNKLDRZQ

As expected, Payara fails to be spec compliant, while SmallRye is compliant.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/OverrideConfigProfileTest.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
```
Emily-Jiang, line 2:
```suggestion
 * Copyright (c) 2023 Contributors to the Eclipse Foundation
```

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/OverrideConfigProfileTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.testng.annotations.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+/**
+ * Test cases for Config profile
+ *
+ * @author Emily Jiang
```
Emily-Jiang, line 41: You either delete this line or update to your name :)

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/OverrideConfigProfileTest.java
poikilotherm: Done.

tck/src/main/java/org/eclipse/microprofile/config/tck/profile/OverrideConfigProfileTest.java
poikilotherm: Done

poikilotherm:
> @poikilotherm can you address the above comments so that I can merge your PR?

Please excuse my slow response time @Emily-Jiang - I'm on vacation. Applied changes as requested. Thank you again for looking into this!

Emily-Jiang (approved):
LGTM! Thank you @poikilotherm for your contribution!

Emily-Jiang:
@poikilotherm Can you fix the following format issue?

> The GHA failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.22.0:validate (default-cli) on project microprofile-config-tck: File '/home/runner/work/microprofile-config/microprofile-config/tck/src/main/java/org/eclipse/microprofile/config/tck/profile/OverrideConfigProfileTest.java' has not been previously formatted. Please format file (for example by invoking `mvn -f tck net.revelc.code.formatter:formatter-maven-plugin:2.22.0:format`) and commit before running validation!

otaviojava (approved):

poikilotherm:
@Emily-Jiang done! :smile:

Emily-Jiang:
I will squash and merge your PR @poikilotherm and deal with the format with a separate PR if the build still fails.

poikilotherm:
Thank you very much @Emily-Jiang !

Much appreciated! Glad this is fixed now.

PR #784, opened 2023-06-29
Revert the updates #773 #770

Emily-Jiang:
Revert the updates done #773 #770 so that the config 3.1 will continue working with EE 9.1.


Emily-Jiang:
This build failure is expected as I have to release mp parent first.

tevans78 (approved):

